PR #228, opened 2013-12-28
Enable `CBLUITableSource` to create sections via delegate method

p2:
Change `CBLUITableSource` to enable table view sections.

One major internal change to how instances handle rows, external API  still works the same but gains a `sections` property and two new delegate methods. If the delegate does **not** implement the delegate method no sections are being created. The delegate has to implement `couchTableSource:sectionizeRows:` and has to create and return sections itself.

Here's an example to create sections by `name.code`, up to the first period in _"code"_:

```
- (NSMutableArray *)couchTableSource:(CBLUITableSource *)source
                      sectionizeRows:(NSArray *)rows
{
    NSSortDescriptor *sort = [NSSortDescriptor sortDescriptorWithKey:@"document.properties.name.code" ascending:YES];
    NSArray *sortedRows = [rows sortedArrayUsingDescriptors:@[sort]];

    // sectionize into arrays
    NSMutableArray *array = [NSMutableArray array];
    NSString *previousKey = nil;
    NSMutableArray *currentArray = nil;
    for (CBLQueryRow *row in sortedRows) {
        NSString *currentKey = row.document.properties[@"name"][@"code"];
        NSUInteger location = [currentKey rangeOfString:@"."].location;
        currentKey = (NSNotFound != location) ? [currentKey substringToIndex:location] : currentKey;
        if (![previousKey isEqualToString:currentKey]) {
            previousKey = currentKey;
            currentArray = [NSMutableArray arrayWithObject:row];
            [array addObject:currentArray];
        }
        else {
            [currentArray addObject:row];
        }
    }

    return array;
}
```

### DIFF
--- a/Source/API/CBLUITableSource.h
+++ b/Source/API/CBLUITableSource.h
@@ -26,16 +26,22 @@
 @property (retain) CBLLiveQuery* query;
 
 /** Rebuilds the table from the query's current .rows property. */
--(void) reloadFromQuery;
+- (void) reloadFromQuery;
 
 
 #pragma mark Row Accessors:
 
-/** The current array of CBLQueryRows being used as the data source for the table. */
-@property (nonatomic, readonly) NSArray* rows;
+/** The current array of sections containing arrays of CBLQueryRows; this is used to feed the table. */
+@property (nonatomic, readonly) NSArray* sections;
 
-/** Convenience accessor to get the row object for a given table row index. */
-- (CBLQueryRow*) rowAtIndex: (NSUInteger)index;
+/** The current array of CBLQueryRows being used as the data source for the table.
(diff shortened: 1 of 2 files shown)


minusplusminus:
Still an option I would like to have.

p2:
I have been using a custom class ``, which is just an altered version of the bundled source, for 2-3 years now, works fine.

I've put it into a Gist, maybe it's useful to you. [Example implementation](https://gist.github.com/p2/815203a0736e5d1dc9b5ed0d32917f88#file-exampletableviewcontroller-m) also included:

https://gist.github.com/p2/815203a0736e5d1dc9b5ed0d32917f88

minusplusminus:
nice, thanks!

PR #383, opened 2014-07-09
cbl incremental store to-many

Rostmen:
fix for to-many relationship


pasin:
@Rostmen I just have a chance to checkout and test your code and it seems to work fine with a condition that the inverse relationship is needed. I plan to do more testing and merge your PR. As it cannot do Auto merge now, can you update your code with the current Master?

Regarding the inverse relationship, we could do another round of fixing that after the PR is merged. @ManuTroquet , as you have a solution for this issue, can you submit a PR after this PR is merged?

pasin:
Can one of your elaborate about the following comment? What are 3 docs that will get updated? Based on my understanding, only 2 docs (the relationship doc and the doc that contains the relationship) will get updated? 

```
CBLIS will update 3 docs instead one in the case you has inverse relationships.
```

ManuTroquet:
@pasin You're right. it's 2 docs are updated. Else this correction is useless except to slow down the code. I do not take this correction into my fork . Look at my review of 12 December and my PR 546. I have to thing 2 features to do this. Of course if it possible! I don’t analyse the stuff for now. Store 2 tables in one doc (eg invoice and invoice line ) . Modify the query code to handle this feature with CBLQueryBuilder.

Rostmen:
Hi @pasin I update my code with current master.

pasin:
@Rostmen I'm sorry to say that we will not accept your PR. I had a meeting with @snej and @zgramana to review what the right solution to support to-many relationship (both with and without inverse relationship) in CBLIncrementalStore. We all agreed that storing a relationship array on the parent entity object is not the optimized solution and would lead to a lot of conflicts e.g. If there are 2 people trying to add a new entity to the relationship, the conflict would occur. Also I would recommend you to use declare an inverse relationship for your to-many relationship as it has been optimized for both storing and fetching.

By the way, below is the list of the fixes proposed in your PR (some of them have already been addressed):
- Fix NSIncrementalStore not firing fault (https://github.com/couchbase/couchbase-lite-ios/commit/cab0902406504373b4669ccd31e0e0e89492e06b)
- Fix query via relationship (https://github.com/couchbase/couchbase-lite-ios/commit/bb8f632c9cee221598232afc059757080233f8e7, https://github.com/couchbase/couchbase-lite-ios/commit/9a958d8b8cb4add786c622c3e573c768c9759ce6)
- Add support for non-inverse to-many relationship (https://github.com/couchbase/couchbase-lite-ios/pull/570)
- Add support for querying via an abstract entity (https://github.com/couchbase/couchbase-lite-ios/issues/605)

PR #546, opened 2014-12-05
All types may be external if the flag is set StoredInExternalRecord.

ManuTroquet:
Any type can be an attachment with the flag StoredInExternalRecord.
The type NSTransformableAttributeType is managed.
The NSBinaryDataAttributeType type can be stored internally.
The save is in a CBL transaction.
For old, you just check the flag StoredInExternalRecord for NSBinaryDataAttributeType attribute to work as before.
According to my tests, the code can be up to 3 times faster.

### DIFF
--- a/Source/API/Extras/CBLIncrementalStore.m
+++ b/Source/API/Extras/CBLIncrementalStore.m
@@ -340,152 +340,164 @@ static NSString *CBLISResultTypeName(NSFetchRequestResultType resultType);
     return success;
 }
 
+- (void) saveRequest:(NSSet *)objects withContext:(NSManagedObjectContext *)context error:(NSError **)outError changedEntities:(NSMutableSet *)changedEntities
+{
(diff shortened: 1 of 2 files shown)


pasin:
@ManuTroquet , I have reviewed your PR with @snej and @zgramana and we are sorry that we will not accept the PR. The main reason is that :

The PR overrides the actual meaning of the isStoredInExternalRecord property, which is supposed to use for Mac Spotlight indexing and have no effect on the iOS platform. Using the property for something else would confuse other developers either Mac or iOS developers.

We already have a way to store data as an attachment by specifying the data type to binary. Of course when storing and retrieving would require data conversion. The same data conversion process happens in your PR as well.

Please feel free to let us know your specific use cases or requirements associated with your PR so that we could help optimizing or finding any other alternative solutions. Again, thanks a lot for the PR and help improving CBLIncrementalStore.

Issue #521, opened 2014-11-10
CBLModel: remove -initWithDocument: from the public API


snej:
-initWithDocument: is dangerous: it can be called with a nil document by accident, which leaves the instance in a confused state where it will later fail assertions, e.g. when you set properties on it. The documentation for -initWithDocument: tries to spell out how it's safe to use it, but doesn't really; it's just too confusing.

Subclassers do need a place they can initialize ivars in a new instance. I can add a new public method that's called during initialization, that they can override to do that.

nl:
After making those changes, I get an assertion fire whenever I try to set properties on a (new) document in awakeFromInitializer. https://gist.github.com/nl/67fc4cf3eee4511a1a92

pasin:
Reopen the issue. This is a bug. The internal document will get created when the database is set, which is after the awakeFromInitializer is called.

pasin:
@nl, I did fix the issue. Could you check it out?

tleyden:
> After trying both one and two init method implementation, I think I prefer adding the database param to the initWithDocument. It feels better and cleaner.

Do the api docs need to be updated?

snej:
> Do the api docs need to be updated?

Yeah; the release-note text I posted earlier describes what changed. The header-docs have been updated, but since we don't have versioned docs on our portal we can't update those until 1.1 is released.

Jamiltz:
Tried using the new changes with swift.
Ran into compiler errors because objective-c convenience constructors (class methods) are bridged to Swift as convenience initializers (see http://stackoverflow.com/questions/24295476/swift-can-not-use-nsimage-imagenamed?answertab=votes#tab-top).
So `+ (instancetype) initWithDocument:` is bridged to `init(document: Document)` as a swift convenience init.
When adding a custom initializer in swift models, the compiler enforces to call a designated initializer.
![screen shot 2015-03-01 at 22 01 51](https://cloud.githubusercontent.com/assets/2589337/6433143/b97d3e9e-c05e-11e4-9e31-aab695ac6acc.png)
It seems that we need to expose a designated init in the public API for compatibility with Swift initialization rules.

snej:
With the recent API change, you shouldn't have an initializer in a CBLModel subclass, and you shouldn't instantiate models by calling initializers. Instead, you should override `-awakeFromInitializer` to perform initialization tasks.

SteffeyDev:
[...]

PR #558, opened 2015-01-02
Fix awakeFromInitializer document assertion fail (#521)

pasin:
* Not checking the existence of the document when the model is new.
* Include API_ModelAwakeFromInitializer unit test.

#521

### DIFF
--- a/Source/API/CBLModel.m
+++ b/Source/API/CBLModel.m
@@ -30,8 +30,9 @@
 @dynamic type;
 
 
-- (instancetype) initWithDocument: (CBLDocument*)document
+- (instancetype) initWithDocument: (CBLDocument*)document orDatabase:(CBLDatabase *)database
 {
+    NSParameterAssert(document || database);
     self = [super init];
     if (self) {
         if (document) {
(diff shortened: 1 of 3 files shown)


snej:
This will fix the immediate crash, but I think it's wrong that the model doesn't have a document at the time that awakeFromInitializer is called. It should be an invariant that a model always has a document.

I think the right fix is to copy these lines from `-setDatabase:` into the init method, in the `document==nil` case:
```
        NSString* docID = [self idForNewDocumentInDatabase: db];
        self.document = docID ? [db documentWithID: docID] : [db createDocument];
```

pasin:
That doesn't work with -initWithDocument as at the time is the document is nil, the database is also nil too. Should we create a new init method (e.g. -initWithDatabase?) for a new fresh model instead?

snej:
Looks good, but you're going to have to merge with master and resolve some conflicts, due to my refactoring of the unit tests.

pasin:
Rebased, fixed the conflict, and re-added the unit test for the awakeFromInitializer method. Merging now.

PR #559, opened 2015-01-05
Refactorings for GroceryList Example

pual:
-Removed unused variable
-Switched to ObjC 2.0 syntax
-Removed unnecessary import statements


snej:
It's not an example, it's just a host app for running tests. If you're looking for the real example, that's [Grocery Sync](https://github.com/couchbaselabs/Grocery-Sync-iOS). (This code was copied from an earlier version of Grocery Sync but there's no point in keeping it up to date.)

PR #561, opened 2015-01-06
Adding a way to set the type of a new document automatically

sarbogast:
based on the mapping available in database.modelFactory

I tried to write a unit test for this new feature, but I couldn't run the unit tests to make sure it passes because I needed a signing identity. And I'm not sure it's actually correct because if you're doing real unit testing then database won't have a proper implementation of modelFactory that would be necessary for this test to pass.

### DIFF
--- a/Source/API/CBLModelFactory.m
+++ b/Source/API/CBLModelFactory.m
@@ -63,14 +63,9 @@ static CBLModelFactory* sSharedInstance;
 }
 
 - (NSString*)documentTypeForClass: (Class)modelClass {
-    NSString *type = nil;
-    for (NSString *key in _typeDict) {
-        if(modelClass == _typeDict[key]){
-            type = key;
-            break;
-        }
-    }
-    return type;
+    NSArray *keys = [_typeDict allKeysForObject:modelClass];
+    if(keys.count == 0) return nil;
+    else return keys[0];
 }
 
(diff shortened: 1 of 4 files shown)


Source/API/CBLModelFactory.m
```diff
@@ -62,6 +62,16 @@ - (Class) classForDocumentType: (NSString*)type {
     return klass;
 }
 
+- (NSString*)documentTypeForClass: (Class)modelClass {
+    NSString *type = nil;
+    for (NSString *key in _typeDict) {
```
snej: You can do this in one line as
```objc
return [[_typeDict allKeysForObject: modelClass] firstObject];
```

snej:
We're still working on making the unit tests completely portable. You can run just the model-related tests, though, using the test navigator pane in Xcode.

sarbogast:
I've tried running Model_Tests from the test navigator in XCode, but if I choose anything else than my iPhone to run the tests with, like a simulator, none of the tests appear in the test navigator anymore. They only appear with the real device selected, and then I need an identity.

snej:
Which scheme did you use? "CBL iOS Test App" should work.
Or you can select the "CBL Mac" scheme and not have to deal with simulator/device issues at all.

sarbogast:
OK, now it worked and the tests passed.

snej:
Thanks for the patch!

Issue #406, opened 2014-08-06
Importing database should import the lastSequence too


tleyden:
iOS version of: https://github.com/couchbase/couchbase-lite-android/issues/384

PR #562, opened 2015-01-08
Import last seq for replication when replace a db (#406)

pasin:
* When replacing a database, creating a local checkpoint document containing the replacing database's localUUID (ID before getting replaced).

* When a replicator fetches remote checkpoint document to setup the last sequence info, if the last sequence is nil, try to get the imported last sequence which is obtained by using the localUUID stored in the local checkpoint document if it's available.

* Added test11_ReplicationWithReplacedDatabase unit test.

#406

### DIFF
--- a/Source/API/CBLManager.m
+++ b/Source/API/CBLManager.m
@@ -446,7 +446,7 @@ static CBLManager* sInstance;
                                                toPath: dstAttachmentsPath
                                                 error: outError]) &&
             [db open: outError] &&
-            [db createLocalCheckpointDocument] &&
+            [db createLocalCheckpointDocument: outError] &&
             [db replaceUUIDs: outError];
 }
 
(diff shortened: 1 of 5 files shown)


Source/API/CBLManager.m
```diff
@@ -446,6 +446,7 @@ - (BOOL) replaceDatabaseNamed: (NSString*)databaseName
                                                toPath: dstAttachmentsPath
                                                 error: outError]) &&
             [db open: outError] &&
+            [db createLocalCheckpointDocument] &&
```
snej: -createLocalCheckpointDocument should take an out NSError parameter. Otherwise, if it fails (returns NO) then this method will return NO but won't have set its own outError to any valid value.

snej:
Looks good, aside from the lack of error return from -createLocalCheckpointDocument!

Issue #569, opened 2015-01-21
Support to-many without inverse relationship in CBLIncrementalStore


pasin:
Currently CBLIncrementalStore doesn't support to-many relationship that doesn't up inverse relationship setup. When users fetch an entity such relationships, the relationship property value will be empty.

ManuTroquet:
Can you give an example of code that show this bug?

pasin:
Steps to reproduce:
1. Create a one-to-many relationship without an inverse relationship.
2. Create parent and child objects and save.
3. Try to fetch a parent entity object and check the to-many relationship property in the next runloop to avoid caching.

Here is a quick test:
```
- (NSManagedObject *)createEnityName:(NSString *)name properties:(NSDictionary *)properties{
    NSManagedObject *object = [NSEntityDescription insertNewObjectForEntityForName:name
                                                            inManagedObjectContext:[self context]];
    for (NSString *key in [properties allKeys]) {
        [object setValue:[properties objectForKey:key] forKey:key];
    }
    return object;
}

- (void)viewDidLoad {
    [super viewDidLoad];

    NSManagedObject *pet1 = [self createEnityName:@"Pet" properties:@{ @"name" : @"Pet1" }];
    NSManagedObject *pet2 = [self createEnityName:@"Pet" properties:@{ @"name" : @"Pet2" }];
    NSSet *pets = [NSMutableSet setWithObjects:pet1, pet2, nil];
    [self createEnityName:@"Student" properties:@{ @"name" : @"Student1", @"pets" : pets }];
    [self save];

    // Fetch students in the next run loop.
    [self performSelector:@selector(fetchStudents) withObject:nil afterDelay:1];
}

- (void)fetchStudents {
    NSEntityDescription *entity = [NSEntityDescription entityForName:@"Student" inManagedObjectContext:[self context]];
    NSFetchRequest *request = [[NSFetchRequest alloc] init];
    request.entity = entity;

    NSError *error;
    NSArray *objects = [[self context] executeFetchRequest:request error:&error];
    for (NSManagedObject *student in objects) {
        NSLog(@"Student: %@", [student valueForKey:@"name"]);
        NSLog(@"Pets: ");
        NSSet *pets = [student valueForKey:@"pets"];
        for (NSManagedObject *pet in pets) {
            NSLog(@"Pet : %@", [pet valueForKey:@"name"]);
        }
    }
}
```

PR #570, opened 2015-01-21
Add support for to-many without inverse relationship (#569)

pasin:
Auto add an inverse relationship to the CBL document (not NSManagedObject) when there is no inverse relationship specified. This allows both with inverse and without inverse to-many relationship can use the inverse view to query the child objects in relation.

### DIFF
--- a/Source/API/Extras/CBLIncrementalStore.m
+++ b/Source/API/Extras/CBLIncrementalStore.m
@@ -26,6 +26,7 @@ NSString * const kCBLISErrorDomain = @"CBLISErrorDomain";
 NSString * const kCBLISObjectHasBeenChangedInStoreNotification = @"kCBLISObjectHasBeenChangedInStoreNotification";
 
 static NSString * const kCBLISTypeKey = @"CBLIS_type";
+static NSString * const kCBLISInverseRelationshipKeyFormat = @"CBLIS_parent_%@";
 static NSString * const kCBLISCurrentRevisionAttributeName = @"CBLIS_Rev";
 static NSString * const kCBLISManagedObjectIDPrefix = @"CBL";
 static NSString * const kCBLISMetadataDocumentID = @"CBLIS_metadata";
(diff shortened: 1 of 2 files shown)


Source/API/Extras/CBLIncrementalStore.m
```diff
@@ -346,255 +348,395 @@ -(BOOL)loadMetadata:(NSError **)outError
...
-            
-            CBLDocument *doc = [self.database documentWithID:[object.objectID couchbaseLiteIDRepresentation]];
-            CBLUnsavedRevision *revision = [doc newRevision];
-            [revision.properties setValuesForKeysWithDictionary:contents];
-            
-            
-            // add attachments
-            NSDictionary *propertyDesc = [object.entity propertiesByName];
+    CBLDocument *doc = [self.database documentWithID:[object.objectID couchbaseLiteIDRepresentation]];
+    if (!doc) return YES;
```
snej: This will never happen — documentWithID: always returns a document even if it doesn't exist yet. To check for a nonexistent document, test for `doc.currentRevision == nil`.

Source/API/Extras/CBLIncrementalStore.m
```diff
@@ -697,89 +839,67 @@ - (NSManagedObjectID *)newObjectIDForEntity:(NSEntityDescription *)entity refere
...
-                        [view setMapBlock:^(NSDictionary *doc, CBLMapEmitBlock emit) {
-                            if ([entityNames containsObject:[doc objectForKey:kCBLISTypeKey]] && [doc objectForKey:inverseRelNameLower]) {
-                                emit([doc objectForKey:inverseRelNameLower], nil);
-                            }
+                    NSRelationshipDescription *invRel = rel.inverseRelationship;
+                    NSString *invertRelPropName = invRel ?
+                    invRel.name : CBLISInverseRelationshipKeyForRelationName(rel.name);
+                    CBLView *view = [self.database viewNamed:viewName];
+                    [view setMapBlock:^(NSDictionary *doc, CBLMapEmitBlock emit) {
+                        if ([entityNames containsObject:[doc objectForKey:kCBLISTypeKey]] &&
```
snej: This will throw an exception if the document doesn't contain kCBLISTypeKey. (Looks like the previous revision had the same problem, so this isn't a regression.)

pasin:
Fix issues according to @snej code review and rebase.

pasin:
@ManuTroquet, thanks for bringing the point about existing apps. One thing that's missing in the PR is migrating existing data logic.

pasin:
After rethinking about this, I'm pulling the PR back. IMO I think we should encourage users to set inverse relationships as:
1. There will be only one structure for to-many relationships in the underlining documents.
2. Inverse relationship was automatically converted to view and was optimized for fetching. 

The new solution will be adding a warning to users when they don't have inverse relationship setup.

Issue #571, opened 2015-01-23
Use CBLQueryBuilder in CBLIncrementalStore


pasin:
Use CBLQueryBuilder in CBLIncrementalStore so that:
1. View can be auto-generated based on fetch request's predicate.
2. Support more types of predicates such as compound predicates.
3. Simplify query coding logic.

Limitation:
No relationship queries (Join) supported by CBLQueryBuilder. We could detect this and fall back to full-table scan as the current CBLIncrementalStore does when there are no views specified.

snej:
Can this be closed now that the PR has been merged? Or is there more work?

pasin:
This can be closed now.

PR #582, opened 2015-02-02
CBLIncrementalStore - Utilize CBLQueryBuilder and add relationship search (#571)

pasin:
CBLIncrementalStore - Utilize CBLQueryBuilder and add relationship search

- Use CBLQueryBuilder to fetch documents with predicates that do not have relationship keypath, otherwise fall back to the old way of using view and post filtering.
- Scan the original predice to generate a new predicate with template and a variable dictionary.
- Refator executeRequest:request:context:error method by seperating out each type of requests into its own method.
- Update unit tests

#571

### DIFF
--- a/Source/API/Extras/CBLIncrementalStore.m
+++ b/Source/API/Extras/CBLIncrementalStore.m
@@ -24,6 +24,9 @@
 #  error This class requires ARC!
 #endif
 
+#define INFO(FMT, ...)  NSLog(@"[CBLIS] INFO " FMT, ##__VA_ARGS__)
+#define WARN(FMT, ...)  NSLog(@"[CBLIS] WARNING " FMT, ##__VA_ARGS__)
+#define ERROR(FMT, ...)  NSLog(@"[CBLIS] ERROR " FMT, ##__VA_ARGS__)
 
 NSString * const kCBLISErrorDomain = @"CBLISErrorDomain";
 NSString * const kCBLISObjectHasBeenChangedInStoreNotification = @"kCBLISObjectHasBeenChangedInStoreNotification";
@@ -36,11 +39,10 @@ static NSString * const kCBLISFetchEntityByPropertyViewNameFormat = @"CBLIS/fetc
 static NSString * const kCBLISFetchEntityToManyViewNameFormat = @"CBLIS/%@_tomany_%@";
 
 
-// utility functions
+// Utility functions
 static BOOL CBLISIsNull(id value);
 static NSString *CBLISToManyViewNameForRelationship(NSRelationshipDescription *relationship);
 static NSString *CBLISResultTypeName(NSFetchRequestResultType resultType);
-
 
 @interface NSManagedObjectID (CBLIncrementalStore)
 - (NSString*) couchbaseLiteIDRepresentation;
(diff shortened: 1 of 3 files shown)


pasin:
As there are many changes across CBLIncrementalStore.m, below shows the core changes in the pull request:

Line 647 - 1085 : [QueryBuilder Related Changes] (https://github.com/couchbase/couchbase-lite-ios/blob/feature/issue_571_cblis_query/Source/API/Extras/CBLIncrementalStore.m#L647-L1085)

Line 1263 - 1369 : [Add Relationship Search to full entity search] (https://github.com/couchbase/couchbase-lite-ios/blob/feature/issue_571_cblis_query/Source/API/Extras/CBLIncrementalStore.m#L1263-L1369)

Source/API/Extras/CBLIncrementalStore.m
```diff
@@ -761,239 +634,797 @@ - (void) _initializeViews
...
-/** Queries the database by a given fetch request. Checks the cache for the result first. */
-- (NSArray*) _queryObjectsOfEntity:(NSEntityDescription*)entity byFetchRequest:(NSFetchRequest*)fetch inContext:(NSManagedObjectContext*)context error:(NSError**)outError
-{
-    id cached = [self _cachedQueryResultsForEntity:entity.name predicate:fetch.predicate];
-    if (cached) {
-        return cached;
+    if (!builder)
+        return nil;
+
+    [self cacheQueryBuilder:builder
```
snej: Don't you only need to cache the builder if it's new? You can move lines 774-779 inside the if() block above.

Source/API/Extras/CBLIncrementalStore.m
```diff
@@ -1229,148 +1629,70 @@ - (CBLQueryEnumerator*) _queryEnumeratorForQuery:(CBLQuery*)query error:(NSError
...
+    SHA1_Final(digest, &ctx);
+    return [NSData dataWithBytes: &digest length: sizeof(digest)];
 }
 
-- (NSArray*) _cachedQueryResultsForEntity:(NSString*)entityName predicate:(NSPredicate*)predicate
-{
-    NSString *cacheKey = [NSString stringWithFormat:@"%@_%@", entityName, predicate];
-    return [_fetchRequestResultCache objectForKey:cacheKey];
+- (NSString *)encodeCacheKey:(id)key {
+    NSData *archive = [NSKeyedArchiver archivedDataWithRootObject:key];
```
snej: I found out recently that this approach doesn't work reliably :(
NSKeyedArchiver doesn't always create identical data for equal objects, maybe because dictionaries don't get written in a consistent order. I need to fix this in CBLQueryBuilder too. I'll probably end up converting the NSPredicate to a string instead of archiving it.

Source/API/Extras/CBLIncrementalStore.m
```diff
@@ -761,239 +634,797 @@ - (void) _initializeViews
...
+#pragma mark - Misc
+
+- (id) _couchbaseLiteValueForAttributeValue:(id)value ofType:(NSAttributeType)type {
+    id result = nil;
+
+    switch (type) {
+        case NSInteger16AttributeType:
+        case NSInteger32AttributeType:
+        case NSInteger64AttributeType:
+            result = [NSNumber numberWithLong:CBLISIsNull(value) ? 0 : [value longValue]];
```
snej: In the typical case where `value` is already an NSNumber, this is an expensive no-op.
If it's not an NSNumber (and not NSNull), it'll throw an exception ... is that acceptable or does this method need to fail gracefully?
The same goes for the float types too.

pasin: I agree that we don't need to create a new NSNumber of it's already NSNumber. This is from the original code so I will run some tests to make sure that it will not break anything.

Regarding the data type check, as the value is coming from CoreData, I think we could expect that the value has a valid type. But when converting from CBLDocument to NSManagedObject , I think it should validate the type there.

PR #589, opened 2015-02-10
Allow a doc to be updated more than once in a single transaction

### LINKED ISSUE #256: Conflict error when updating a document multiple times in transaction block
Test Definition: Create a document and update a value for n times. 
When updating the value using the "in-transaction" block an exception is received. 

The bug can be reproduced by the following code: 
https://github.com/couchbaselabs/iOS-WorkerBee/blob/Performance_Testing_2/Tests/Test8_DocRevisions.m

The exception does not occur when the document revision is done without a in-transaction block.

snej:
This didn't use to work correctly because the CBLDocument's currentRevision didn't
update to the new revision after the first update; then the second update would use
an obsolete _rev and fail with a conflict.
Added code to inform the CBLDocument immediately, even during a transaction, but
prevent the change notification from being posted till the transaction ends.
Fixes #256

### DIFF
--- a/Source/API/CBLDatabase.m
+++ b/Source/API/CBLDatabase.m
@@ -109,7 +109,7 @@ static id<CBLFilterCompiler> sFilterCompiler;
     BOOL external = NO;
     for (CBLDatabaseChange* change in changes) {
         // Notify the corresponding instantiated CBLDocument object (if any):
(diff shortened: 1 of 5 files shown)


snej:
Updated the commit to handle an aborted transaction (thanks Pasin!) 

Look at the -storageExitedTransaction: method, and at the second transaction in the test case.

Issue #536, opened 2014-11-21
Mac Application Reachability issue


pasin:
Per reported in #522, when using CBL in mac app, reachability stops working correctly sometime and always reported that the connected server is not reachable.

nl:
This is still very much a showstopper for using continuous replications in Mac apps.

snej:
[Logs here](https://gist.githubusercontent.com/nl/22abd61c0e880171ae7f/raw/ceecc22a1aa6552692a4ff05218c86c91ec7746f/gistfile1.txt).

There are some well-known networking bugs in 10.10, especially having to do with DNS and Bonjour, and I wonder if they're what's causing this. (But @nl said in #522 that this happens on 10.9 as well...)

snej:
@pasin, didn't you change the replicator so it tries to connect whether or not Reachability says the server is reachable? I remember that being discussed a few months ago...

nl:
I believe this change affected one-shot replications, not continuous. I can confirm one-shot replications don't exhibit this odd reachabilty behavior, and that whenever I run into this bug with continuous replications (in 1.0.4), switching them to noncontinuous fixes it and the revisions atart pushing/pulling right away.

pasin:
We rolled back all the changes as we don't have a complete solution. 

We tried to change the replicators to connect anyway and use the reachability only for triggering the retry logic in replicators and change trackers. The replicator also has an option to set network type (WiFi or Cell). 

As we can't use the reachability to detect and limit network access, we tried to limit network type access by setting the NSMutableURLRequest's allowsCellularAccess property and CFStream's kCFStreamPropertyConnectionIsCellular property directly. The thing was we also support WebSocket and we are using CocoaHTTPServer which we couldn't find a way to limit the network access. As a result we decided to rollback all changes.

pasin:
@nl , we haven't put the fix for both one-shot or continuous replicator. Does this mean that the reachability is correct first time but the reachability is not reliable when switching the network on / off?

nl:
I'm only experiencing the bug for continuous replications. Here's what that means, whenever the bug happens (which isn't always, but when it does, it does until I reboot my Mac or switching the network on/off):
I build my app with continuous replications. The replications aren't firing, and they should.
I change the continuous = YES to NO in Xcode and build & run. The replications fire.
I change the continuous = NO to YES in Xcode and build & run. The replications don't fire.
[...]

PR #593, opened 2015-02-11
Fix continuous replication reachablity issue

pasin:
## Changes:

- Always try to bring the replication online first time regardless of the replicator mode

- When the replication is in the continuous mode and the protocol is not CBL_URLProtocol, start the reachability. The reachability is now responsible for 2 things when the host is reachable (Ignore not-reachable status, which could be false negative):

 (1) It checks if the host is reachable on an allowed network (based on the kCBLReplicatorOption_Network option value). If the host is reachable in a forbidden network, it will bring the replicator offline.

 (2) It triggers the replicator retry logic if the replication is already online. If the replicator is not online, it will bring the replicator online.

- When `ChangeTracker` or `CBLRemoteRequest` gets a `reachability` error, it will report the error immediately without retrying. Once the error gets propagated to the CBL_Replicator, the CBL_Replicator will bring the replicator offline manually and then it will try to bring it online after 5 mins.

- Expand CBLIsOfflineError to cover more `reachability` error codes.

## Note:
The reachability is still using to restrict the network access based on value set for kCBLReplicatorOption_Network option. This is not perfect as the reachability sometimes reports a wrong status. In another word, I'm ignoring this issue for now. To fix this we can set the NSMutableURLRequest's allowsCellularAccess property and CFStream's kCFStreamPropertyConnectionIsCellular property instead of using the reachability. However the solution is not completed as we are using CocoaAsycSocket for the WebSocket version of the ChangeTracker and we couldn't find a way to restrict the cellular network access.

#536

### DIFF
--- a/Source/CBL_Replicator.m
+++ b/Source/CBL_Replicator.m
@@ -434,6 +434,19 @@ NSString* CBL_ReplicatorStoppedNotification = @"CBL_ReplicatorStopped";
 }
 
 
+- (void) retryOnline {
+    [self goOnline];
+    [self reachabilityChanged:_host];
+}
+
+
+- (void) retryOnlineAfterDelay {
+    [NSObject cancelPreviousPerformRequestsWithTarget: self
+                                             selector: @selector(retryOnline) object: nil];
+    [self performSelector: @selector(retryOnline) withObject: nil afterDelay: kRetryOnlineDelay];
+}
+
+
 - (BOOL) goOffline {
     if (!_online)
         return NO;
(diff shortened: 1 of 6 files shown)


pasin:
Just notice that CBL_Puller also calls the goOffline() method by itself when its change tracker receives a reachability error (offline error). I need to fix this and propagate the error to CBL_Replicator so it can kick off a retry cycle. Taking a PR back to fix this.

pasin:
Resubmitted the PR again.

Source/CBL_Replicator.m
```diff
@@ -465,8 +476,8 @@ - (void) reachabilityChanged: (CBLReachability*)host {
 
     BOOL reachable = host.reachable && !_suspended;
     if (reachable) {
-    // Parse "network" option. Could be nil or "WiFi" or "!Wifi" or "cell" or "!cell".
-    NSString* network = [$castIf(NSString, _options[kCBLReplicatorOption_Network])
+        // Parse "network" option. Could be nil or "WiFi" or "!Wifi" or "cell" or "!cell".
+        NSString* network = [$castIf(NSString, _options[kCBLReplicatorOption_Network])
```
snej: Oops, looks like there are tab characters here.

pasin: When I look at that in my xcode, it's spaces.

pasin:
@snej , I tested more with my test app and fixed an issue that the pusher kept itself in an offline mode when coming back from the background mode while no internet connection and then having internet connection after. Basically when the app coming from the background mode, we need to try to goOnline right way as well. I revised the methods and variable names. If you have time, please review the PR.

pasin:
Taking the PR back. The solution alters the current behavior of the replication. A simpler solution might be just trying to connect to the server when the remote host is not reachable reported by the SCNetworkReachability.

Issue #584, opened 2015-02-04
Router should support "?stale=" parameter on view query requests


snej:
Our REST API implementation doesn't support the `?stale=` query parameter from the [CouchDB API](http://docs.couchdb.org/en/latest/api/ddoc/views.html). This means there's no way for a PhoneGap client to query a view without updating the index first.

PR #598, opened 2015-02-13
Add stale parameter support on view query requests

pasin:
- Based on CouchDB spec (http://wiki.apache.org/couchdb/HTTP_view_API), the stale option can be 'ok' or 'update_after'.

- If there is no stale specified or if the view has never updated its index before (first time), the index will be updated first. The result will be queried and returned after the index is updated.

- If stale=ok is set, the view will not update its index and the result will be returned as is.

- If stale=update_after, the view will return the result and then update its index after.

#584

### DIFF
--- a/Source/CBL_Router.m
+++ b/Source/CBL_Router.m
@@ -204,6 +204,17 @@
     options->group = [self boolQuery: @"group"];
     options->content = [self contentOptions];
 
+    // Stale options (ok or update_after):
+    NSString *stale = [self query: @"stale"];
+    if (stale) {
+        if ([stale isEqualToString:@"ok"])
+            options->indexUpdateMode = kCBLUpdateIndexNever;
+        else if ([stale isEqualToString:@"update_after"])
+            options->indexUpdateMode = kCBLUpdateIndexAfter;
+        else
+            return nil;
+    }
+    
     // Handle 'keys' and 'key' options:
     NSError* error = nil;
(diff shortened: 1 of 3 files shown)


Source/CBL_Router.m
```diff
@@ -204,6 +204,17 @@ - (CBLQueryOptions*) getQueryOptions {
...
 
+    // Stale options (ok or update_after):
+    NSString *stale = [self query: @"stale"];
+    if (stale) {
+        if ([stale isEqualToString:@"ok"])
+            options->indexUpdateMode = kCBLUpdateIndexNever;
+        else if ([stale isEqualToString:@"update_after"])
+            options->indexUpdateMode = kCBLUpdateIndexAfter;
+        else
+            return nil;
```
snej: Shouldn't return here; it'll skip setting the other options below.

pasin: I checked with CouchDB v1.6.0 and it also reported an error when I gave an invalid value.

Returning nil will allow the router handler to report an error which is kCBLStatusBadRequest when the stale option is invalid. The error status is inline with the current implementation when view query params such as startKey, endKey or, bbox is invalid. 

Let me know if you confirm that we shouldn't return the error and should use the default update_before mode for this case.

snej: Oh, I didn't know it rejected other values. I know that Couchbase Server accepts `stale=false` as well, but that's probably just because it defaults (defaulted?) to `update_after`. Never mind!

Issue #596, opened 2015-02-12
LiveQuery should re-run query from scratch after options are changed


snej:
Changing the query options of an already-running CBLLiveQuery should be supported, and should re-run the query from scratch (regardless of the view's current sequence number.)

atkit:
thanks

atkit:
Am I correct, every time we change LiveQuery parameters we should call:
```
queryOptionsChanged
```
as well

pasin:
Yes, you are correct.

snej:
Rationale was that there are a _lot_ of properties that can be set, and we thought it wasn't worth creating custom setters for every single one of them to detect changes (or adding dozens of KV observers.) Especially since it's likely you'll change more than one of them at a time, so there'd be extra complexity to wait for you to stop making changes before triggering the query...

PR #600, opened 2015-02-17
Add a method to notify LiveQuery when query options changed

pasin:
* Added a method called 'queryOptionsChanged', used for notifying LiveQuery that query params were changed and LiveQuery should re-run the query from scratch.

* Current I didn't reset the rows properties before re-run the query. Assuming that a user is observing rows property, by not resetting the rows property, users will not get any notification before the new re-run query result is done.

#596

### DIFF
--- a/Source/API/CBLQuery.h
+++ b/Source/API/CBLQuery.h
@@ -175,6 +175,10 @@ typedef NS_ENUM(unsigned, CBLIndexUpdateMode) {
     If nil, the last execution of the query was successful. */
 @property (readonly, nullable) NSError* lastError;
 
+/** Call this method to notify that the query parameters have been changed, the CBLLiveQuery object
+    should re-run the query. */
+- (void) queryOptionsChanged;
+
 @end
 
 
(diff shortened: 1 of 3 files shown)


snej:
Hm, I don't have a good feeling about adding another instance variable, since the logic in CBLLiveQuery is already complicated enough. Wouldn't it be enough to just make `queryOptionsChanged` call `[self viewChanged: nil]`? That will already force another query.

pasin:
First I have thought about doing that too, but at line [428] (https://github.com/couchbase/couchbase-lite-ios/blob/master/Source/API/CBLQuery.m#L428-L429), I understand that the sequence number given to the `runAsyncIfChangedSince: onComplete :` method should be reset as well. Please let me know if I misunderstood that.

pasin:
I removed the _queryOptionDidChange and use _lastSequence variable instead. By checking _lastSequence variable to reset the since-last-sequence param when re-querying the view, we also fixes a side bug that the since-last-sequence should be reset when the view is changed as well.

Issue #597, opened 2015-02-13
iOS Dynamic Framework / Module support


lordkev:
Opening this per the list discussion here:

https://groups.google.com/forum/#!topic/mobile-couchbase/q2kLQv01LKQ

This would allow the library to be built as an iOS 8+ dynamic framework under a new additional target. Should have a PR incoming in the next few days.

snej:
The "CBL iOS Framework" target builds a true dynamic-library framework for use in iOS 8.

Note: This library isn't fat. I'm actually unsure whether we _should_ build a fat dynamic library; I don't know if the simulator code will get stripped out when the developer's app is built.

lordkev:
Great, thanks for this. Apologies for not getting around to finishing it up - got busy with other stuff.

snej:
No prob. I felt better about doing it myself anyway … no slight intended, it's just that the build settings in the project are pretty complex and prone to breaking if not treated exactly right :)

lordkev:
None take, I understand. :)

PR #602, opened 2015-02-17
Add iOS 8 dynamic framework target

lordkev:
This adds a dynamic framework target for iOS 8+ per issue #597.

Notes:
- I had to conditionally define the Gregorian calendar identifier due to a deprecation in iOS 8
- I didn't create a framework for the listener as we don't use it, but I can add that too when I get a chance if you'd like.

### DIFF
--- a/Source/API/CBLJSON.m
+++ b/Source/API/CBLJSON.m
@@ -125,7 +125,7 @@ static NSDateFormatter* getISO8601Formatter() {
         // Thanks to DenNukem's answer in http://stackoverflow.com/questions/399527/
         sFormatter = [[NSDateFormatter alloc] init];
         sFormatter.dateFormat = @"yyyy-MM-dd'T'HH:mm:ss.SSSXXX";
-        sFormatter.calendar = [[NSCalendar alloc] initWithCalendarIdentifier:NSGregorianCalendar];
+        sFormatter.calendar = [[NSCalendar alloc] initWithCalendarIdentifier:NSCalendarIdentifierGregorian];
         sFormatter.locale = [[NSLocale alloc] initWithLocaleIdentifier:@"en_US"];
     }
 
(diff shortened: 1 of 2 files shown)


Source/API/CBLJSON.m
```diff
@@ -125,7 +125,11 @@ + (size_t) estimateMemorySize: (id)object {
         // Thanks to DenNukem's answer in http://stackoverflow.com/questions/399527/
         sFormatter = [[NSDateFormatter alloc] init];
         sFormatter.dateFormat = @"yyyy-MM-dd'T'HH:mm:ss.SSSXXX";
+#ifdef __IPHONE_8_0
+        sFormatter.calendar = [[NSCalendar alloc] initWithCalendarIdentifier:NSCalendarIdentifierGregorian];
```
snej: There's no need to #ifdef this — NSCalendarIdentifierGregorian is available back to iOS 4. Just change the constant.

lordkev: Fixed. I saw that originally when I looked at the definition in the NSCalendar header, but then I looked around just to make sure and saw some reports of people having issues with it being unavailable on iOS. Looks like they were building with Xcode 5 against iOS 7 so it shouldn't be an issue anymore.

lordkev: As long as we don't need to maintain the ability to build against previous SDKs.

snej:
I checked this out and it builds, but the output isn't fat — you get a framework that's either ARM-only or x86-only depending on whether you selected a device or simulator in the scheme pop-up. Having two frameworks like this is a pain to use in an app: you have to set up per-SDK search paths to link in one framework or the other depending on which SDK you build with.

To fix this I think you'll need to do something like what the 'CBL iOS library' target does: use a custom build script to run a second build for the 'other' target and then use lipo to merge the libraries together.

I think it would also be best to make this target link in libCouchbaseLite.a (built by the 'CBL iOS library' target) rather than build all the source files. That way it won't break whenever we add new source files, which already happened due to adding CBLSymmetricKey.m as part of the db-encryption merge.

lordkev:
Sorry, going to take another look at this but it may be a bit!

snej:
I just checked in my own change to add a dynamic framework. Try it out and let me know if it works for your needs.

Issue #542, opened 2014-12-01
Live queries in CBL iOS not updating on first usage


jpaas:
I have an iOS app that uses a few live queries and they all work quite well, except the first time the app is launched. The main view of the app is a table backed by CBLUITableSource that displays each document as a CBLModel. On first launch the list is empty until the user adds the first model. The view has a + button to create new CBLModels. Each time I add a CBLModel, they are persisted, but the livequery backing the list never gets notified of any changes. But this ONLY happens on first launch. If I kill the app and restart it, the models I added before will appear in the list and the live query works from then on and for every app launch after. I can add, delete, and edit and all of those changes will be picked up by the live query. Its only the first time the app is launched that it never works. I've turned on verbose debugging and I don't see any errors. I've attached two logs, one that shows what happens on first launch when I set up the table and create a new CBLModel. The other log shows what happens on the second and subsequent launches when I do the same thing.

I also tried killing the app immediately after initial launch and and restarting before adding my first model. In this case it worked fine.

I found a workaround which is to stop/start the livequery after the query gets updated in the 
couchTableSource:updateFromQuery:previousRows: method. This seems to to allow it receive all future notifications.

This is a swift project running on iOS8 and using a manually built head version of couchbase-lite-ios. I had to build it manually because I needed 1.0.3.1 before it was available on cocoapods. It sure would be nice if the podspec was part of the repo, so that it can be used like most other cocoapods...

```pod 'couchbase-lite-ios', git: 'https://github.com/couchbase/couchbase-lite-ios.git```

Here's the log from first launch:
```
2014-11-28 12:44:17.847 [INFO ] tapshotpro[5639:607] AppDelegate:140 - App started
2014-11-28 12:44:17.892 [INFO ] tapshotpro[5639:607] MigrationManager:29 - Last migration: 0
2014-11-28 12:44:17.892 [INFO ] tapshotpro[5639:607] MigrationManager:32 - Performing migration: 1
2014-11-28 12:44:17.909 [INFO ] tapshotpro[5639:607] TeamsViewController:35 - viewDidLoad
12:44:17.914| Query: CBLLiveQuery[kCblViewTeamsByName, mapOnly=YES]: Async query tapshotpro/kCblViewTeamsByName...
[...]

PR #603, opened 2015-02-18
Reset _willUpdate at the top of the update method

pasin:
Adding a new document which will not get emitted prior to creating a live query object could cause the _lastSequence tracked by the live query object equal to the database lastsequence (e.g. both are one). This could happen easily during the first luanch of the app and the database is fresh.

When the above situation happens, _willUpdate flag will never get reset as the update method will return immedidately before the flag is reset due to an if check at the top. As a result, the live query stops updating itself when there are new documents added to the database as the databaseChanged: call back method will return immdiately when it sees _willUpdate flag = YES.

#542

### DIFF
--- a/Source/API/CBLQuery.m
+++ b/Source/API/CBLQuery.m
@@ -408,6 +408,8 @@
 
 
 - (void) update {
+    _willUpdate = NO;
+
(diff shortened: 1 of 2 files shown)


snej:
This is tricky code, but I'm pretty sure that's the right fix.

PR #604, opened 2015-02-18
Don't create SQLite FTS or RTree tables until needed

snej:
This allows CBL to work with a SQLite build that doesn't include the fts3 or rtree modules
(for example, a stock build of SQLCipher.)
It probably also speeds up database creation somewhat.
Also, I noticed we didn't have any unit tests for geo anymore, so I went back in time and
restored the old one.


snej:
Updated the patch to handle the case where a geo/fts query is made before any geo/fts rows have been added to this index. (Initially this would fail because the tables didn't exist yet.)

PR #607, opened 2015-02-20
Added CBLModel support for inverse relations [API CHANGE]

### LINKED ISSUE #606: Inverse relation support for CBLModel
Add CBLModel support for computed inverse-relation properties (a typical ORM feature, also found in Core Data.)

For example, if the `Album` class has a relation `@property Artist* artist` (i.e. the JSON property value is the docID of a document corresponding to an `Artist` model), then it should be possible to define a property on `Artist` like `@property (readonly) NSArray* albums`, whose value is an array of the `Album` objects whose `artist` property points to the receiver.

Behind the scenes, there's no JSON property backing this. Its value is computed on demand by querying a view that indexes the `artist` property of every album document. (If this turns out to be too slow we can cache the value until the view index changes.) 

Caveats:

* The property value is **recomputed on every call**. 
* These properties are **not KV-observable** (that would be great to add later, but it would require something like a LiveQuery, and would have to be carefully optimized to be scaleable.)

To define a property like this, you have to do three things:
1. Define the [dynamic] property itself, with type `NSArray*`.
2. Implement a class method +<i>property</i>`ItemClass` that returns the class that has the inverse relation, i.e. the class of the items in the array to be returned.
3. Implement a class method +<i>property</i>`InverseRelation` that returns a string naming the relation property in the other class that should be queried.

In the example:

```
@interface Artist
    @property (readonly) NSArray* albums;
    ...

@implementation Artist
    ...
    @dynamic albums;
    + (Class) albumsItemClass { return [Album class]; }
    + (NSString*) albumsInverseRelation { return @"artist"; }
    ...
```

snej:
* -[CBLModel findInverseOfRelation:fromClass:] explicitly computes an
  inverse relation.
* An array-of-models-valued property can be declared as a computed
  inverse relation by overriding +inverseRelationForArrayProperty or by
  implementing +{property}InverseRelation.
  
See the doc-comments and issue #606 for more details.

Fixes #606

### DIFF
--- a/Source/API/CBLModel+Properties.m
+++ b/Source/API/CBLModel+Properties.m
@@ -182,6 +182,18 @@ static ValueConverter arrayValueConverter(ValueConverter itemConverter) {
 }
 
 
++ (BOOL) hasRelation: (NSString*)relation {
+    objc_property_t property = class_getProperty(self, relation.UTF8String);
+    if (!property)
+        return NO;
+    char* dyn = property_copyAttributeValue(property, "D");
+    if (!dyn)
+        return NO;
+    free(dyn);
+    return YES;
+}
+
+
 #pragma mark - DYNAMIC METHOD GENERATORS:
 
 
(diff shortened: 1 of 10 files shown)


pasin:
This is like magic!

tleyden:
@pasin we should add this to the todolite example app if possible

Issue #543, opened 2014-12-03
Replicators should have independent cookie storage


snej:
By default every NSURLConnection in a process shares the same cookies. Since we do Facebook and Persona login using sessions/cookies, that means that all replicators effectively share a single login. This causes two problems:

* An app can't switch the FB/Persona user it's authenticating as, because the old cookie is still present.
* You can't have two replications to the same host that authenticate as different FB/Persona users.

The fix, I think, is to create a separate NSHTTPCookieStore for each CBL_Replicator and have its NSURLConnections use that instead of the shared one.

SebastianBO:
NSHTTPCookieStorage *cookieJar = NSHTTPCookieStorage.sharedHTTPCookieStorage;
    for (NSHTTPCookie *aCookie in cookieJar.cookies) {
        if ([aCookie.name  isEqual: @"SyncGatewaySession"]) {
            [cookieJar deleteCookie:aCookie];
        }
Works for now i guess? Best Regards

PR #609, opened 2015-02-25
Feature/issue #543 Implement a cookie store per replicator

pasin:
Implement CBLCookieStorage:
* Stores cookies in the local checkpoint document. Each storage is referenced by a unique store key given when initializing the cookie storage instance. When integrating with the CBL_Replicator, the unique store key will be remoteCheckpointDocID.
* Implement the most of the feature sets of NSHTTPCookieStorage except the following 2 features: NSHTTPCookieAcceptPolicyOnlyFromMainDocumentDomain policy and setCookies:forURL:mainDocumentURL: method.
* Simply handle multiple thread access by using synchronized directive.
* Include a Unit Test file named CookieStorage_tests.m.

Replicator Integration:
* CBLRemoteRequest now has a cookieStorage property. If the cookie storage is set, the remote request object will get the cookies from the specified storage and set the cookies into the request header accordingly.
* CBLChangeTracker now has a cookieStorage property. When the cookie storage is set, its CBLSocketChangeTracker and CBLWebSocketChangeTracker subclass will use the cookies from the storage and set to request header accordingly.
* Add _pendingCookies to CBLReplicator to handle set and delete cookies before starting the replicator. Once the replicator is started, the cookie objects in _pendingCookies will be added to the cookie storage and the deleted cookie named in _pendingCookies will be used to delete the cookies from the storage.
* Manually tested the storage and integration with ToDoLite-iOS with both socket and websocket change tracker.

Reference issue #543

### DIFF
--- a/Source/CBLCookieStorage.h
+++ b/Source/CBLCookieStorage.h
@@ -10,6 +10,14 @@
 
 @class CBLDatabase;
 
+/** NSNotification posted when the cookies stored in the CBLCookieStorage instance have changed. 
+    The notification does not contain a userInfo dictionary */
+extern NSString* const CBLCookieStorageCookiesChangedNotification;
+
+/** NSNotification posted when the acceptance policy of the CBLCookieStorage instance has changed. 
+    The notification does not contain a userInfo dictionary. */
+extern NSString* const CBLCookieStorageAcceptPolicyChangedNotification;
+
 @interface CBLCookieStorage : NSObject
 
 /** All cookies that haven't been expired. */
@@ -29,11 +37,14 @@
 /** Returns an array of cookies orted according to a given set of sort descriptors. */
 - (NSArray*) sortedCookiesUsingDescriptors: (NSArray*)sortOrder;
 
-/** Stores a cookie in the cookie storage */
+/** Stores a cookie in the cookie storage. */
 - (void) setCookie: (NSHTTPCookie*)aCookie;
 
 /** Deletes a specified cookie from the cookie storage. */
 - (void) deleteCookie: (NSHTTPCookie*)aCookie;
+
+/** Deletes all cookies by name case-sensitively. */
+- (void) deleteCookiesNamed: (NSString*)name;
 
 /** Deletes all cookies from the cookie storage. */
 - (void) deleteAllCookies;
--- a/Source/CBLCookieStorage.h
+++ b/Source/CBLCookieStorage.h
@@ -38,7 +38,13 @@
 /** Deletes all cookies from the cookie storage. */
 - (void) deleteAllCookies;
 
-/** Close the storage, clear reference to the database */
-- (void) close;
+@end
+
+
+@interface CBLCookieStorage (NSURLRequestResponse)
+
+- (void) addCookieHeaderForRequest: (NSMutableURLRequest*)request;
+
+- (void) setCookieForResponse: (NSHTTPURLResponse*)response;
 
 @end
--- a/Source/CBLCookieStorage.h
+++ b/Source/CBLCookieStorage.h
@@ -1,0 +1,44 @@
+//
+//  CBLCookieStorage.h
+//  CouchbaseLite
+//
+//  Created by Pasin Suriyentrakorn on 2/18/15.
+//  Copyright (c) 2015 Couchbase, Inc. All rights reserved.
+//
+
+#import <Foundation/Foundation.h>
+
+@class CBLDatabase;
+
+@interface CBLCookieStorage : NSObject
+
+/** All cookies that haven't been expired. */
+@property(readonly, copy) NSArray* cookies;
+
+/** The cookie storage’s cookie accept policy. 
+    Currently support NSHTTPCookieAcceptPolicyAlways (Default) and NSHTTPCookieAcceptPolicyNever. */
+@property NSHTTPCookieAcceptPolicy cookieAcceptPolicy;
+
+/** Creates a cookie storage that will store cookies inside the given database as a local document 
+    referenced by a unique storage key. */
+- (instancetype) initWithDB: (CBLDatabase*)db storageKey: (NSString*)storageKey;
+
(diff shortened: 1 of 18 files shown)


Source/CBLCookieStorage.m
```diff
@@ -0,0 +1,424 @@
...
+        NSHTTPCookie* cookie = (NSHTTPCookie*)obj;
+        if (!cookie.sessionOnly) {
+            NSDictionary *props = [self JSONDocumentFromCookieProperties: cookie.properties];
+            [cookies addObject: props];
+        }
+    }];
+
+    NSString* key = [self localDocKey];
+
+    NSLog(@"KEYYYYYY: %@", key);
```
snej: Leftover temporary logging!

Source/CBLCookieStorage.h
```diff
@@ -0,0 +1,61 @@
...
+
+/** Deletes all cookies from the cookie storage. */
+- (void) deleteAllCookies;
+
+@end
+
+
+@interface CBLCookieStorage (NSURLRequestResponse)
+
+- (void) addCookieHeaderForRequest: (NSMutableURLRequest*)request;
```
snej: I think you mean _To_ Request?

Source/CBLCookieStorage.h
```diff
@@ -0,0 +1,61 @@
...
+
+- (void) setCookieForResponse: (NSHTTPURLResponse*)response;
```
snej: Similarly, I think this should probably be _From_ Response.

Source/CBLCookieStorage.m
```diff
@@ -0,0 +1,424 @@
...
+
+
+NSString* const CBLCookieStorageCookiesChangedNotification = @"CookieStorageCookiesChanged";
+NSString* const CBLCookieStorageAcceptPolicyChangedNotification = @"CookieStorageAcceptPolicyChanged";
+
+#define kLocalDocKeyPrefix @"cbl_cookie_storage"
+#define kLocalDocCookiesKey @"cookies"
+
+@interface CBLCookieStorage ()
+- (NSString*) localDocKey;
```
snej: FYI, this isn't necessary anymore. The Obj-C compiler doesn't require forward declarations of methods.

Source/CBLCookieStorage.h
```diff
@@ -0,0 +1,61 @@
...
+
+@class CBLDatabase;
+
+/** NSNotification posted when the cookies stored in the CBLCookieStorage instance have changed. 
+    The notification does not contain a userInfo dictionary */
+extern NSString* const CBLCookieStorageCookiesChangedNotification;
+
+/** NSNotification posted when the acceptance policy of the CBLCookieStorage instance has changed. 
+    The notification does not contain a userInfo dictionary. */
+extern NSString* const CBLCookieStorageAcceptPolicyChangedNotification;
```
snej: Is this needed for anything? I couldn't find any code in the PR that observes it.

Source/CBLCookieStorage.m
```diff
@@ -0,0 +1,424 @@
...
+
+- (void) setCookieAcceptPolicy: (NSHTTPCookieAcceptPolicy)cookieAcceptPolicy {
+    @synchronized(self) {
+        if (cookieAcceptPolicy == NSHTTPCookieAcceptPolicyOnlyFromMainDocumentDomain)
+            Warn(@"%@: Currently NSHTTPCookieAcceptPolicyOnlyFromMainDocumentDomain \
+                 is not supported.", self);
+
+        if (_cookieAcceptPolicy != cookieAcceptPolicy) {
+            _cookieAcceptPolicy = cookieAcceptPolicy;
+            [[NSNotificationCenter defaultCenter] postNotificationName: CBLCookieStorageAcceptPolicyChangedNotification
```
snej: If the notification isn't needed, you can just `@synthesize` this property instead of writing accessors.

Source/CBLCookieStorage.m
```diff
@@ -0,0 +1,424 @@
...
+    @synchronized(self) {
+        return _cookieAcceptPolicy;
+    }
+}
+
+
+- (NSArray*)cookies {
+    @synchronized(self) {
+        NSMutableArray *cookies = [NSMutableArray array];
+        [_cookies enumerateObjectsUsingBlock:^(id obj, NSUInteger idx, BOOL* stop) {
```
snej: Curious why you're using this enumeration method? A for...in... loop is clearer and more efficient. (And the `-[NSArray my_map:]` method from CollectionUtils.h is ideal for doing this sort of thing; it would make this almost a one-liner.)

Source/CBLCookieStorage.h
pasin: We are not using this policy. It's just a parity feature to NSHTTPCookieStorage.

Source/CBLCookieStorage.m
pasin: I still get used to declaring these methods. Will remove them.

Source/CBLCookieStorage.m
pasin: Honestly no good reasons for this. I like my_map: and will change to use that.

Source/CBLCookieStorage.m
pasin: Umm...I thought I have removed all of these.

pasin:
@snej I push the fix according to your review.

Source/CBLCookieStorage.m
pasin: I removed the CBLCookieStorageAcceptPolicyChangedNotification notification but I still wanna keep the warning though.

pasin:
@snej I forgot to mention that we also need to make a small change to WebSocketHTTPLogic.m in WebSockets-Cocoa so that we can disable auto-using the shared cookie storage. 

Here is the PR:
https://github.com/couchbaselabs/WebSockets-Cocoa/pull/2

PR #614, opened 2015-03-01
iOS Framework output dir not optimal for subprojects

laptobbe:
Changed the output dir of the project. I am using it as a submodule and linking to the build product. With this it works to integrate with both ios and osx targets in the same project. This might break something else in the distribution if we are dependent on the path. I have never distributed the project so please check that before you merge this PR.


snej:
Yeah, sorry, that's going to totally break our build. The binaries we build are fat (device + simulator) and need to go into a common location regardless of which target device was selected at build time.

Issue #613, opened 2015-03-01
-[CBL_Puller pullBulkRevisions:] crash


nl:
This crash happened during a pull replication from the gateway in CBL 1.0.4.
https://gist.github.com/nl/6ebd699c515d0078b2b6

snej:
```
__32-[CBL_Puller pullBulkRevisions:]_block_invoke274 + 425
-[CBLRemoteRequest respondWithResult:error:] + 214
-[CBLRemoteRequest connection:didFailWithError:] + 490
-[CBLRemoteRequest cancelWithStatus:] + 203
-[CBLRemoteRequest connection:didReceiveResponse:] + 731
-[CBLBulkDownloader connection:didReceiveResponse:] + 495
__65-[NSURLConnectionInternal _withConnectionAndDelegate:onlyActive:]_block_invoke + 69
```

pasin:
As the crash happens inside the block, the crash log doesn't reveal exactly where the crash happened. I tried to reproduce the crash myself by simulating an error based on your log in several ways that I can think of but I still have no luck.  

Can you help trying to make the app crash again with zombie enabled? or at least, can you capture where the XCode highlights when the app crashes?

pasin:
Wait! I think I was able to reproduce it. To reproduce it, the error code needs to be one of the permanent errors which will causes the stop method to get called when setting the error. I hope this is the same scenario that you encountered. It would be great if you can let me the error code that you got when the app crashed.

snej:
Well, we know it's 425 bytes into the block. If you check out the 1.0.4 tag and disassemble the optimized x86-64 version of CBLPuller, you can look at what's at that offset past `_block_invoke274`.

pasin:
Thanks Jens for pointing this out. Reading the assemble code now, just disassemble the binary and it the offset (+425) is at same line that I got the crash.

```
00000000000167b3         mov        rsi, qword [ds:0x266a20]                    ; @selector(setChangesProcessed:), argument "selector" for method imp___stubs__objc_msgSend
00000000000167ba         mov        rcx, qword [ss:rbp+var_68]
00000000000167be         mov        rdi, rcx                                    ; argument "instance" for method imp___stubs__objc_msgSend
00000000000167c1         mov        rdx, rax
00000000000167c4         call       imp___stubs__objc_msgSend
00000000000167c9         mov        rax, qword [ss:rbp+var_28]
[...]

PR #616, opened 2015-03-03
Fix CBLPuller crash (#613)

pasin:
If a permanent error occurrs during bulk downloading, the completion block will get called twice, one from the error itself and the other one from cancelling the request when setting the error (strongSelf.error == error).

To prevent the above to happen, move the removeRequest to the top of the block so that the request will not get cancelled.

#613


snej:
Good fix. Have you checked whether this problem occurs anywhere else?

pasin:
I have checked only CBLPuller and that is the only place that I can see the problem. Mostly we wrapped remote request calls in sendAsyncRequest method which is safe from that issue.

PR #633, opened 2015-03-19
Try to connect when rechability reports not-reachable

pasin:
- Attempt to connect to the remote url when the reachability reports not-reachable. If an offline error returns, bring the replicator offline. Otherwise, bring the replicator online.

- Used request timeout of 10 seconds.

- Added autoRetry parameter to CBLRemoteRequest to be able to enable/disable retry feature; set auto retry to NO when trying to check if the server is reachable nor not.

- Extended NSURLErrorCannotConnectToHost, NSURLErrorCannotFindHost, and NSURLErrorNetworkConnectionLost to the CBLMisc.CBLIsOfflineError method.

#536

### DIFF
--- a/Source/CBLMisc.m
+++ b/Source/CBLMisc.m
@@ -299,9 +299,6 @@ BOOL CBLIsOfflineError( NSError* error ) {
     if ($equal(domain, NSURLErrorDomain))
         return code == NSURLErrorDNSLookupFailed
             || code == NSURLErrorNotConnectedToInternet
-            || code == NSURLErrorCannotConnectToHost
-            || code == NSURLErrorCannotFindHost
-            || code == NSURLErrorNetworkConnectionLost
 #ifndef GNUSTEP
             || code == NSURLErrorInternationalRoamingOff
 #endif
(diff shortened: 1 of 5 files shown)


Source/CBLMisc.m
```diff
@@ -299,6 +299,9 @@ BOOL CBLIsOfflineError( NSError* error ) {
     if ($equal(domain, NSURLErrorDomain))
         return code == NSURLErrorDNSLookupFailed
             || code == NSURLErrorNotConnectedToInternet
+            || code == NSURLErrorCannotConnectToHost
+            || code == NSURLErrorCannotFindHost
+            || code == NSURLErrorNetworkConnectionLost
```
snej: These error codes don't mean the device is offline, just that the server isn't reachable. I don't think it would be a good idea to have the replicator go offline if, say, the server URL has the wrong port number; this should be reported as an error.

Source/CBL_Replicator.m
```diff
@@ -476,32 +480,73 @@ - (void) reachabilityChanged: (CBLReachability*)host {
...
         [self goOnline];
-    else if (host.reachabilityKnown || _suspended)
+    else if (_suspended || !allowed)
         [self goOffline];
+    else if (host.reachabilityKnown)
+        [self mayGoOffline];
 }
 
+
+- (void) mayGoOffline {
```
snej: The name of this method is confusing to me ... Maybe make it -checkIfOffline? Or something like that?

pasin:
@snej I push changes per code review and discussion. Please review it again when you have time. Thanks.

Issue #624, opened 2015-03-12
CBL (master) crash in bg_replicationProgressChanged


nl:
This crash happens over and over using 789d0d899a2bed9c6e3e21b81af5b633dee267ee
https://gist.github.com/nl/77c6d376af2bbab9635a

snej:
```
Thread 3 Crashed:: CouchbaseLite
0   libobjc.A.dylib               	0x00007fff868b20dd objc_msgSend + 29
1   com.couchbase.CouchbaseLite   	0x000000010e095847 -[CBLReplication bg_replicationProgressChanged:] + 120
2   com.apple.CoreFoundation      	0x00007fff85824b4c __CFNOTIFICATIONCENTER_IS_CALLING_OUT_TO_AN_OBSERVER__ + 12
```

Likely caused by cfaaf58f1d40 or f80d5887701493b3f3c80359ccabb1a1d51673d4, which made changes to CBLReplication to change the way cookies are managed.

pasin:
@nl Sorry for the crash. Did this happen when a one-time replicator finished replicating or stopped? Could you give me a little more context about the scenario when the crash happened?

pasin:
From the crash log, the crash happened when bg_setReplicator: method was called. After talking with Jens, look like the CBLReplicator itself got released when bg_replicationProgressChanged: notification is called.

```
Application Specific Information:
objc_msgSend() selector name: bg_setReplicator:
```

pasin:
@nl can you pull a recent update from the master and see if the fixed solved your issue. Feel free to reopen the issue if the fix doesn't solve your issue.

nl:
Yep, that fixed it!

nl:
Or maybe it just turned it into another crash. https://gist.github.com/nl/9af6cd20d79e56e03349
This makes me think that this crash might be related to https://github.com/couchbase/couchbase-lite-ios/issues/592.
Reopening both.

nl:
Well I can't reopen issues I guess, but please have a look at that other one for me :)

nl:
It also seems related to that 2013 issue I opened.
https://github.com/couchbase/couchbase-lite-ios/issues/220

nl:
Another occurence of this. http://cl.ly/image/2F1a2x1l0M1m https://gist.github.com/nl/b1b29b53f77ade6dd6e6

pasin:
[...]

PR #634, opened 2015-03-24
Fix crash due to notification race condition in CBLReplication

pasin:
- Ensure to remove the progress change notification in the replication thread before forgetting the CBLReplication self object.

- Remove previous crash fix that added retainSelf in the bg_updateProgress method.

#624

### DIFF
--- a/Source/API/CBLReplication.m
+++ b/Source/API/CBLReplication.m
@@ -306,7 +306,7 @@ NSString* const kCBLReplicationChangeNotification = @"CBLReplicationChange";
         [bgReplicator stop];
     }];
     _started = NO;
-    [_database forgetReplication: self];
+    [self forgetReplication];
 }
 
 
@@ -343,7 +343,7 @@ NSString* const kCBLReplicationChangeNotification = @"CBLReplicationChange";
         return;
     if (status == kCBLReplicationStopped) {
         _started = NO;
-        [_database forgetReplication: self];
+        [self forgetReplication];
     }
 
     _pendingDocIDs = nil; // forget cached IDs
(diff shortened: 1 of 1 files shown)


Source/API/CBLReplication.m
```diff
@@ -306,7 +306,7 @@ - (void) stop {
         [bgReplicator stop];
     }];
```
snej: Could you remove self as an observer right here, since you're already on the bg thread, and then not have to make another round-trip in -forgetReplication?

Source/API/CBLReplication.m
```diff
@@ -343,7 +343,7 @@ - (void) updateStatus: (CBLReplicationStatus)status
         return;
     if (status == kCBLReplicationStopped) {
         _started = NO;
-        [_database forgetReplication: self];
+        [self forgetReplication];
```
snej: I don't think removing the observer here is needed, because this is invoked by `_bg_updateProgress`, which when the replicator stops calls `_bg_setReplicator:nil`, which removes the observer.

Source/API/CBLReplication.m
```diff
@@ -387,6 +387,21 @@ - (void) updateStatus: (CBLReplicationStatus)status
     }
 }
 
+- (void)forgetReplication {
```
snej: So because of the above two notes, I don't think this call to the bg thread is needed. Or did I overlook something?

Source/API/CBLReplication.m
pasin: We need to removing the observer here because the updateStatus method is actually called on the database thread. For example If I manually put some sleep as below, it will certainly crash.

In bg_updateProgress method:
```
__weak CBLReplication *weakSelf = self;
    [_database doAsync: ^{
        CBLReplication *strongSelf = weakSelf;
        [strongSelf updateStatus: status error: error processed: changes ofTotal: total
                          serverCert: serverCert];
        cfrelease(serverCert);
    }];
    
    if (status == kCBLReplicationStopped) {
        sleep(1);
        [self bg_setReplicator: nil];
    }
```

Another way to fix this would be moving if (status == kCBLReplicationStopped) block up before [_database doAsync: ]. What do you think?

Source/API/CBLReplication.m
pasin: Yes.

Source/API/CBLReplication.m
snej: Oh, I see. Yes, moving that block up above the async call sounds like a good fix.

pasin:
@snej I put up changes per your comment. Please review it whenever you have time. Thanks!

snej:
Crossing my fingers that this will be the fix!

pasin:
It's gotta be :)

PR #643, opened 2015-04-01
API: Added kCBLBySequence to CBLAllDocsOptions (aka "changes feed")

snej:
In this mode an all-docs query returns docs in order by sequence number.
The startKey and endKey are interpreted as sequence numbers.
Note: Descending order isn't yet supported.


snej:
FYI @borrrden @hideki

Issue #235, opened 2014-01-05
Error message passed to `rejectWithMessage:` in validation block does not propagate to out-error


p2:
I have a validation block that rejects saves with a custom message. However when I call the model's `save:` method and the validation block rejects the save, the returned `[error localizedDescription]` always says **403 forbidden**.

Validation block:
```
[_database setValidationNamed:@"ownerValidation"
    asBlock:^(CBLRevision *newRevision, id<CBLValidationContext> context) {
        if (!newRevision.isDeletion) {
            NSString *type = newRevision.properties[@"type"];
            if ([type isEqualToString:@"result"]
                || [type isEqualToString:@"vital"]
                || [type isEqualToString:@"media"]) {
                if (!newRevision.properties[@"owner"]) {
                    NSString *message = [NSString stringWithFormat:@"Documents of type \"%@\" must have an owner", type];
                    [context rejectWithMessage:message];
                }
            }
        }
}];
```

When provoking saving without an owner, as required by the validation block:
```
if (![image save:&bError]) {
    NSLog(@"Failed to save image: %@", [bError localizedDescription]);
}
```

Logged error message:
> Failed to save image: 403 forbidden

Maybe this is intended, I would find it useful to see my message in the returned error.

PR #644, opened 2015-04-02
Propagate validation rejection message

pasin:
- Propagate rejection message inside an NSError object. This requires to refactor several method to include out NSError object parameters.

- For Router, the result will be sent as part of the response body, for example:
{"status": 403, "error": "forbidden", "reason": "The document cannot be updated!"}

#235

### DIFF
--- a/Source/CBLDatabase+Attachments.m
+++ b/Source/CBLDatabase+Attachments.m
@@ -379,16 +379,20 @@ static UInt64 smallestLength(NSDictionary* attachment) {
 /** Replaces or removes a single attachment in a document, by saving a new revision whose only
     change is the value of the attachment. */
 - (CBL_Revision*) updateAttachment: (NSString*)filename
-                            body: (CBL_BlobStoreWriter*)body
-                            type: (NSString*)contentType
-                        encoding: (CBLAttachmentEncoding)encoding
-                         ofDocID: (NSString*)docID
-                           revID: (NSString*)oldRevID
-                          status: (CBLStatus*)outStatus
+                              body: (CBL_BlobStoreWriter*)body
+                              type: (NSString*)contentType
+                          encoding: (CBLAttachmentEncoding)encoding
+                           ofDocID: (NSString*)docID
+                             revID: (NSString*)oldRevID
+                            status: (CBLStatus*)outStatus
+                             error: (NSError**)outError
 {
     *outStatus = kCBLStatusBadAttachment;
-    if (filename.length == 0 || (body && !contentType) || (oldRevID && !docID) || (body && !docID))
+    if (filename.length == 0 || (body && !contentType) || (oldRevID && !docID) || (body && !docID)) {
+        if (outError)
+            *outError = CBLStatusToNSError(*outStatus, nil);
         return nil;
+    }
 
     CBL_MutableRevision* oldRev = [[CBL_MutableRevision alloc] initWithDocID: docID
                                                                        revID: oldRevID
@@ -399,9 +403,11 @@ static UInt64 smallestLength(NSDictionary* attachment) {
         if (CBLStatusIsError(*outStatus)) {
             if (*outStatus == kCBLStatusNotFound
                 && [self getDocumentWithID: docID revisionID: nil withBody: NO
-                                        status: outStatus] != nil) {
+                                    status: outStatus] != nil) {
                 *outStatus = kCBLStatusConflict;   // if some other revision exists, it's a conflict
             }
+            if (outError)
+                *outError = CBLStatusToNSError(*outStatus, nil);
             return nil;
         }
     } else {
@@ -426,6 +432,8 @@ static UInt64 smallestLength(NSDictionary* attachment) {
     } else {
         if (oldRevID && !attachments[filename]) {
             *outStatus = kCBLStatusAttachmentNotFound;
+            if (outError)
+                *outError = CBLStatusToNSError(*outStatus, nil);
             return nil;
         }
         [attachments removeObjectForKey: filename];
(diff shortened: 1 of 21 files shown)


Source/CBLDatabase+Attachments.m
```diff
@@ -424,10 +432,16 @@ - (CBL_Revision*) updateAttachment: (NSString*)filename
...
     // Store a new revision with the updated _attachments:
-    CBL_Revision* newRev = [self putRevision: oldRev prevRevisionID: oldRevID
-                             allowConflict: NO status: outStatus];
+    CBL_Revision* newRev = [self putRevision: oldRev prevRevisionID: oldRevID allowConflict: NO
+                                      status: outStatus error: outError];
     if (!body && *outStatus == kCBLStatusCreated)
         *outStatus = kCBLStatusOK;
+
+    if (CBLStatusIsError(*outStatus)) {
+        if (outError && !*outError)
```
snej: `*outError` is unknown at this point since we haven't initialized it — the caller didn't necessarily set it to nil. I think you can just take out `&& !*outError`.

Source/CBL_ForestDBStorage.mm
```diff
@@ -992,8 +998,11 @@ - (CBL_Revision*) addDocID: (NSString*)inDocID
         return (CBLStatus)status;
     }];
 
-    if (CBLStatusIsError(*outStatus))
+    if (CBLStatusIsError(*outStatus)) {
+        if (outError && !*outError)
```
snej: Same issue — `!*outError` isn't valid.

Source/CBL_ForestDBStorage.mm
```diff
@@ -1063,6 +1079,11 @@ - (CBLStatus) forceInsert: (CBL_Revision*)inRev
 
     if (change)
         [_delegate databaseStorageChanged: change];
+
+    if (CBLStatusIsError(status)) {
+        if (outError && !*outError)
```
snej: `!*outError` isn't valid.

Source/CBL_Router+Handlers.m
```diff
@@ -769,15 +779,22 @@ - (CBLStatus) update: (CBLDatabase*)db
     else
         *outRev = [db putRevision: rev prevRevisionID: prevRevID
                     allowConflict: allowConflict
-                           status: &status];
+                           status: &status
+                            error: outError];
+
+    if (CBLStatusIsError(status)) {
+        if (outError && !*outError)
```
snej: `!*outError` isn't valid.

Source/CBLDatabase+Attachments.m
snej: Ah, just noticed that `outError` is passed to the `putRevision:` call above. Hm. I think that means that `putRevision:` should have already put an NSError into it and this method shouldn't touch it.

Source/CBL_ForestDBStorage.mm
snej: Oog, this is messy. I think you'll need to initialize `*outError` to nil at the top of the method.

Source/CBL_ForestDBStorage.mm
snej: As above, I think you'll need to initialize `*outError` to nil at the top.

Source/CBL_Router+Handlers.m
```diff
@@ -293,11 +296,17 @@ - (CBLStatus) do_POST_bulk_docs: (CBLDatabase*)db {
...
+                        _response.statusReason = error.localizedFailureReason;
                     return status;  // all_or_nothing backs out if there's any error
                 } else {
-                    NSString* error = nil;
-                    status = CBLStatusToHTTPStatus(status, &error);
-                    result = $dict({@"id", docID}, {@"error", error}, {@"status", @(status)});
+                    NSString* errorMessage = nil;
+                    status = CBLStatusToHTTPStatus(status, &errorMessage);
+                    NSString* reason = error.localizedFailureReason;
+                    if (reason)
```
snej: You don't need to check `reason` — `$dict` ignores nil values (unlike `@{...}`.) So you can just use the first assignment line (307).

Source/CBL_Router+Handlers.m
snej: You can put the assignment to `*outError` right after the `-putLocalRevision:` call, since that's the only one that would set `status` but not `*outError`.

Source/CBL_SQLiteStorage.m
```diff
@@ -1600,8 +1603,11 @@ - (CBL_Revision*) addDocID: (NSString*)inDocID
         return deleting ? kCBLStatusOK : kCBLStatusCreated;
     }];
     
-    if (CBLStatusIsError(*outStatus)) 
+    if (CBLStatusIsError(*outStatus)) {
+        if (outError && !*outError)
```
snej: Initialize `*outErrror` to nil at the top.

Source/CBL_SQLiteStorage.m
```diff
@@ -1737,7 +1744,11 @@ - (CBLStatus) forceInsert: (CBL_Revision*)inRev
                                                               winningRevisionID: winningRevID
                                                                      inConflict: inConflict
                                                                          source: source]];
+    } else {
+        if (outError && !*outError)
```
snej: Initialize `*outErrror` to nil at the top.

pasin:
@snej I fixed per your code review and make sure that out error value has value set. Could you please re-review when you have time?

Source/CBL_ForestDBStorage.mm
pasin: !*outError check is required here to not overriding the validation error from line 972.

Source/CBL_ForestDBStorage.mm
pasin: !*outError check is required here to not overriding the validation error from line 1060.

pasin:
@snej I rebased the branch with the current master (I tried both merge and rebase, and rebase result is more preferable as we have a lot of new commits in the master branch). 

The last commit https://github.com/couchbase/couchbase-lite-ios/commit/2de5f143c2348c31332970e99c14e429d621a2df contains the fix according to your comment.

snej:
Since there was a lot of stuff changed, I pulled the branch locally to review and test. Then I rebased the commits into one, and added some changes to the error utilities in CBLStatus.m to make things cleaner. I've pushed this all on the master branch.

Issue #649, opened 2015-04-08
Attachment crash (CBL_BlobStoreWriter's initializer failed?)


nl:
Commit b4850d3a1f4bd944eb9db8fe20fb4a409ebeff82
https://gist.github.com/nl/39b3d5b2bbbbd4533d71

nl:
NB: The data I'm trying to attach is _NOT_ nil.

snej:
```
*** Terminating app due to uncaught exception 'NSInvalidArgumentException', 
reason: '*** setObjectForKey: object cannot be nil (key: preview.jpg)'
terminating with uncaught exception of type NSException
abort() called
 
Application Specific Backtrace 1:
0   CoreFoundation                      0x00007fff8456766c __exceptionPreprocess + 172
1   libobjc.A.dylib                     0x00007fff8d10b76e objc_exception_throw + 43
2   CoreFoundation                      0x00007fff8444f876 -[__NSDictionaryM setObject:forKey:] + 1174
3   CouchbaseLite                       0x000000010eff1e62 -[CBLMultipartDocumentReader startedPart:] + 391
4   CouchbaseLite                       0x000000010efeb90d -[CBLMultipartReader appendData:] + 800
5   CouchbaseLite                       0x000000010eff164c -[CBLMultipartDocumentReader appendData:] + 68
6   CouchbaseLite                       0x000000010f00b804 -[CBLBulkDownloader appendToPart:] + 58
7   CouchbaseLite                       0x000000010efeb521 -[CBLMultipartReader appendAndTrimBuffer] + 151
8   CouchbaseLite                       0x000000010efeba34 -[CBLMultipartReader appendData:] + 1095
9   CouchbaseLite                       0x000000010f00b572 -[CBLBulkDownloader connection:didReceiveData:] + 130
```

Looks like the line that triggered it is CBLMultipartDocumentReader.m:259; `_curAttachment` must be nil:
```objc
                _attachmentsByName[name] = _curAttachment;
```

snej:
The only way for `_curAttachment` to be nil is if CBL_BlobStoreWriter's initializer failed and returned nil, and I think the only way for that to happen is if it can't create a temporary file.

Nico, is it possible that the device ran out of storage space?

nl:
I checked the Mac this is running on, and the storage space is fine.

snej:
Sorry, I keep forgetting you're writing a Mac app!

[...]

PR #657, opened 2015-04-13
Remove NSDataReadingMappedIfSafe when reading attachment content

pasin:
NSDataReadingMappedIfSafe option can keep the file descriptors of the attachments open if the client application keeps holding the NSData object returned by CBLAttachment.content. For a very big attachment, alternatively, users can call contentURL and read the NSData with the NSDataReadingMappedIfSafe option themselves.

Changes:
1. Remove NSDataReadingMappedIfSafe option from  CBLAttachment.content method.
2. Still keep NSDataReadingMappedIfSafe option in the CBL_BlobStore.blobForKey() method as it gets used internally and no retaining NSData returned by the method.
3. Still keep NSDataReadingMappedIfSafe option in the CBL_BlobStore.blobData() method as the method doesn't get called by any codes.

#649

### DIFF
--- a/Source/API/CBLAttachment.m
+++ b/Source/API/CBLAttachment.m
@@ -118,7 +118,7 @@
             return _body;
         else if ([_body isKindOfClass: [NSURL class]] && [_body isFileURL]) {
             return [NSData dataWithContentsOfURL: _body
-                                         options: NSDataReadingMappedIfSafe | NSDataReadingUncached
+                                         options: NSDataReadingUncached
                                            error: nil];
         }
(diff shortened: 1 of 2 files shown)


snej:
This is a worthwhile change, but it's not the important fix because this code path only runs if you add a new attachment and then access the contents of the same CBLAttachment object. The important case that I think @nl is running into is the one in -[CBL_BlobStore blobForKey:], which is the code path for loading an already-saved attachment. Could you add that fix to this PR too, please?

PR #659, opened 2015-04-13
Fix circular reference between CBLUnsavedRevision and CBLAttachment

pasin:
- As CBLUnsavedRevision keeps CBLAttachment objects in an array so setting the revision (self) object to the CBLAttachment objects will cause circular reference memory leak.

- Instead of setting the revision (self) object when adding a CBLAttachment to the CBLUnsavedRevision object, overiding the attachmentNamed: method and creating a new CBLAttachment object and adding a revision object there.

### DIFF
--- a/Source/API/CBLRevision.m
+++ b/Source/API/CBLRevision.m
@@ -371,24 +371,10 @@
     // NOTE: Not setting the revision to the attachment object (attachment.revision = self) as
     // [1] The UnsavedRevision object is not used during save operation
     // [2] Setting the UnsavedRevision object here will cause the circular reference memory leak.
-    // The revision object will be set in the overridden attachmentName: method.
 }
 
 - (void) removeAttachmentNamed: (NSString*)name {
     [self _addAttachment: nil named: name];
 }
-
-- (CBLAttachment*) attachmentNamed: (NSString*)name {
-    CBLAttachment* base = self.attachmentMetadata[name];
-    if (base) {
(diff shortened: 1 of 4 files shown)


Source/API/CBLRevision.m
```diff
@@ -367,12 +367,28 @@ - (void) _addAttachment: (CBLAttachment*)attachment named: (NSString*)name {
...
+    // [2] Setting the UnsavedRevision object here will cause the circular reference memory leak.
+    // The revision object will be set in the overridden attachmentName: method.
 }
 
 - (void) removeAttachmentNamed: (NSString*)name {
     [self _addAttachment: nil named: name];
 }
 
+- (CBLAttachment*) attachmentNamed: (NSString*)name {
+    CBLAttachment* base = self.attachmentMetadata[name];
```
snej: The object might be an NSDictionary not a CBLAttachment, if the attachment exists in the parent revision. (Look at the superclass method.)

snej: I actually don't think you need this override at all. The superclass method looks fine for this case.

pasin: I looked at the super class but as I didn't set the the revision to the attachment object, the super class will return the same CBLAttachment without revision object set.

pasin: For your 1st comment, I will need to handle that case.

pasin:
@snej I updated the code and unit test. Please review them when you have time.

Issue #648, opened 2015-04-08
Dynamic Property has type '@' which is not a known class or protocol issue


24601:
Just updated to the current master build of couchbase-lite-ios from my previous known-good-in-my-app version built on Feb 27 and launching the app causes this error. 

I have an class named VTCase, of course, with a property caseName which is an NSString (Swift project), @NSManaged property, of course, displaying this object in a form and loading that view controller tries to access this property and this is where it croaks. 

Going back to the Feb 27 build of the library fixes the problem.

Swift 1.2 project with Xcode 6.3 6D543q on OS x 10.10.3 (14D130a).

23:54:38.118| WARNING: MYDynamicObject: VTCase.caseName has type @ which is not a known class or protocol
23:54:38.119| WARNING: Dynamic property VTCase.caseName has type '@' unsupported by VTCase
2015-04-07 23:54:38.119 Voltaire[49935:1641226] -[VTCase caseName]: unrecognized selector sent to instance 0x788a72f0

snej:
Let's see. The [last commit to MYDynamicObject](c05cc90541d495dadbcb31d919e4115ed9bca765) was on Feb 19, but it didn't get merged into CBL until March 12 (commit 82ec99c7566448a69cece593e8381136c74cd39c).

That commit affected property type parsing, to handle protocol types, but I'm surprised it would break with a common type like NSString — our unit tests would have caught that. Unless there's something weird about the property type encoded by Swift.

snej:
Fixed — for some reason the property type got encoded as `id` not `NSString`, and that MYDynamicObject change broke support for such properties.

24601:
Jens,

Thank you so much, again! Will pull and rebuild and try again!

Cheers!

Basit

24601:
@snej - Just pulled the latest version from master and built it and still seem to be getting the issue (al beit without the MYDynamicObject warning line):


19:45:35.726| WARNING: Dynamic property VTCase.caseName has type '@' unsupported by VTCase
2015-04-08 19:45:35.727 Voltaire[13066:128226] -[VTCase caseName]: unrecognized selector sent to instance 0x7afdb8b0
(lldb) bt
* thread #1: tid = 0x1f4e2, 0x04caa69a libsystem_kernel.dylib`__pthread_kill + 10, queue = 'com.apple.main-thread', stop reason = signal SIGABRT
    frame #0: 0x04caa69a libsystem_kernel.dylib`__pthread_kill + 10
[...]

PR #669, opened 2015-04-20
Fix for CBL-ios issue 648

24601:
This is my fix for #648. Intentionally excluding NSDate in the NSObject subclass check since it is handled best in the final else catch-all case, and NSObject handing of NSDate is not appropriate/does not work. Maybe more elegant way to implement this as there might be other cases (e.g. other classes) that the coe should be letting the final catch-all else handle using ValueConverter handle (I doubt NSDate is the only such case?), but wanted to avoid significantly changing the structure/flow of this code as it is not originally my code and my understanding of it is not good enough and wanted to leave the least trace possible with this fix, I can imagine reworking this in a few ways to handle this, but it would probably change the method and break other cases, so excluding NSDate subclasses in the NSObject else if I added for now...

https://github.com/couchbase/couchbase-lite-ios/issues/648

### DIFF
--- a/Source/API/CBLModel+Properties.m
+++ b/Source/API/CBLModel+Properties.m
@@ -239,10 +239,6 @@ static ValueConverter arrayValueConverter(ValueConverter itemConverter) {
                 };
             }
         }
-        else if ( [itemClass isSubclassOfClass: [NSObject class]] && ![itemClass isSubclassOfClass:[NSDate class]])
-        {
-            return [super impForGetterOfProperty: property ofClass: propertyClass];
-        }
         else {
             // Typed array of scalar class:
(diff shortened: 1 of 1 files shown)


24601:
To add on to my above comments on the nature of excluding certain classes in the NSObject subclass check (which I think is a bit kludgy), perhaps a cleaner way of accomplishing this is to, in the final catch-all else (in both array item and property code paths) to try value converter first and if converter is nil, then do return [super impForGetterOfProperty: property ofClass: propertyClass];, would do a better job of "deferring to ValueConverter" like I mention above than the current implementation, perhaps? 


Also make updates to ValueConverter to handle more data types "automatic" in this case....

24601:
Implemented the above fix as it is much cleaner and will resubmit this PR.

Issue #672, opened 2015-04-22
Pulled docs with a stub attachment failed to insert


pasin:
The issue was originally reported in CBL Android (https://github.com/couchbase/couchbase-lite-java-core/issues/498).

For CBL iOS, there are 2 bugs related to this issue:
1. CBL iOS have never sent atts_since parameter in _bulk_get requests.
2. In some certain cases that the pulled attachment has stub = true and its immediate parent rev has no attachment info, it will fail to insert the revision with the attachment not found error.

pasin:
Patch to apply from @snej:

Jens Alfke: Basically, if the attachment is a stub, and the parent revision's JSON isn't known, but the attachment exists locally (based on its digest), just leave the stub alone.

```
diff --git a/Source/CBLDatabase+Attachments.m b/Source/CBLDatabase+Attachments.m
index 30ac44d..b2a1b9c 100644
--- a/Source/CBLDatabase+Attachments.m
+++ b/Source/CBLDatabase+Attachments.m
@@ -303,11 +303,19 @@ static UInt64 smallestLength(NSDictionary* attachment) {
         } else if ([attachInfo[@"stub"] isEqual: $true]) {
             // "stub" on an incoming revision means the attachment is the same as in the parent.
             if (!parentAttachments && prevRevID) {
+                CBLStatus status;
                 parentAttachments = [self attachmentsForDocID: rev.docID revID: prevRevID
-                                                       status: outStatus];
+                                                       status: &status];
                 if (!parentAttachments) {
-                    if (*outStatus == kCBLStatusOK || *outStatus == kCBLStatusNotFound)
-                        *outStatus = kCBLStatusBadAttachment;
+                    if (status == kCBLStatusNotFound
+                            && [_attachments hasBlobForKey: attachment.blobKey]) {
+                        // Parent revision's body isn't known (we are probably pulling a rev along
+                        // with its entire history) but it's OK, we have the attachment already
+                        return attachInfo;
+                    }
+                    if (status == kCBLStatusOK || status == kCBLStatusNotFound)
+                        status = kCBLStatusBadAttachment;
+                    *outStatus = status;
                     return nil;
                 }
             }
diff --git a/Source/CBL_SQLiteStorage.m b/Source/CBL_SQLiteStorage.m
index bbf8112..853e4c2 100644
--- a/Source/CBL_SQLiteStorage.m
+++ b/Source/CBL_SQLiteStorage.m
[...]

PR #673, opened 2015-04-23
Fix pulling documents with sub attachments failure

pasin:
Two related issues have been fixed:
1. Correct minRevPos value when stripping the (non modified) attachments. This corrects the performance optimization when pushing documents that have non-modified attachments.
2. Applying a patch from @snej that "if the attachment is a stub, and the parent revision's JSON isn't known, but the attachment exists locally (based on its digest), just leave the stub alone".

#672

### DIFF
--- a/Source/CBLDatabase+Insertion.m
+++ b/Source/CBLDatabase+Insertion.m
@@ -228,8 +228,8 @@
         NSString* prevRevID = history.count >= 2 ? history[1] : nil;
         if (![self processAttachmentsForRevision: updatedRev
                                        prevRevID: prevRevID
-                                          status: &status])
-        return status;
+                                          status: &status] && CBLStatusIsError(status))
+            return status;
         inRev = updatedRev;
     }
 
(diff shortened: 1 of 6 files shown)


Source/CBLDatabase+Insertion.m
```diff
@@ -228,8 +228,8 @@ - (CBLStatus) forceInsert: (CBL_Revision*)inRev
         NSString* prevRevID = history.count >= 2 ? history[1] : nil;
         if (![self processAttachmentsForRevision: updatedRev
                                        prevRevID: prevRevID
-                                          status: &status])
-        return status;
+                                          status: &status] && CBLStatusIsError(status))
```
snej: Why is this necessary? -processAttachmentsForRevision shouldn't return NO unless there's an error. (If it does, we should fix that instead of testing for it here.)

snej: Oh, I see —it's because -mutateAttachments returns NO if there were no changes. So the right fix is to make -processAttachmentsForRevision ignore the return value of -mutateAttachments, and instead return `!CBLStatusIsError(*outStatus)`.

Issue #686, opened 2015-05-01
The bg replication server's database not getting closed


pasin:
The bg server's database of the replicator should be closed as well when calling database.close().  

Maybe when a replication is stopped, we also close the database own by the background's server. This causes the -wal files not getting merged into the database file (the -wal file size > 0 byte).

pasin:
Discussed about solution with @snej, and closing the background database is complicated and has a risk to the release. We will fix this after 1.1.0 release instead. For users to prepare a database file for other platforms to use, users will need to do as follows:

1. Instead of using a shared CBLManager, create a new CBLManager.
2. Create a database (e.g. pull documents from server and so on).
3. Close the manager.
4. All connections to the sqlite database should be closed and the -wal file content should be merged to the main SQLite file.

PR #689, opened 2015-05-01
Fix bg database not getting closed

pasin:
1. Close the background database when the replication stopped.

2. Close the background database when async query is finished.

3. Fixed replication stop method that the updateStatus: method is not get called to update the stopped status when the stop method is called from CBLDatabase's close: method.

Note: I manually verified as of sqlite v.3.8.5 if the WAL mode is enabled, a proper close sqlite database doesn't remove the -wal and -shm file but the -wal file size will be zero bytes.

#686

### DIFF
--- a/Source/API/CBLQuery.m
+++ b/Source/API/CBLQuery.m
@@ -261,6 +261,11 @@
             }
         }
 
+        NSError* error;
+        LogTo(Query, @"%@: Close background database: %@", self, bgdb);
+        if (![bgdb close: &error])
+            Warn(@"Failed to close background database: %@", error);
+
         [_database doAsync: ^{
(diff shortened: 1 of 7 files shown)


snej:
It's not safe to close a database this way, because the calling code doesn't know if something else is using the db and has a reference to it. For example, if there are both a push and a pull replication involving a db, then if the push finishes first and closes the db, the pull will break — it'll either prematurely stop, or just crash. Same thing if a bg query of a database happens while a replication is active.

It's not really safe to close a database except when its CBLManager is being closed. When a foreground CBLManager is closed, it stops its background thread's CBLServer, which closes the CBLManager being used in the background.

Issue #696, opened 2015-05-12
Start-restart replication problem in CBL 1.1.0


ryuzmukhametov:
Hi,

Recently I stumbled one interesting issue in Couchbase Lite framework.
If after starting a replication I try restart it then the replication is stopped.
I got version 1.1.0 (f1aa2cd07433900d8eebc29f8eb8fe1a81df94ca) with ForestDB support.

For reproduce it you can CBL iOS Test App, and in method updateSyncURL change:
[_pull start];
[_push start];
to:
[_pull start];
[_pull restart];
[_push start];
[_push restart];

After these change my databases replications are stopped.

snej:
I know the example above is simplified to make it easy to describe — have you run into this problem in a real-world situation? The -restart method is literally exactly the same as calling -stop and then -start, so I'm wondering if the problem only occurs if you've just started the replication, or if it'll happen at other times too.

pasin:
I think this is a regression bug. In https://github.com/couchbase/couchbase-lite-ios/commit/5438f4727a9713100c1e5814774687042acc272c, I removed line 312 and 313. I thought those 2 lines are redundant as those will be executed again asynchronously when the -updateStatus: method is called.

Basically when the start method is called right after the stop method, the _started doesn't get reset immediately.

ryuzmukhametov:
Regarding how I have got this bug.
In my app I have method 'actualizeAllReplications' which runs by some application notifications.
In this method I check a replication with my server. If I don't have such replication I create it and start.
If I already have this server replication then I restart it. It happened that 'actualizeAllReplications' was called twice immediately from main thread by notification.
So I got the described situation

PR #699, opened 2015-05-13
Fix replicator restart issue

pasin:
- In CBLReplication -stop method, before returning, wait until the background stop notification comes.

- In CBLReplication -restart method, spin the -start method call into the next runloop or the next dispatching if the dispatch queue is used.

#696

### DIFF
--- a/Source/API/CBLReplication.m
+++ b/Source/API/CBLReplication.m
@@ -49,6 +49,7 @@ NSString* const kCBLReplicationChangeNotification = @"CBLReplicationChange";
     bool _started;
     CBL_Replicator* _bg_replicator;       // ONLY used on the server thread
     NSMutableArray* _bg_pendingCookies;
+    NSConditionLock* _stopLock;
 }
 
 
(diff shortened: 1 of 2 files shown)


Source/API/CBLReplication.m
```diff
@@ -537,6 +554,11 @@ - (void) bg_updateProgress {
                           serverCert: serverCert];
         cfrelease(serverCert);
     }];
+
+    if (status == kCBLReplicationStopped) {
+        [_stopLock lockWhenCondition: 0];
```
snej: I think there's a possible race condition, because the variable _stopLock gets modified on the main thread. For example, if the replicator stops on its own, and the app also calls -stop, then it would be possible for the value of `_stopLock` to be nil on the first call but non-nil on the second, which would cause an exception because it's calling -unlock before -lock...

snej:
Looks good! Except — by the naming convention, the ivar should be renamed `_bg_stopLock`. Sorry to nitpick, but it's important to keep that straight.

pasin:
Thanks Jens. I like nitpicking. I have been going back and forth thinking if I should use _bg prefix or some kind of _shared. I will made the change.

pasin:
@snej I updated the code accordingly.

snej:
Hm, yeah, I guess it's different than the other bg ivars because it's also used from the main thread. But the bg is still an indication that it should only be assigned to on the background thread. Thanks for making the change!

PR #707, opened 2015-05-15
Add test22_ReplaceDatabase unit test

pasin:
Added test22_ReplaceDatabase to test -replaceDatabaseNamed: method against the databases from multiple platforms (iOS, Android, .NET) of both 1.0.4 and 1.1.0 version.


hideki:
@pasin, I am curios if CBL iOS v1.1.0 is compatible with CBL .NET v1.0.4 for query?

pasin:
I think yes because the unit test covers query, and all the tests passed.

Issue #718, opened 2015-05-19
Crash when create a document with attachment in one single revision


pasin:
The crash happened because the NSMutableDictionary object was changed to NSDictionary at runtime when setting .properties value to the unsavedRevision. So when trying to add an attachment, .properties dictionary cannot be altered.

```
"NSInvalidArgumentException", "-[__NSDictionaryI setObject:forKeyedSubscript:]: unrecognized selector sent to instance 0x7fb90a64c520"
(
	0   CoreFoundation                      0x0000000111834c65 __exceptionPreprocess + 165
	1   libobjc.A.dylib                     0x00000001114cdbb7 objc_exception_throw + 45
	2   CoreFoundation                      0x000000011183c0ad -[NSObject(NSObject) doesNotRecognizeSelector:] + 205
	3   CoreFoundation                      0x000000011179213c ___forwarding___ + 988
	4   CoreFoundation                      0x0000000111791cd8 _CF_forwarding_prep_0 + 120
	5   CBL Test                            0x000000010eccea3c -[CBLUnsavedRevision _addAttachment:named:] + 316
	6   CBL Test                            0x000000010ecce7b8 -[CBLUnsavedRevision setAttachmentNamed:withContentType:content:] + 184
	7   CBL iOS Unit Tests                  0x0000000119d5cfb9 -[Database_Tests test22_CreateDocWithAttachmentInSingleRevision] + 361
	8   CoreFoundation                      0x000000011172adec __invoking___ + 140
	9   CoreFoundation                      0x000000011172ac42 -[NSInvocation invoke] + 290
	10  XCTest                              0x000000011a18a12a -[XCTestCase invokeTest] + 253
	11  CBL iOS Unit Tests                  0x0000000119ce0665 -[CBLTestCaseWithDB invokeTest] + 69
	12  XCTest                              0x000000011a18a32b -[XCTestCase performTest:] + 150
	13  XCTest                              0x000000011a193be5 -[XCTest run] + 260
	14  XCTest                              0x000000011a18903b -[XCTestSuite performTest:] + 379
	15  XCTest                              0x000000011a193be5 -[XCTest run] + 260
	16  XCTest                              0x000000011a18903b -[XCTestSuite performTest:] + 379
	17  XCTest                              0x000000011a193be5 -[XCTest run] + 260
	18  XCTest                              0x000000011a18903b -[XCTestSuite performTest:] + 379
[...]

PR #719, opened 2015-05-19
Fix UnsavedRevision.properties synthesizing

pasin:
The UnsavedRevision.properties propery is declared as copy and that causes an issue that the new NSMutableDictionary object set to the property is changed to an immutable one at runtime by the SDK. Instead of synthesizing, implementing a pair get/set method with doing mutableCopy instead.

#718

### DIFF
--- a/Source/API/CBLRevision.m
+++ b/Source/API/CBLRevision.m
@@ -288,7 +288,7 @@
     NSMutableDictionary* _properties;
 }
 
-@synthesize parentRevisionID=_parentRevID, properties=_properties;
+@synthesize parentRevisionID=_parentRevID;
 @dynamic isDeletion, userProperties;     // Necessary because this class redeclares them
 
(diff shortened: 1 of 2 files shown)


Source/API/CBLRevision.m
```diff
@@ -308,6 +308,20 @@ - (CBLSavedRevision*) parentRevision {
     return _parentRevID ? [_document revisionWithID: _parentRevID] : nil;
 }
 
+- (void) setProperties: (NSMutableDictionary*)properties {
+    @synchronized(self) { // Atomic property
```
snej: Doesn't actually need to be atomic since our API is explicitly not thread-safe, so you can take out the `@synchronized`...

snej:
BTW it's confirmed that the effect of the `(copy)` modifier on a synthesized setter is correct, or at least as designed. There isn't a feasible way for the compiler to tell whether it should call -mutableCopy instead of -copy. So if you need a mutable copy you should write the setter yourself, as we're doing now :)

PR #721, opened 2015-05-19
Fix error re-adding a document after it's been purged.

evankyle:
Fixed a bug in SQLLiteStorage/_createDocNumericID that should insert or ignore that was returning -1 when it really should return the documents row id. For a unit test to call _createDocNumericID twice and fail correctly I had to add a function to clear the document from the key cache _docIDs.

### DIFF
--- a/Source/CBL_SQLiteStorage.m
+++ b/Source/CBL_SQLiteStorage.m
@@ -677,9 +677,16 @@ static void CBLComputeFTSRank(sqlite3_context *pCtx, int nVal, sqlite3_value **a
 - (SInt64) _createDocNumericID: (NSString*)docID {
     if (![_fmdb executeUpdate: @"INSERT OR IGNORE INTO docs (docid) VALUES (?)", docID])
         return -1;
-    if (_fmdb.changes == 0)
-        return -1;
+    if (_fmdb.changes == 0) {
+      SInt64 result = [self _getDocNumericID:docID];
+      if(result > 0) return result; 
+      return -1;
+    }
     return _fmdb.lastInsertRowId;
+}
+
+-(void) removeKeyFromDocIDCache:(NSString*)docID {
+  [_docIDs removeObjectForKey:docID];
 }
 
 // Registers a docID and returns its numeric row ID in the 'docs' table.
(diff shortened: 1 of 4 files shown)


Source/CBL_SQLiteStorage.m
```diff
@@ -677,11 +677,18 @@ - (SInt64) _getDocNumericID: (NSString*)docID {
 - (SInt64) _createDocNumericID: (NSString*)docID {
     if (![_fmdb executeUpdate: @"INSERT OR IGNORE INTO docs (docid) VALUES (?)", docID])
         return -1;
-    if (_fmdb.changes == 0)
-        return -1;
+    if (_fmdb.changes == 0) {
+      SInt64 result = [self _getDocNumericID:docID];
```
snej: Actually the only change that's needed here is to replace `return -1` with `return 0`. The caller (-createOrGetDocNumericID:) will do the right thing if 0 is returned.

Source/CBL_SQLiteStorage.m
```diff
@@ -677,11 +677,18 @@ - (SInt64) _getDocNumericID: (NSString*)docID {
...
+      if(result > 0) return result; 
+      return -1;
+    }
     return _fmdb.lastInsertRowId;
 }
 
+-(void) removeKeyFromDocIDCache:(NSString*)docID {
```
snej: You don't need to add this method. It would be better to have the unit test close and re-open the database, by calling -reopenTestDB.

snej:
Thanks! I think the fix is even simpler, though. (Looks like when I wrote the code I put in `return -1` when I meant `return 0`.)

Source/CBL_SQLiteStorage.m
evankyle: I figured that was a bit much but needed something to test with my project was dead in the water with this bug.. I will give this a try. Thanks!

evankyle:
Nice I will give it a test. If you want I can submit another pull request after testing. If so what branch would you like it for or does that matter?

snej:
Yes, please, for the master branch. Thanks!

Issue #727, opened 2015-05-21
ForestDBViewStorage doesn't filter _design docs out when indexing


pasin:
I noticed this while I was working on refactoring the JSViewCompiler. 

When running Router_Tests.test_JSViews. The `emit(doc.message.length, doc.message)` failed as `doc.message.length` is undefined, and the doc id was `_design/design`. The error happened only with ForestDB storage engine.

PR #731, opened 2015-05-21
ForestDBViewStorage, filter out _design docs when indexing

pasin:
- In CocoaIndexer, check to filter out a design doc.
- Update ViewInternal_Tests's -createViewNames to assert if there is a design doc get indexed.

#727

### DIFF
--- a/Source/CBL_ForestDBViewStorage.mm
+++ b/Source/CBL_ForestDBViewStorage.mm
@@ -78,6 +78,8 @@ public:
             indexIt = false;
         } else if (flags & VersionedDocument::kDeleted) {
             indexIt = false;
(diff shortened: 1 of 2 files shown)


pasin:
@snej could you also review this as well so the JSView unit test for ForestDB get passed again?

snej:
Just noticed a nit: The conversion to NSString* is happening outside of the autorelease pool, so those instances are going to pile up during the indexing. Probably not a big problem. It would be better to do that test without having to go to NSString, but there's no existing method on slice to do a hasPrefix type of test.

pasin:
OK, I created a ticket to add a prefix check method to the slice class.

https://github.com/couchbase/couchbase-lite-ios/issues/735

Issue #701, opened 2015-05-14
Couchbase lite iOS memory consumption indexing views


opicciotto:
We have  reproducible case for crashing CBL iOS with a 13K document database.
This app is a standard PhoneGap app using the CBL plugin 1.04; We made it the simplest possible to be sure no other component was interfering with memory consumption.
You can download it from here :

http://d2lm6861ur7e1f.cloudfront.net/webrepository/DebugMemory.zip

This a a cordova /phone app you can use to build with Phonegap Build or locally. The app will connect to a cloud based CouchDB server .

You can use the "Sync" button to trigger the sync downloading all the database to the mobile. Then you can use the "View" button to launch a query on the database using the javascript view. The view will start indexing and the app crashes about 10 to 15 secs later.

This has been tested on an iPad mini iOS8.1 and an iPhone 4 iOS 8.1

pasin:
Thanks for reporting this. I will look at the issue shortly.

pasin:
Reference: https://forums.couchbase.com/t/couchbase-lite-ios-memory-consumption/3578

Some info taken from the forum.

#Document: 13K

Map function:
```Javascript
function (doc) {
// Get Products AND Categories by levelIdFather
if (doc.shopCode && doc.levelIdFather) {
	if (doc.levelId) { // Category
		emit([doc.shopCode, doc.levelIdFather], {name : doc.name, levelId : doc.levelId});
	} else if (doc.sku) { // Product
		if (typeof doc.levelIdFather == "object") {
			for (var key in doc.levelIdFather) {
		        emit([doc.shopCode, key], {sku : doc.sku, name : doc.name, packaging : doc.packaging, priceOfUnit : doc.priceOfUnit, unitPrice : doc.unitPrice, base_url : doc.base_url, imageNumber : doc.imageNumber});
			}
		} else {
			emit([doc.shopCode, doc.levelIdFather], {sku : doc.sku, name : doc.name, packaging : doc.packaging, priceOfUnit : doc.priceOfUnit, unitPrice : doc.unitPrice, base_url : doc.base_url, imageNumber : doc.imageNumber});
		}
	}
}
```

pasin:
I tested the app on an iPhone 6 with iOS 8.3 and the app crashed with memory warning while waiting for the query (or indexing) to return the result. Look like the lot of memory is hold by JS Garbage Collector.


[...]

PR #734, opened 2015-05-21
Refactor CBLJSViewCompiler and CBLJSFunction

pasin:
- Refactor CBLJSViewCompiler and CBLJSFunction to use the new Objective-C JavaScriptCore API.
- Fix out of memory issue created by JSValueMakeFromJSONString(ctx, jsStr) method for some JSON documents (#701)
- Please merge https://github.com/couchbase/couchbase-lite-ios/pull/731 first.


pasin:
I don't see performance gains after changing to use the new API, but the user memory is completely gone.

Indexing the beers data:
Native: ~0.4 secs
Old JSC: ~1.0 secs
New JSC: ~1.0 secs

Source/CBLJSFunction.m
```diff
@@ -48,121 +43,82 @@ - (instancetype) init {
...
-        JSStringRelease(jsBody);
-        for (NSUInteger i = 0; i < _nParams; ++i)
-            JSStringRelease(jsParamNames[i]);
-        
-        if (!_fn) {
-            WarnJSException(_compiler.context, @"JS function compile failed", exception);
+
+        // Evaluate source code:
+        _fnName = [self functionNameFromSourceCode: source];
+        NSString* script = [NSString stringWithFormat: @"var %@ = %@", _fnName, source];
```
snej: I don't think you need to assign the function to a named variable in JS — if you evaluate the source, the result will be the JS function object. Then you can call that directly in the -call method instead of looking up a variable from the context.

pasin:
Not sure what went wrong with the previous tests, but I cleaned up everything and retested. 

Here are the new result in seconds (3 times):
Native: 0.40. 0.39, 0.42
Old JSC: 1.09, 1.08, 1.14
New JSC: 0.75, 0.76, 0.74

Source/CBLJSFunction.m
pasin: I did tried that and I got an exception saying that, `SyntaxError: Function statements must have a name.`.

pasin:
My bad, the map functions are different so stay the same, no performance gain.

Source/CBLJSFunction.m
pasin: One alternative would be letting CBLJSFilterCompiler assign the function name (e.g. function1, 2, 3), and inside CBLJSFilterCompiler has a counter which will be increased every time that the compileXXXFunction is called.

snej: IIRC, you can just wrap parens around the function to make it compilable.

pasin: That works. Great tip!

pasin:
@snej I wrapped the function with a parenthesis so no needs to generate function name anymore. Thanks for the tip.

PR #715, opened 2015-05-19
Added test for deep relationships (NSPredicate)

cflorion:
This new test should succeed but it doesn't.

PR #740, opened 2015-05-27
Fix #715 CBLIS fetch result via nested relationships

pasin:
- Getting the relationship document as the document id instead of the NSManagedObject object.

- Support multiple level relationship search. The default max releationship search depth of 3. Users can customize the max depth by specifiying the custom properties called `kCBLISCustomPropertyMaxRelationshipSearchDepth `.


pasin:
Did have an API review with @zgramana.

PR #744, opened 2015-05-29
Added tests for nil relationships (NSPredicate)

cflorion:
Predicate with nil relationships filter crash.


pasin:
Thanks for reporting this. I will look into this today.

PR #748, opened 2015-06-01
add failing test for sort descriptiors with keyPaths

mastohhh:
Sort descriptors with keyPath as key returns empty result.


pasin:
@mastohhh , thanks a lot for the PR. 

Have you signed the contributor license agreement? If not, could you please sign and let us know? Currently signing with google account doesn't work. You will need to use Yahoo account. Sorry for the inconvenience.

mastohhh:
ok done !

Thanks

PR #751, opened 2015-06-02
add failing test for predicate with bool property

mastohhh:
CBLIncrementalStore fails with requests that have predicates with boolean.

This pull request just adds a test failure to show that bug.

Thanks


pasin:
Thanks for reporting this. I will look into this shortly.

pasin:
CBLIS stores boolean value as 0 or 1 (CoreData boolean value) instead of true or false. The fix would be converting the value from 0 or 1 to NSNumber with the correct boolean value. Users will then need to update their documents to have correct boolean values otherwise they would need to query by using the predicate like 'checked == YES or checked == 1'.

pasin:
I manually picked your commit and merged along with the fix to the master branch. Thanks for the unit test.

PR #753, opened 2015-06-03
cannot nullify NSData property with CBLIncrementalStore

mastohhh:
I just added a simple test that shows when we set to nil a NSData property, the old value is kept.


pasin:
Thanks again for the PR. I will look into the issue shortly.

pasin:
@mastohhh I manually cherry-picked your commit and merged it together with the fix to the master branch. Thanks a lot for the unit test.

PR #754, opened 2015-06-03
Added some way to add information to a document on the fly

cflorion:
In my use case I want users to only see THEIR data, not other's.
So I need to add a "owner" property on each document. But I don't want to add property to all tables in my Core Data model, since I think it is more a metadata than a data, since on the device I don't need that information since, as a user, all I have on my device is my data anyway. But Couchbase needs this information to do a proper rooting.
So I want to add a "owner" property on every documents. created from NSManagedObjects.
But to make it usable by any developer I added a delegate methods that is called when the Document is created to give the opportunity for developer to add any properties they want.

I therefore don't have to touch my datamodel. What do you think ?

### DIFF
--- a/Source/API/Extras/CBLIncrementalStore.m
+++ b/Source/API/Extras/CBLIncrementalStore.m
@@ -74,6 +74,15 @@ static NSError* CBLISError(NSInteger code, NSString* desc, NSError *parent);
     CBLLiveQuery* _conflictsQuery;
     NSString * _documentTypeKey;
     NSUInteger _relationshipSearchDepth;
+    
+    // this struct stores the result of [self.delegate respondsToSelector:]
+    // once and for good instead of calling it each time before calling
+    // the actual methods. This is a good optimization (used by Apple, ie UITableView)
+    //when a method is called a lot.
+    // http://stackoverflow.com/questions/16434395/storing-ios-ui-component-internal-usage-flags
+    struct {
+        unsigned int respondsToKeyValuesToAddToManagedObject:1;
+    } _flags;
 }
 
 @synthesize database = _database;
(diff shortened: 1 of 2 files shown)


pasin:
@cflorion Thanks a lot for the PR. In general, I think this feature is useful. As this PR contains updates to the CBLIS APIs, the review may talk a bit more time. I will keep you posted.

cflorion:
A call to the delegate was missing in method propertiesForDeletingDocument:
Usecase:
I added a "owner" property to all my documents. In my Sync Function on the Gateway  I have : 
```
if (!doc.owner) throw({forbidden : "Documents must have a owner"});
```

So when I was deleting a document on my device, the "owner" property was missing in the dictionary created by propertiesForDeletingDocument:

Source/API/Extras/CBLIncrementalStore.m
```diff
@@ -1708,6 +1724,13 @@ - (id) couchbaseLiteRepresentationOfManagedObject: (NSManagedObject*)object
         [proxy setObject: [object.objectID couchbaseLiteIDRepresentation] forKey: @"_id"];
     }
     
+    if (_flags.respondsToKeyValuesToAddToManagedObject) {
+        NSDictionary *keysAndValuesToAdd = [_delegate keyValuesToAddToManagedObject:object];
+        if (keysAndValuesToAdd.count) {
+            [proxy setValuesForKeysWithDictionary:keysAndValuesToAdd];
```
snej: Shouldn't this be -addEntriesFromDictionary (like the call on line 2203)?

pasin:
@cflorion I manually merged your PR and refactor the delegate according to code review with @zgramana. Thanks a lot for the PR and idea.

PR #773, opened 2015-06-17
CBLIS can't nullify attributes and relationships

mastohhh:
I add two tests to show the problem.

This problem is for every kind of properties (NSString, NSNumber, ...). You recently fix that for NSData properties.

For relationships, I only test to nullify a to-one relationship, but the problem should happen for a to-many ...

### DIFF
--- a/Unit-Tests/IncrementalStore_Tests.m
+++ b/Unit-Tests/IncrementalStore_Tests.m
@@ -827,6 +827,77 @@ static NSArray *CBLISTestInsertEntriesWithProperties(NSManagedObjectContext *con
     AssertNil(att);
 }
 
+- (void) test_NullifyProperty {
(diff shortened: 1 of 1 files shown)


pasin:
The issue nullifying the basic type properties is at the data conversion. We will need to review that.

https://github.com/couchbase/couchbase-lite-ios/blob/master/Source/API/Extras/CBLIncrementalStore.m#L1588-L1621

For releationship properties, I will have to debug the issue. Again, thanks for reporting this.

pasin:
@mastohhh I have fixed the issue and the fixed was merged into the master branch. I manually merged your PR along with the fix. Again, thanks for reporting the issue and the PR.

mastohhh:
Thank you !

Issue #671, opened 2015-04-21
"_removed" properties received from SG cause errors pushing to another server


snej:
If CBL used to have access to a document but loses access, SG will send it one more revision whose only property is `"_removed":true`, to let it know. The problem is that, if CBL pushes the same database to a _different_ server (SG or CouchDB or whatever) it will try to push that document, and the server will (correctly) reject the `_removed` property as illegal because it has a leading underscore.

Since this isn't a "real" revision at all, CBL shouldn't be pushing it. We need to filter it out.

snej:
Marked the severity as "Major" because, if this happens, it stops the entire replication.

pasin:
Currently _remove property is stored in the JSON doc. It's not efficient to do the filtering on the _removed property as we will need to always include the JSON body when querying for the rev changes. Should we introduce removed field in the revs table?

snej:
Probably best to check for `_removed` after the doc's body has been loaded and it's about to be pushed; then just skip it.

zgramana:
@pasin will schedule a meeting to discuss this with the other teams.

pasin:
For some reasons, I couldn't make Sync Gateway to send _removed revisions. I think I don't understand how this actually works. Here is my test.

1. I prepared sync gateway config file with one user names user1 (GUEST is also disabled). I setup sync function to grant access or remove access based on a document property named 'grant'.  The config file is here: https://gist.github.com/pasin/ddc0cad171630874126d#file-config-json.

2. On client side, I have setup continuous push and pull replication. I setup the basic auth to both replicators. Then I created a document with grant = true. Waited for 5 seconds, I updated the document with grant = false. My test code is here:  https://gist.github.com/pasin/ddc0cad171630874126d#file-viewcontroller-m

The test result was no _removed property sent to client. 
- I checked isGone property of the document, and it returned false.
- I checked the actual json in the sqlite file, and there is no _removed revision. 
- I checked the client log and sync gateway log, and there is no indiction of sending or receiving _removed revision after removing the channel access.

Client log :
https://gist.github.com/pasin/ddc0cad171630874126d#file-client-log

Sync Gateway log:
https://gist.github.com/pasin/ddc0cad171630874126d#file-sync_gateway-log

@snej or @adamcfraser could you review my test and let me know what is wrong in my test?

snej:
You're removing the user from the channel. There's a bug/limitation in SG that this doesn't trigger removal of documents (I don't remember the bug #.)

[...]

PR #776, opened 2015-06-18
Filter out _removed revision from push replication

pasin:
- Filter out _removed revision in CBL_Pusher's processInbox when preparing revisions to send to the remote server.

- No changes in CBL_Pusher.unpushedRevisions properties. Therefore unpushedRevisions could contained _removed revisions. The same for getting _changes feed from the CBLRouter.

- Added a unit test

- Added cbl_replicator_removed_rev remote database config to Unit-Tests/Server/cbl_unit_tests.json

#671

### DIFF
--- a/Source/CBL_Pusher.m
+++ b/Source/CBL_Pusher.m
@@ -279,8 +279,8 @@
                             [self revisionFailed];
                             continue;
                         }
-                        
-                        if ([loadedRev[@"_removed"] boolValue]) {
+
+                        if ($castIf(NSNumber, loadedRev[@"_removed"]).boolValue) {
                             // Filter out _removed revision:
                             [self removePending: rev];
                             continue;
--- a/Source/CBL_Pusher.m
+++ b/Source/CBL_Pusher.m
@@ -279,6 +279,12 @@
                             [self revisionFailed];
                             continue;
                         }
+                        
+                        if ([loadedRev[@"_removed"] boolValue]) {
+                            // Filter out _removed revision:
+                            [self removePending: rev];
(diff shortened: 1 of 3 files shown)


Source/CBL_Pusher.m
```diff
@@ -279,6 +279,12 @@ - (void) processInbox: (CBL_RevisionList*)changes {
                             [self revisionFailed];
                             continue;
                         }
+                        
+                        if ([loadedRev[@"_removed"] boolValue]) {
```
snej: This will throw an exception if the `_removed` property contains anything but a NSNumber. Probably best to compare it to `@YES`.

Unit-Tests/Replication_Tests.m
```diff
@@ -969,4 +970,96 @@ - (void) test16_Restart {
...
 
+- (void)test17_RemovedRevision {
+    NSURL* remoteDbURL = [self remoteTestDBURL: kRemovedRevDBName];
+    if (!remoteDbURL)
+        return;
+    [self eraseRemoteDB: remoteDbURL];
+
+    NSError* error;
+
+    // Create a new document with grant = true:
```
snej: This seems too complex for a unit test. It would be fine to just manually create a revision with a `"_removed":true` property and then verify that it doesn't get pushed.

pasin:
@snej I have updated the code according to your review comment. I also rebased with the current master. 

To check _removed value, I use `$castIf(NSNumber, loadedRev[@"_removed"]).boolValue` the same way as the other places.

PR #783, opened 2015-06-23
New replicator

### LINKED ISSUE #702: Integrate new replicator
I've implemented a new replicator that uses a new protocol (BLIP-over-WebSocket). 

* Spec is [here](https://github.com/couchbaselabs/couchbase-lite-api/wiki/New-Replication-Protocol).
* Cocoa implementation is in the `feature/new_replicator` branch. It's up to date with the master branch (as of last night.)
* Sync Gateway implementation is in the `feature/blipsync` branch.

Architecturally what I've done is create a protocol, `CBL_Replicator`, that defines the replicator implementation's API. The CBLReplicator class is renamed CBLRestReplicator and implements CBL_Replicator. The new replicator is CBLBlipReplicator. Everything else only uses the protocol.

CBLManager has a property `replicatorClassName` that controls what replicator class to use. It defaults to `@"CBLRestReplicator` but can be overridden programmatically or by setting the `"CBLReplicatorClass"` user default. `-[CBLManager replicatorWithProperties:status:` uses that property to instantiate a replicator.

To simplify the replicator implementation and allow more code sharing, all of the configuration state is extracted into a class `CBL_ReplicatorSettings`. The replicator itself is given an instance of this when it's created.

snej:
Fixes #702


snej:
Some implementation notes:

* I refactored CBL_Replicator into a protocol, to act as the common base of the two replicators.
* A new class CBL_ReplicatorSettings provides a place to share common code between the replicator implementations.
* The existing implementation of CBL_Replicator was renamed CBLRestReplicator; Pusher and Puller were renamed the same way.
* The new replicator is CBLBlipReplicator. It's mostly an adapter between the CBL_Replicator API and a lower-level class CBLSyncConnection that does the actual work.
* CBLManager has a new property `replicatorClassName` which defaults to `"CBLRestReplicator`. Set it to `"CBLBlipReplicator"` to enable the new replicator.

Issue #777, opened 2015-06-18
P2P replication error (lack of Digest auth support in CBLChangeTracker)


joshblour:
I'm not able to replicate from a CBLListener with authentication turned on. Works great with authentication turned off.

CBLListener code:
```` objc
self.listener = [[CBLListener alloc] initWithManager:[CBLManager sharedInstance] port:52021];
self.listener.requiresAuth = YES;
[self.listener setPasswords:@{[self authenticatorUsername] : [self authenticatorPassword]}];
[self.listener setBonjourName:[self myUUID] type: [self bonjourType]];
[self.listener start:&error]
````
Client code:
```objc
NSURLComponents *components = [[NSURLComponents alloc] init];
components.scheme = @"http";
components.host = service.hostName;
components.port = @(service.port);
components.path = [NSString stringWithFormat:@"/%@", [[self database] name]];
components.user = [self authenticatorUsername];
components.password = [self authenticatorPassword];
self.pullReplication = [[self database] createPullReplication:components.URL];
self.pullReplication.continuous = self.isContinuous;
self.pullReplication.customProperties = @{@"websocket": @NO};
self.pullReplication.filter = @"exceptUUIDAndMinUpdatedAt";
self.pullReplication.filterParams = @{@"exceptUUID": uuid, @"minUpdatedAt": [CBLJSON JSONObjectWithDate:minUpdatedAt]};
[self.pullReplication start];
```

Logs from the client here: https://gist.github.com/joshblour/59b76b2531dd0f2eb3b4
Logs from the CBLListener here: https://gist.github.com/joshblour/163a9d101f1f4b10dc56

The error seems to happen on the CBLListener side: 
````
CBLListener: Login attempted for user 'p2p_v1'
CBLListener: GET /proximator/_local/e95ce455ab12e1261e4f6b592ee4cc8f478c8eaa
CBLListener:     Response[GET /proximator/_local/e95ce455ab12e1261e4f6b592ee4cc8f478c8eaa] --> 404
````

**Comment by joshblour** [comment]
I'm able to replicate this in my fork of the Grocery Sync app: 
https://github.com/joshblour/Grocery-Sync-iOS/tree/password

**Comment by snej** [comment]
FYI, the 404 status when GETting a `_local/` document is not an error. Those documents are replication checkpoints, and the 404 just means the server/peer doesn't yet have a saved checkpoint for this replication.

**Comment by snej** [comment]
The error on the client side:
```
[...]

PR #792, opened 2015-06-26
Fix change tracker digest authentication

pasin:
- In CBL_Replication.start method, not creating the basic authorizer from the credentials specified in the remote url. The credentails in the URL will be automatically picked up by the NSURLConnection in CBLRemoteRequest to do (basic or digest) authentication. This also allows CBLSocketChangeTracker to get the challenge response and apply credentials to the challenge approriately based on the authentication scheme.

- In CBLRemoteRequest's connection:willSendRequestForAuthenticationChallenge: delegate method, check if the auth method is basic before creating and saving a basic authorizer.

- In CBLScoketChangeTracker, handle the case of getting authentication challenge responses (401) correctly. When getting the challenge response, we will get a NSStreamEventEndEncountered event without getting any NSStreamEventHasBytesAvailable events. Therefore we also need to check the header when getting the NSStreamEventEndEncountered event.

- Fixed WebSocketHTTPLogic's CFHTTPMessageAddAuthentication  () call by passing NULL to the auth method parameter. This allows the auth method to be selected based on the auth failure response. (PR: https://github.com/couchbaselabs/WebSockets-Cocoa/pull/3)

#777

### DIFF
--- a/Source/CBLRemoteRequest.m
+++ b/Source/CBLRemoteRequest.m
@@ -294,8 +294,11 @@ void CBLWarnUntrustedCert(NSString* host, SecTrustRef trust) {
             if (cred) {
                 LogTo(RemoteRequest, @"    challenge: useCredential: %@", cred);
                 [sender useCredential: cred forAuthenticationChallenge:challenge];
-                // Update my authorizer so my owner (the replicator) can pick it up when I'm done
-                _authorizer = [[CBLBasicAuthorizer alloc] initWithCredential: cred];
+
+                if ($equal(authMethod, NSURLAuthenticationMethodHTTPBasic)) {
+                    // Update my authorizer so my owner (the replicator) can pick it up when I'm done
+                    _authorizer = [[CBLBasicAuthorizer alloc] initWithCredential: cred];
+                }
                 return;
             }
         }
--- a/Source/CBL_Replicator.m
+++ b/Source/CBL_Replicator.m
@@ -341,13 +341,6 @@ NSString* CBL_ReplicatorStoppedNotification = @"CBL_ReplicatorStopped";
                  }
                 ];
 
-    // If client didn't set an authorizer, use basic auth if credential is available:
-    if (!_authorizer) {
-        _authorizer = [[CBLBasicAuthorizer alloc] initWithURL: _remote];
-        if (_authorizer)
-            LogTo(SyncVerbose, @"%@: Found credential, using %@", self, _authorizer);
-    }
-
     self.running = YES;
     _startTime = CFAbsoluteTimeGetCurrent();
     
--- a/Source/ChangeTracker/CBLSocketChangeTracker.m
+++ b/Source/ChangeTracker/CBLSocketChangeTracker.m
@@ -161,7 +161,13 @@
     CFHTTPMessageRef response;
     response = (CFHTTPMessageRef) CFReadStreamCopyProperty((CFReadStreamRef)_trackingInput,
                                                            kCFStreamPropertyHTTPResponseHeader);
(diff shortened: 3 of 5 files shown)


Source/CBLRemoteRequest.m
```diff
@@ -294,8 +294,11 @@ - (void)connection:(NSURLConnection *)connection
             if (cred) {
                 LogTo(RemoteRequest, @"    challenge: useCredential: %@", cred);
                 [sender useCredential: cred forAuthenticationChallenge:challenge];
-                // Update my authorizer so my owner (the replicator) can pick it up when I'm done
-                _authorizer = [[CBLBasicAuthorizer alloc] initWithCredential: cred];
+
+                if ($equal(authMethod, NSURLAuthenticationMethodHTTPBasic)) {
```
snej: Why this test? It shouldn't matter here whether basic or digest auth is used; this code is just handling looking up a credential from the keychain.

Source/CBL_Replicator.m
```diff
@@ -341,13 +341,6 @@ - (void) start {
                  }
                 ];
 
-    // If client didn't set an authorizer, use basic auth if credential is available:
-    if (!_authorizer) {
-        _authorizer = [[CBLBasicAuthorizer alloc] initWithURL: _remote];
```
snej: Same here — pulling credential out of the URL string should work with both basic and digest auth.

pasin: We don't know if the authentication will be basic or digest from the url. Setting the basic authorizer from the URL here will cause CBLRemoteRequest and CBLSocketChangeTracker to always send the 1st request with Basic Authorization header. If the authentication happens to be digest, the password will be leaked in the header as the Basic authentication.

1.https://github.com/couchbase/couchbase-lite-ios/blob/feature/issue_777_digest_auth/Source/CBLRemoteRequest.m#L102-L108

2. https://github.com/couchbase/couchbase-lite-ios/blob/feature/issue_777_digest_auth/Source/ChangeTracker/CBLSocketChangeTracker.m#L80-L86

Source/CBLRemoteRequest.m
pasin: Just saw this. Same reason as below.

Source/CBL_Replicator.m
pasin: One thing that I should ask. Do we intend to use CBLBasicAuthorizer for digest as well? If the answer is Yes, then my comment above doesn't matter because we can't know in advance if it's basic or digest from CBLBasicAuthorizer as of now.

snej: I think we need to, because we need an authorizer to handle digest auth, and because the two are pretty interchangeable except for the details of what goes in the Authorization header. Probably should be renamed CBLPasswordAuthorizer.

You're right that we shouldn't send the username/password before receiving the 401 authentication challenge from the server. We'll also need to improve the CBLAuthorizer API so the implementation can see the challenge info when responding — at a minimum it should be given the value of the response's WWW-Authenticate header.

snej:
It should be possible to have the PasswordAuthorizer call CFHTTPMessageAddAuthentication so that it doesn't have to re-implement digest auth on its own. The only problem is the mismatch between CFHTTPMessage and NSURLRequest/Response. I suppose it could translate between the two (I think we've already got some code somewhere that does that, at least in one direction.)

pasin:
The scope of this ticket gets larger. Let me take the PR back and work on this.

pasin:
I plan to revert those 2 changes back first, and we could have another ticket to take care CBLPasswordAuthorizer and so on.

PR #795, opened 2015-06-30
CBLIS evaluate relationship with nsmanagedobjectid

mastohhh:
Sometimes, when managed objects are not fault, predicates are evaluated with ManagedObjectIDs, and CBLIS didn't handle that.

### DIFF
--- a/Source/API/Extras/CBLIncrementalStore.m
+++ b/Source/API/Extras/CBLIncrementalStore.m
@@ -1448,6 +1448,8 @@ static CBLManager* sCBLManager;
             value = [expression constantValue];
             if ([value isKindOfClass: [NSManagedObject class]])
                 value = [[value objectID] couchbaseLiteIDRepresentation];
+            if ([value isKindOfClass: [NSManagedObjectID class]])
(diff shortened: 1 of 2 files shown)


snej:
Thanks for the PR! Pasin's on vacation for a few weeks, but he'll review it when he gets back.

Source/API/Extras/CBLIncrementalStore.m
```diff
@@ -1448,6 +1448,8 @@ - (id) evaluateExpression: (NSExpression*)expression
             value = [expression constantValue];
             if ([value isKindOfClass: [NSManagedObject class]])
                 value = [[value objectID] couchbaseLiteIDRepresentation];
+            if ([value isKindOfClass: [NSManagedObjectID class]])
```
pasin: Can you change from if to else if?

pasin:
Thanks for the PR. I have one minor comment as above.

pasin:
I manually merged and added a minor fix to use scanContantValue: method. Thanks again for the fix and the unit test.

PR #796, opened 2015-06-30
CBLIS use a ios7 compatible method

mastohhh:
use rangeOfString instead of containsString ios8 only method


snej:
Looks good, thanks!

Issue #793, opened 2015-06-26
CBLIS - Slow down UI thread when saving CoreData context


pasin:
@mastohhh reported this to me via email:

I’m @mastohhh, I'm writing directly to you because I don’t find answers on Github or Google Groups.
(I created a topic but no answer : https://groups.google.com/forum/#!topic/mobile-couchbase/SSUuoF_DJ_Q)

My question is : I have performance problems with couchbase-lite and CBLIncrementalStore, do you ? What’s the best way to create core data stack with CBLIS that doesn’t slows the UI (main thread) ?

I did a little hack to get the private context of the NSManagedObjectContext, here’s my code.

    CBLManager *manager = [[CBLManager sharedInstance] copy];
    [CBLIncrementalStore setCBLManager:manager];

    self.parentContext = [[NSManagedObjectContext alloc] initWithConcurrencyType:NSPrivateQueueConcurrencyType];
    [self.parentContext setPersistentStoreCoordinator: coordinator];

    [self.parentContext performBlock:^{
        manager.dispatchQueue = dispatch_get_current_queue();

        NSError *error;

        self.incrementalStore = (CBLIncrementalStore*)[coordinator addPersistentStoreWithType:[CBLIncrementalStore type]
                                                                               configuration:nil
                                                                                         URL:incrementalStoreURL
                                                                                     options:nil
                                                                                       error:&error];

        self.context = [[NSManagedObjectContext alloc] initWithConcurrencyType:NSMainQueueConcurrencyType];
        self.context.parentContext = self.parentContext;

        [self.incrementalStore addObservingManagedObjectContext:self.context];
        [self.incrementalStore addObservingManagedObjectContext:self.parentContext];

        NSURL *remoteDbURL = [NSURL URLWithString:COUCHBASE_SYNC_URL];
        [self startReplication:[self.incrementalStore.database createPullReplication:remoteDbURL]];
        [self startReplication:[self.incrementalStore.database createPushReplication:remoteDbURL]];
    }]; 

Before Couchbase, this configuration was quite speed, because the context with NSMainQueueConcurrencyType didn’t write directly on the SQL file, it was his parent’s context with NSPrivateQueueConcurrencyType that did it.

Now I think there is work on the main thread with Couchbase that slows the UI after every context’s save.

I call save on every user’s actions, because I wan’t a fast sync.

What’s your strategy ? How is your core data stack ? Main queue ? Private queue ? Custom Queue ? CBLManager’s queue ?

[...]

PR #797, opened 2015-06-30
Add few optimizations

mastohhh:
try to send more work on background queues instead of main thread

### DIFF
--- a/Source/API/Extras/CBLIncrementalStore.m
+++ b/Source/API/Extras/CBLIncrementalStore.m
@@ -2039,64 +2039,71 @@ static CBLManager* sCBLManager;
     [_coalescedChanges addObjectsFromArray: changes];
     [self processCouchbaseLiteChanges];
 #else
-    [NSThread cancelPreviousPerformRequestsWithTarget: self selector: @selector(processCouchbaseLiteChanges) object: nil];
-    
-    @synchronized(self) {
-        [_coalescedChanges addObjectsFromArray: changes];
-    }
-    
-    [self performSelector: @selector(processCouchbaseLiteChanges) withObject: nil afterDelay: 0.1];
+    dispatch_async(dispatch_get_main_queue(), ^{
+        [NSThread cancelPreviousPerformRequestsWithTarget: self selector: @selector(processCouchbaseLiteChanges) object: nil];
+
+        @synchronized(self) {
+            [_coalescedChanges addObjectsFromArray: changes];
+        }
+
+        [self performSelector: @selector(processCouchbaseLiteChanges) withObject: nil afterDelay: 0.1];
+    });
 #endif
 }
 
 - (void) processCouchbaseLiteChanges {
-    NSArray* changes = nil;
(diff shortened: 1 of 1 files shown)


snej:
I'll let Pasin review this when he gets back. As far as thread-safety, if you're doing all the background work by calling -backgroundTellDatabase:, then you should be fine.

mastohhh:
Ok thanks. I know Pasin manages the `CBLIncrementalStore` part of the framework, so maybe you can give me an idea on that : https://github.com/couchbase/couchbase-lite-ios/issues/793 ? Slow `CBLQueryEnumerator`, or if not slow, maybe it's called too often, so needs cache ?

snej:
The CPU profiler in the Instruments app can help you track down why the query enumeration is slow. Usually the slow part is updating the index, which happens if the database has been changed since the view was last queried; every changed document gets run through the map function again and the old emitted rows get replaced by the new ones.

mastohhh:
I tried to investigate and I don't think I can optimize this method.

My idea is : I saw that CoreData calls a lot that function, and maybe multiple times for the same query. In this case, I could reuse results instead of querying again.

I'll pull request you if I found something interesting. 

Thanks

snej:
Caching results might be a good idea. You can check whether a query is still valid by saving the `CBLQueryEnumerator.sequenceNumber` after you run the query, and then later comparing that value with the `CBLDatabase.lastSequenceNumber`. If it's lower, you need to re-run the query.

mastohhh:
I tried it ! But 2 tests are failing, so no need to pull request what I did ...

Or maybe if you want to know what not to do.

The good thing is : between half and quarter of queries can be reused.

pasin:
@mastohhh Thanks for the PR. Is this PR fixing the performance issue when saving core data context?
Do you have performance result before and after applying the PR?

pasin:
@mastohhh I have fixed the issue in https://github.com/couchbase/couchbase-lite-ios/commit/653ebafadc11bd3a8a2a71a4654098c692a14bf7. I'm closing the PR now. Please let me know if you have any issues.

PS: I also cache the relationship query result by checking CBLQueryEnumerator.sequenceNumber. I didn't see any unit tests failed.

mastohhh:
Hi @snej @pasin, I'm back on CouchbaseLite project ! Really thank you for what you did.

I still have performance problems, and I'm working on it, I'll come back to you if I found something.

mastohhh:
@pasin I think you forgot to initialize `_relationshipCache` ivar in `CBLIncrementalStore`.

Tell me if you want a PR, but I think it's not necessary just for one line.

pasin:
That's bad. Let me do it. I would like to re-test everything.

pasin:
@mastohhh I'm sorry for the bug. I have just fixed it. Recently we deprecated the enumerator's reset method in the master branch, and the enumerator should be reset before returning it from the cache. I have just fixed that issue as well. Please let me know if it solves your issue.

PR #819, opened 2015-07-22
Fix CBLReplication.pendingDocumentIds when replication is not running

pasin:
1. Query unpushed revisions from the last sequence. When the bg replicator and the bg replication's last sequence is not null, use the last sequence from the bg replicator. Otherwise use the last sequence from the database.

2. In CBLBlipReplicator, for non-continuous replication, update the status to kCBLReplicatorStopped after connection is closed. This will allow the current last sequence to be synced with the remote and saved into the database before kCBLReplicatorStopped status is notified. This means that for new replicator, we will see the (one-time) replication status like below when it replicates 10 docs:

 ```
Replication status=3; completedChangesCount=0; changesCount=0
Replication status=3; completedChangesCount=10; changesCount=10
Replication status=0; completedChangesCount=10;
```

3. I didn't refactor CBLRestReplicator to wait until the last sequence is saved before notifying kCBLReplicatorStopped status as the current code around that area is quite complex and we already did optimistic saving for the last sequence when the replicator is stopped.

4. I didn't refactored CBLSyncConnection+Push when it queries pending documents as I feel it looks good the way it is now. 

#807

### DIFF
--- a/Source/API/CBLReplication.m
+++ b/Source/API/CBLReplication.m
@@ -403,7 +403,7 @@ NSString* CBL_ReplicatorStoppedNotification = @"CBL_ReplicatorStopped";
                                                 toDatabase: nil
                                                     status: &status];
         if (!settings) {
-            Warn(@"Error parsing replicatior settings : %@", CBLStatusToNSError(status));
+            Warn(@"Error parsing replicator settings : %@", CBLStatusToNSError(status));
             return nil;
         }
 
--- a/Source/API/CBLReplication.m
+++ b/Source/API/CBLReplication.m
@@ -390,18 +390,41 @@ NSString* CBL_ReplicatorStoppedNotification = @"CBL_ReplicatorStopped";
 
 #pragma mark - PUSH REPLICATION ONLY:
 
-
 - (NSSet*) pendingDocumentIDs {
-    if (!_pendingDocIDs && _started && !_pull) {
-        _pendingDocIDs = [self tellReplicatorAndWait: ^NSSet*(id<CBL_Replicator> bgReplicator) {
-            if ([_bg_replicator respondsToSelector: @selector(pendingDocIDs)])
-                return _bg_replicator.pendingDocIDs;
-            else
-                return nil;
-        }];
-    }
+    if (_pull || (_started && _pendingDocIDs))
+        return _pendingDocIDs;
+
+    _pendingDocIDs = [_database.manager.backgroundServer
+                      waitForDatabaseNamed: _database.name to: ^id(CBLDatabase* bgDb)
+    {
+        CBLStatus status;
+        CBL_ReplicatorSettings* settings =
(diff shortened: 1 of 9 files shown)

### CODE REVIEW

File: Source/CBL_Replicator.h [hunk 1]
```diff
@@ -77,8 +77,10 @@ extern NSString* CBL_ReplicatorStoppedNotification;
 /** Called by CBLDatabase to notify active replicators that it's about to close. */
 - (void) databaseClosing;
 
-#if DEBUG // for unit tests
+/** Current lastSequence tracked by the replicator. */
 @property (readonly) id lastSequence;
```
snej: In the implementations of CBL_Replicator, `lastSequence` is only synthesized in a debug build, so I think a release build is going to fail with an error about a missing property implementation. You'll need to move the `@synthesize` out of `#if DEBUG`.

Source/API/CBLReplication.m
```diff
@@ -390,19 +390,42 @@ - (void) updateStatus: (CBLReplicationStatus)status
...
+    _pendingDocIDs = [_database.manager.backgroundServer
+                      waitForDatabaseNamed: _database.name to: ^id(CBLDatabase* bgDb)
+    {
+        CBLStatus status;
+        CBL_ReplicatorSettings* settings =
+            [bgDb.manager replicatorSettingsWithProperties: self.properties
+                                                toDatabase: nil
+                                                    status: &status];
+        if (!settings) {
+            Warn(@"Error parsing replicatior settings : %@", CBLStatusToNSError(status));
```
snej: Typo in `replicator settings`.

pasin:
Thanks for the review. I have fixed the last sequence debug synthesizing and the typo per the review. I also rebased with the current master to get all unit-tests passed when testing.

Source/API/CBLReplication.m
```diff
@@ -390,19 +390,42 @@ - (void) updateStatus: (CBLReplicationStatus)status
 
 #pragma mark - PUSH REPLICATION ONLY:
 
-
 - (NSSet*) pendingDocumentIDs {
-    if (!_pendingDocIDs && _started && !_pull) {
-        _pendingDocIDs = [self tellReplicatorAndWait: ^NSSet*(id<CBL_Replicator> bgReplicator) {
-            if ([_bg_replicator respondsToSelector: @selector(pendingDocIDs)])
-                return _bg_replicator.pendingDocIDs;
```
snej: Looks like the property `CBL_Replicator.pendingDocIDs` and its implementations can be deleted, since you've deleted the only call site.

pasin:
Yes, I removed the pendingDocIDs and push. Thanks.

tleyden:
@zgramana / @snej any idea if this made it into the official API docs?

Ref: https://forums.couchbase.com/t/how-to-know-if-a-document-is-synchonized-with-a-remote-server/5774/3

PR #842, opened 2015-08-04
Fix #74 CBLRouter's processRequestRanges to return 416 when the range is invalid

### LINKED ISSUE #74: Listener doesn't return 416 for unsatisfiable Range: header
The 3rd-party CocoaHTTPServer framework doesn't quite handle HTTP range requests properly. If a range isn't satisfiable, it just ignores the Range: header (and sends back the whole body with a 200 status) instead of returning a 416 status.

This causes a failure in a CouchDB unit test -- attachment_ranges.js:146

```
$ ./bin/couchdb-harness -p 59840 attachment_ranges
1..1
not ok 1 attachment_ranges
Reason: expected '416', got '200'
Trace back (most recent call first):
    
  46: /Couchbase/couchdb-harness/util/cli_runner.js
      T(false,"expected '416', got '200'","fetch 5-3")
 321: /Couchbase/couchdb-harness/script/couch_test_runner.js
      TEquals(416,200,"fetch 5-3")
 146: /Couchbase/couchdb-harness/script/test/attachment_ranges.js
      ()
  53: /Couchbase/couchdb-harness/util/cli_runner.js
      runTestConsole(1,"attachment_ranges",(function (debug) {var db = new C
  72: /Couchbase/couchdb-harness/util/cli_runner.js
      runAllTestsConsole()
  85: /Couchbase/couchdb-harness/util/cli_runner.js
      
Test failures: 1
```

pasin:
- Aligned with the fix in CocoaHTTPServer (d44cf160dd72fa5a1df9924e33ad367c3e37b798), return 416 when the range is invalid.

- Added unit test cases

#74

### DIFF
--- a/Source/CBL_Router.m
+++ b/Source/CBL_Router.m
@@ -553,9 +553,11 @@ static NSArray* splitPath( NSURL* url ) {
     NSUInteger from, to;
     if (fromStr.length > 0) {
         from = (NSUInteger)fromStr.integerValue;
-        if (toStr.length > 0)
+        if (toStr.length > 0) {
             to = MIN((NSUInteger)toStr.integerValue, bodyLength - 1);
-        else
+            if (to < from)
+                return;  // invalid range
+        } else
             to = bodyLength - 1;
     } else if (toStr.length > 0) {
         to = bodyLength - 1;
@@ -571,9 +573,6 @@ static NSArray* splitPath( NSURL* url ) {
         _response.body = nil;
         return;
     }
-
-    if (to < from)
-        return;  // invalid range
 
     if (from == 0 && to == bodyLength - 1)
         return; // No-op; entire body still causes a 200 response
--- a/Source/CBL_Router.m
+++ b/Source/CBL_Router.m
@@ -554,9 +554,10 @@ static NSArray* splitPath( NSURL* url ) {
     if (fromStr.length > 0) {
         from = (NSUInteger)fromStr.integerValue;
         if (toStr.length > 0) {
-            to = MIN((NSUInteger)toStr.integerValue, bodyLength - 1);
+            to = (NSUInteger)toStr.integerValue;
             if (to < from)
                 return;  // invalid range
+            to = MIN(to, bodyLength - 1);
         } else
             to = bodyLength - 1;
     } else if (toStr.length > 0) {
--- a/Source/CBL_Router.m
+++ b/Source/CBL_Router.m
@@ -550,7 +550,6 @@ static NSArray* splitPath( NSURL* url ) {
         toStr = [rangeHeader substringWithRange: r];
 
     // Now convert those into the integer offsets (remember that 'to' is inclusive):
-    BOOL missingRange = NO;
     NSUInteger from, to;
     if (fromStr.length > 0) {
         from = (NSUInteger)fromStr.integerValue;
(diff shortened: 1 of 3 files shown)


snej:
It's been a while since I implemented this, but I just looked up the HTTP spec and I think my original code is correct. There is a difference between an _unsatisfiable_ byte range (one that doesn't overlap the content) and an _invalid_ byte range (syntactically incorrect.)

[RFC 2616](http://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.35) says:
>The recipient of a byte-range- set that includes one or more syntactically invalid byte-range-spec values MUST ignore the header field that includes that byte-range- set.

So the places where my original code does something like `return;  // invalid range` are complying with the RFC to ignore the `Range:` header instead of rejecting the request. (I'm not exactly sure why it's supposed to be this way...)

snej:
This probably means some of the changes in your patch to CocoaHTTPServer were incorrect too. Sorry! I didn't get curious enough to look up the spec until I saw that you were changing my code in the router.

pasin:
I see but this is confusing. Does this mean that CouchDB unit test (bytes=5-3 returns 416) is not correct?

pasin:
[RFC 2616](http://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.35) says:
> If the last-byte-pos value is present, it MUST be greater than or equal to the first-byte-pos in that byte-range-spec, or the byte- range-spec is syntactically invalid.

Does that mean something like bytes=5-3 is syntactically invalid so it shouldn't be treated at 416?

snej:
Hm, yes I think so.

pasin:
I have modified and push the changes (also at CocoaHTTPServer https://github.com/couchbaselabs/CocoaHTTPServer/pull/2). Could you please review? Thank you.

Source/CBL_Router.m
```diff
@@ -573,6 +571,9 @@ - (void) processRequestRanges {
         return;
     }
 
+    if (to < from)
```
snej: I think the `to < from` check needs to happen before the `from >= bodyLength` check. For example, if the bodyLength is 100 and the range is `500-400`, the response status should be 200 (since the range is invalid) not 416.

pasin: Agreed. Checking the valid of the range should take precedent. 

What do you think about the case of `500-` (bodyLength=100)? I understand that it is syntactically valid and it would return 416.

pasin:
I have push the fix again. I believe it is inline with the spec now. I handled the case of `500-` above as syntactically valid range request according to what the spec said about the presence of the last-byte-pos value here:

>If the last-byte-pos value is present, it MUST be greater than or equal to the first-byte-pos in that byte-range-spec, or the byte- range-spec is syntactically invalid. The recipient of a byte-range- set that includes one or more syntactically invalid byte-range-spec values MUST ignore the header field that includes that byte-range- set.
>
>If the last-byte-pos value is absent, or if the value is greater than or equal to the current length of the entity-body, last-byte-pos is taken to be equal to one less than the current length of the entity- body in bytes.

Sorry for the bug and Could you please review it one more time?

snej:
This needs a merge or rebase now...

pasin:
I have just rebased and forced push. Could you please re-review?

Source/CBL_Router.m
```diff
@@ -553,12 +553,12 @@ - (void) processRequestRanges {
     NSUInteger from, to;
     if (fromStr.length > 0) {
         from = (NSUInteger)fromStr.integerValue;
-        if (toStr.length > 0)
+        if (toStr.length > 0) {
             to = MIN((NSUInteger)toStr.integerValue, bodyLength - 1);
-        else
+            if (to < from)
+                return;  // invalid range
+        } else
```
snej: This would incorrectly make "200-201" an invalid range if the bodyLength is 100... because `from` would be 200 and `to` would be 99.
I think you need to do the `MIN` operation after the `to < from` comparison.

pasin: Yes, your are right. CocoaHTTPServer is not right for this case too. I will fix CocoaHTTPServer separately.

pasin:
I have fixed the the case "200-201" bug as above comment. More unit tests added. Could you please review? Thanks!

PS: I will submit another PR to CocoaHTTPServer to match the same logic.

Issue #843, opened 2015-08-05
CBLDatabase not getting attachments from ancestry when processing attachments


pasin:
Originally reported by https://github.com/couchbase/couchbase-lite-java-core/issues/677.

PR #844, opened 2015-08-05
Fix process attachments when the parent rev is a _deleted rev

pasin:
When the intermediate parent rev is _deleted rev, the status returned when getting the parentAttachments will be kCBLStatusOK instead of kCBLStatusNotFound. So it should go to the same logic as when the stastus is kCBLStatusNotFound to get the attachment from the attachment revpos instead.

#843


pasin:
I have just refactored my previous fix and re-push to ensure that the output status is returned correctly. Thanks.

PR #848, opened 2015-08-07
Fix #698 replication stopped notification not being sent when closing the database

### LINKED ISSUE #698: CBLReplicator could still be in the running status after closing the database
I have noticed this behavior while looking at #696. The issue was that the running status of the CBLReplicator will be set to 'NO' in the - updateStatus: method, which will be called asynchronously on the database thread within a block given to CBLDatabase's -doAsync: method when the replicator receives the -bg_updateProgress message from the bg_replicator thread.

However by the time that the async block get called, the database has already been closed so actually there is no opt (CBLDatabase's -doAsync: checks first if the database is still opened before excuting the block).

CBLReplicator:
```Objective-C
- (void) bg_updateProgress {
   ...
    __weak CBLReplication *weakSelf = self;
    [_database doAsync: ^{
        CBLReplication *strongSelf = weakSelf;
        [strongSelf updateStatus: status error: error processed: changes ofTotal: total
                          serverCert: serverCert];
        cfrelease(serverCert);
    }];
}
```

CBLDatabase:
```Objective-C
static void catchInBlock(void (^block)()) {
    @try {
        block();
    }catchAndReport(@"-[CBLDatabase doAsync:]");
}

- (void) doAsync: (void (^)())block {
    block = ^{
        if (_isOpen)
            catchInBlock(block);
    };
    if (_dispatchQueue)
        dispatch_async(_dispatchQueue, block);
    else
        MYOnThreadInModes(_thread, CBL_RunloopModes, NO, block);
}
```

pasin:
When updating the replication on the main thread, instead of calling -doAsync: on the database, calling it on the manager object to loose the database open status check.

#698

### DIFF
--- a/Source/API/CBLReplication.m
+++ b/Source/API/CBLReplication.m
@@ -599,11 +599,15 @@ NSString* CBL_ReplicatorStoppedNotification = @"CBL_ReplicatorStopped";
     }
 
     // Communicate its state back to the main thread:
+    // Call -doAsync: on the manager object instead of on the database object
+    // to allow the block to be executed regardless of the database open status.
+    // Without loosing the database open status check, the stopped notification
+    // can't be sent when the replication is stopped from closing the database.
     __weak CBLReplication *weakSelf = self;
-    [_database doAsync: ^{
+    [_database.manager doAsync:^{
         CBLReplication *strongSelf = weakSelf;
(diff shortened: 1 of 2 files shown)


snej:
This needs a merge or rebase now...

pasin:
I have rebased and forced push. Could you please re-review?

PR #855, opened 2015-08-10
Fixed #686 background database not closed when close database

### LINKED ISSUE #686: The bg replication server's database not getting closed

pasin:
Observed the CBL_DatabaseWillCloseNotification notification to close the background database. This is the same machanisim used when deleting the database. Note that the background database will be closed asynchronously.

#686

### DIFF
--- a/Source/CBLDatabase+Internal.m
+++ b/Source/CBLDatabase+Internal.m
@@ -196,7 +196,9 @@ static BOOL sAutoCompact = YES;
         return NO;
     }
 
+    [self willChangeValueForKey: @"isOpen"];
     _isOpen = YES;
+    [self didChangeValueForKey: @"isOpen"];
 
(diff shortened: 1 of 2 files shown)


snej:
This implementation means that closing a database will close _every other_ connection to that database. I don't think that's the right thing to do. If I open two CBLManager instances for separate threads, and open a CBLDatabase on the same db file in each one, then closing one of them shouldn't close the other.

The case to be fixed here is pretty specific: when a database is closed, and its manager has a background manager, then the background manager should close the database with the same name. This seems like it can be handled in -[CBLManager _forgetDatabase:].

pasin:
I overlooked that case. I will take the PR back and re-work on this. Thanks for reviewing.

Issue #726, opened 2015-05-21
CBLListener.readOnly=YES prevents clients from storing checkpoints


snej:
If an app with a CBLListener sets its `readOnly` property to YES (to prevent clients from modifying its databases), it has the unwanted side effect that clients' pull replicators can't store checkpoints, because the request to PUT the `_local` doc gets rejected with a 403. This means that every replication has to start over from the beginning, which hurts performance.

(The same thing happens when replicating with a CouchDB/Cloudant server without write access.)

Options:

* Make the listener allow PUTs to `_local` docs even in readOnly mode
* Make CBLReplicator more graceful about an inability to store a remote checkpoint, i.e. remember the fact and don't require the remote checkpoint to exist the next time.

snej:
Another option:

* Give the app finer grained control over remote access to docs, allowing the peer to update `_local` docs but not any regular docs.

pasin:
@snej does the CBLSyncListener support readOnly mode yet? I tried to follow the code but I couldn't find where that got implemented.

snej:
Oops, it doesn't. Just filed #838 to remind myself.

PR #867, opened 2015-08-14
Fix #838 Implement readOnly to CBLSyncListener

### LINKED ISSUE #838: Implement readOnly mode in CBLSyncListener
CBLSyncListener, the listener for the new replication protocol, doesn't yet implement read-only (i.e. pull-only) access.

pasin:
- Added OnSyncAccessCheckBlock property to CBLSyncConnection similarly to CBL_Router.OnAccessCheckBlock property. OnSyncAccessCheckBlock (if specified) will be called to check the access for each sync request coming to the CBLSyncConnection's handlers. The handlers will then return the corresponding error responses depending on the result of the OnSyncAccessCheckBlock call.

- In read only mode, CBLSyncListener will block access to setCheckpoint, changes, and rev requests.

#838

### DIFF
--- a/Source/CBLSyncConnection+Pull.m
+++ b/Source/CBLSyncConnection+Pull.m
@@ -164,7 +164,7 @@
     NSDictionary* attachments = nil;
     NSString* docID;
     NSData* json = request.body;
-    if (self.onSyncAccessCheck && memmem(json.bytes, json.length, "\"_attachments\":", 15) != NULL) {
+    if (self.onSyncAccessCheck || memmem(json.bytes, json.length, "\"_attachments\":", 15) != NULL) {
         NSDictionary* props = [NSJSONSerialization JSONObjectWithData: json options: 0 error: NULL];
         attachments = $castIf(NSDictionary, props[@"_attachments"]);
         docID = props[@"_id"];
--- a/Source/CBLSyncConnection+Pull.m
+++ b/Source/CBLSyncConnection+Pull.m
@@ -72,6 +72,8 @@
     // (Note: even if we got 0 changes (i.e. caught up) we still need to go through the db queue
     // before announcing it, so previously queued change processing blocks get to run first.)
     LogTo(Sync, @"Received %u changes", (unsigned)changes.count);
+    if (![self accessCheckForRequest: request docID: nil])
+        return;
     [request deferResponse];
     [self onDatabaseQueue: ^{
         CFAbsoluteTime time = CFAbsoluteTimeGetCurrent();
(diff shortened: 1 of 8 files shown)


Source/CBLSyncConnection.m
```diff
@@ -265,6 +266,22 @@ - (void) removeAttachmentProgress: (NSProgress*)attProgress
...
 
 
+- (BOOL) accessCheckForRequest: (BLIPRequest*)request
+                         docID: (NSString*)docID
+{
+    if (self.onSyncAccessCheck) {
+        CBLStatus status = self.onSyncAccessCheck(request, docID);
+        if (CBLStatusIsError(status)) {
+            NSString* message;
+            [request respondWithErrorCode: CBLStatusToHTTPStatus(status, &message)
```
snej: This is a race condition: it's not guaranteed that CBLStatusToHTTPStatus will have been called before the `message` parameter is pushed on the stack. You'll need to break the function call out and put it above this line.

Source/CBLSyncConnection+Pull.m
```diff
@@ -162,11 +164,17 @@ - (void) handleIncomingRevision: (BLIPRequest*)request {
     NSDictionary* attachments = nil;
     NSString* docID;
     NSData* json = request.body;
-    if (memmem(json.bytes, json.length, "\"_attachments\":", 15) != NULL) {
+    if (self.onSyncAccessCheck && memmem(json.bytes, json.length, "\"_attachments\":", 15) != NULL) {
```
snej: Why is `self.onSyncAccessCheck` added there? This test checks for whether there could be attachments in the doc; it doesn't depend on whether there's an access check registered.

snej: Actually I think it would be better to remove the `docID` parameter from the access check block. It isn't actually used. The reason for this `if` test is to avoid having to parse the doc's JSON unless necessary, for speed, and this patch would cause it to always get parsed, for no good reason, if the db is read-only.

pasin: I intended to use || not &&, that was the bug. I have docID there so that (#726) if we allow users to grant access via the listener delegate, the users can use doc ids for in their logic. We can chat about this next week.

snej:
> I have docID there so that (#726) if we allow users to grant access via the listener delegate, the users can use doc ids for in their logic.

That makes sense. We can optimize this by adding the docID as a property of the request so that it's quicker to get it. (And then we have to make sure that it matches the docID parsed from the JSON, in case the peer tries to cheat...)

PR #871, opened 2015-08-17
Fix #686 close database not closing background database

### LINKED ISSUE #686: The bg replication server's database not getting closed

pasin:
- Closed background database in CBLManager._forgetDatabase: method.
- Added CBLManager._closeDatabaseNamed: method to facilitate closing the opened database of the given name.
- Added Database_Tests.test24_CloseDatabase
- Made CBLDatabase.isOpen value change observable (used in the unit test)

#686

### DIFF
--- a/Source/API/CBLManager.m
+++ b/Source/API/CBLManager.m
@@ -726,9 +726,25 @@ static CBLManager* sInstance;
         return [repl localDatabase] == db;
     }];
     [_databases removeObjectForKey: name];
-    [_shared closedDatabase: name];
+
+    CBL_Shared* strongShared = _shared;
+    [strongShared closedDatabase: name];
+
+    // Close background database:
+    [strongShared.backgroundServer tellDatabaseManager: ^(CBLManager* bgmgr) {
+        NSError* error;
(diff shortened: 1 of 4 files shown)


pasin:
I have reimplemented this in CBLManager's _forgetDatabase: method as suggested in the previous PR. Thanks!

Source/API/CBLManager.m
```diff
@@ -726,10 +726,26 @@ - (void) _forgetDatabase: (CBLDatabase*)db {
...
+        NSError* error;
+        if (![bgmgr _closeDatabaseNamed: name error: &error])
+            Warn(@"Cannot close background database named %@: %@", name, error);
+    }];
 }
 
 
+- (BOOL) _closeDatabaseNamed: (NSString*)name error: (NSError**)outError {
+    CBLDatabase* db = _databases[name];
+    if (db)
```
snej: This doesn't initialize `*outError` if db is nil. It should at least set it to nil.

Issue #874, opened 2015-08-19
Some Unit tests when run from the device are failing.


ashvindersingh:
Some of the Unit tests when run on device are failing due to unable/trying to connect with syncgateway url 127.0.0.1 ipaddress. Instead they should be sending data to provisioned server in xcscheme file.

snej:
I don't think the tests have any hardcoded URLs; they should be honoring the `CBL_TEST_SERVER` and `CBL_SSL_TEST_SERVER` environment variables. (Except of course for P2P and listener tests that hardcode `localhost` because they run both client and listener in-process.)

Have you verified that the scheme is setting those environment variables?

snej:
In the commit above it looks like you enabled those variables but left them with their default values that point to `localhost`. For this to fix the on-device testing, you have to change the host of those URLs to point to the machine that's running the Sync Gateway.

ashvindersingh:
Hi @snej, I deliberately did it so that Unit tests can pass.  I'm replacing the 'localhost' with sync gateway URL in bots configuration. Now we have 7 tests which are failing with the device. The remaining replication tests are passing. I have discussed with @pasin. He is looking into it.

pasin:
There are 3 tests failed due to an invalid SSL certificate. To this with devices, the self-signed certificate needs to be regenerated with the name of the Sync-Gateway IP Address.

```
$openssl genrsa -out privkey.pem 2048
$openssl req -new -x509 -key privkey.pem -out cert.pem -days 1095 
<Use Sync-Gateway IP Address as the name>
$openssl x509 -inform PEM -in cert.pem -outform DER -out SelfSigned.cer
```
Replace TestData/SelfSigned.cer, Server/cert.pem, and Server/privkey.pem with the re-generated ones.

pasin:
For the other failures:
- Database_Tests.test23_ReplaceOldVersionDatabase (#875)
- Misc_Tests.test_AnonymousIdentity (Couldn't reproduce the failure)
- DatabaseInternal_Tests.test25_FileProtection (Pending investigation)

I have also found that IncrementalStore_test.test_DefaultConflictHandler failed sometime. I will have to investigate this.

snej:
>There are 3 tests failed due to an invalid SSL certificate.

Yup, I've been pointing this out for as long as the bot has been running (and failing) those tests.

pasin:
[...]

PR #876, opened 2015-08-19
Feature/issue 874 test environment

ashvindersingh:
- Enabled environment variables in iOS and Mac test schemes, so that tests can be run against any SyncGateway URL.

### DIFF
--- a/CouchbaseLite.xcodeproj/xcshareddata/xcschemes/CBL Mac Test App.xcscheme
+++ b/CouchbaseLite.xcodeproj/xcshareddata/xcschemes/CBL Mac Test App.xcscheme
@@ -76,12 +76,12 @@
          <EnvironmentVariable
             key = "CBL_TEST_SERVER"
             value = "http://localhost:4984/"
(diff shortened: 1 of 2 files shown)


snej:
I don't understand this commit — its effect looks like a no-op. You've just changed the schemes to set those environment variables, but the values specified in the schemes are the same as the default values that are used by the code if no environment variable is set. So this commit won't cause any change in the way the tests run. Could you explain what the rationale is?

snej:
Also, please reference the bug number in the pull request by putting `Fixes #874` in the comment. Otherwise Github and Waffle don't see a connection.

pasin:
I think the fix is to enabling the environment variables. If I understand correctly, the test bot has a script to replace the CBL_TEST_SERVER and CBL_SSL_TEST_SERVER with the correct IP address of the Sync-Gateway.

snej:
If the script is modifying the scheme's XML file, can't it set the `enabled` flags too?

PR #878, opened 2015-08-20
Fixed #875 Rename lowercase attachment files in CBLDatabaseUpgrade

### LINKED ISSUE #875: Support lowercase attachment blob filenames
Android 1.0.4 and .NET 1.1.0 have lowercase blob file names and iOS expected uppercase filename. Somehow when testing this with simulator, everything works file even with lowercase file name but this is not the case on the real devices.

pasin:
CBL Android 1.0.4 and .NET 1.1.0 have lowercase attachment file names. This method will detect that and uppercase the file names if needed. This makes the Database_Tests.test23_ReplaceOldVersionDatabase passed on the real device.

#875

### DIFF
--- a/Source/CBLDatabaseUpgrade.m
+++ b/Source/CBLDatabaseUpgrade.m
@@ -182,51 +182,44 @@ static int collateRevIDs(void *context,
         return CBLStatusFromNSError(error, kCBLStatusAttachmentError);
 
     BOOL success = YES;
-    NSMutableArray* processed = [NSMutableArray array];
-    for (NSString* fileName in content) {
-        if (![[fileName pathExtension] isEqualToString:@"blob"])
+    NSMutableDictionary* renDict = [NSMutableDictionary dictionary];
+    for (NSString* oldFileName in content) {
+        if (![[oldFileName pathExtension] isEqualToString:@"blob"])
             continue;
-
-        NSString* filePath = [dir stringByAppendingPathComponent: fileName];
-        NSString* newPath = [self uppercaseAttachmentFileNameAtPath: filePath];
-
+        
+        NSString* newFileName = [[[oldFileName stringByDeletingPathExtension] uppercaseString]
+                                    stringByAppendingPathExtension: @"blob"];
         // Assume no both lowercase and uppercase filename mixed.
-        // Skip the renaming process if detecting that the file is already
-        // uppercased:
-        if ([filePath isEqualToString: newPath])
+        // Skip the renaming process if detecting that the file name is already uppercase:
+        if ([newFileName isEqualToString: oldFileName])
             break;
-
-        if ([fmgr moveItemAtPath: filePath toPath: newPath error: &error]) {
-            [processed addObject: filePath];
+        
+        NSString* oldPath = [dir stringByAppendingPathComponent: oldFileName];
+        NSString* newPath = [dir stringByAppendingPathComponent: newFileName];
+        if ([fmgr moveItemAtPath: oldPath toPath: newPath error: &error]) {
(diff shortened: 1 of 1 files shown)


Source/CBLDatabaseUpgrade.m
```diff
@@ -163,7 +163,70 @@ - (CBLStatus) moveAttachmentsDir {
...
+        return CBLStatusFromNSError(error, kCBLStatusAttachmentError);
+
+    BOOL success = YES;
+    NSMutableArray* processed = [NSMutableArray array];
+    for (NSString* fileName in content) {
+        if (![[fileName pathExtension] isEqualToString:@"blob"])
+            continue;
+
+        NSString* filePath = [dir stringByAppendingPathComponent: fileName];
+        NSString* newPath = [self uppercaseAttachmentFileNameAtPath: filePath];
```
snej: This code appends the filename to the path, extracts the filename back out of the path, uppercases it, truncates the filename from the path, and adds the new filename. Couldn't you just uppercase the filename _before_ adding it to the path?

Source/CBLDatabaseUpgrade.m
```diff
@@ -163,7 +163,70 @@ - (CBLStatus) moveAttachmentsDir {
...
+            [processed addObject: filePath];
+        } else {
+            Warn(@"Upgrade failed: Cannot rename attachment file %@: %@", filePath, error);
+            success = NO;
+            break;
+        }
+    }
+    
+    if (!success) {
+        // Backing out if there is an error found:
```
snej: Is it necessary to back out the incomplete changes if there's an error? I don't see how they could do any harm.

Source/CBLDatabaseUpgrade.m
pasin: Yes, I could optimize this.

Source/CBLDatabaseUpgrade.m
pasin: Because I made an assumption that there should not be both uppercase and lowercase file names in the attachments folder. In the for-loop, if it found that there is already an uppercase file name, it will break out of the loop and finish the renaming process. So if there was an error in the middle of processing, without backing-out the renaming process will not be executed again next time.

pasin: Actually it depends. If the attachment folder was copied to the new location instead of moved, there is no need to backing that out.

pasin: One option would be putting an empty file (e.g. ".rename") to mark as in renaming process. The file will be removed if the renaming is done. Next time, we could check .rename file along with the logic to check whether there is already an uppercase file name.

pasin:
I replaced the renaming the file path logic with just renaming file name & attaching to the dir path. I still keep the backing out logic but instead of using array, use dictionary to keep both new and old file names (storing file names not path to save some memory space).

Source/CBLDatabaseUpgrade.m
```diff
@@ -163,7 +163,63 @@ - (CBLStatus) moveAttachmentsDir {
...
+            [renDict setObject: newFileName forKey: oldFileName];
+        } else {
+            Warn(@"Upgrade failed: Cannot rename attachment file from %@ to %@: %@",
+                 oldPath, newPath, error);
+            success = NO;
+            break;
+        }
+    }
+    
+    if (success && [renDict count] > 0) {
```
snej: That should be `!success`.

pasin: Sorry, I removed ! while manually debugging to check if the logic works or not but forgot to put it back.

pasin:
I have fixed the !success mistake. Could you please re-review when you have time?

PR #888, opened 2015-08-22
Add support for sort descriptor using comparator.

ufosky:
The comparator block is used more convenience than selector.
Fix an exception issue. See https://github.com/couchbase/couchbase-lite-ios/issues/890


snej:
Thanks!

PR #891, opened 2015-08-24
Encrypt attachment store if db encrypted, and fix existing unencrypted stores

snej:
Fixes a serious security problem: Attachments in an encrypted database didn't get encrypted. :(

The bug was simply that when CBLDatabase instantiated its CBL_BlobStore, it didn't set the blob-store's
encryption key. The first commit fixes that, and has a unit test to verify that when an encrypted database is created, the raw attachment blob files are encrypted on disk.

The harder part is fixing existing databases. We have to (a) detect an existing encrypted database with
an unencrypted attachment store that hasn't been fixed yet, and (b) rewrite all the blob files with encryption.

Part (a) is done by adding a file called `_encryption` to the BlobStore directory when encryption is being used. The contents of the file name the encryption algorithm, currently always `AES`. Existing databases won't have this file, which indicates that their attachments aren't encrypted. So when opening an existing BlobStore, if an encryption key is given but the `_encryption` file doesn't exist, we've encountered this bug and need to encrypt the files.

Part (b) is done by the new `-changeEncryptionKey:error:` method I started working on last Friday. This is a general purpose method to add/remove/change the encryption key. It creates a new BlobStore in a new directory with the new key, writes all the blobs into it, then replaces the old directory with the new one. This uses more disk space during the operation, but makes it possible to recover safely if there's an error or crash partway through.

There are new unit tests that test upgrading the buggy BlobStore, as well as all three modes of changing the encryption key.

### DIFF
--- a/Source/CBL_BlobStore+Internal.h
+++ b/Source/CBL_BlobStore+Internal.h
@@ -1,0 +1,25 @@
+//
+//  CBL_BlobStore+Internal.h
+//  CouchbaseLite
+//
+//  Created by Jens Alfke on 8/23/15.
+//  Copyright © 2015 Couchbase, Inc. All rights reserved.
+//
+
+#import "CBL_BlobStore.h"
+
+
+// Name of file in blob-store dir that records encryption type used (currently "AES")
+#define kEncryptionMarkerFilename @"_encryption"
+
+
+@interface CBL_BlobStore ()
+
+- (NSString*) rawPathForKey: (CBLBlobKey)key;
+@property (readonly, nonatomic) NSString* tempDir;
+@property (readonly) CBLSymmetricKey* encryptionKey;
+
+#if DEBUG
+- (BOOL) markEncrypted: (BOOL)encrypted error: (NSError**)outError; // exposed for testing ONLY
+#endif
+@end
--- a/Source/CBL_BlobStore.m
+++ b/Source/CBL_BlobStore.m
@@ -417,28 +417,30 @@
 
 
 - (NSString*) createTempDir: (NSError**)outError {
+    NSError* error;
+    NSString* tempDir;
 #ifdef GNUSTEP
     NSString* name = $sprintf(@"CouchbaseLite-Temp-%@", CBLCreateUUID());
-    NSString* tempDir = [NSTemporaryDirectory() stringByAppendingPathComponent: name];
+    tempDir = [NSTemporaryDirectory() stringByAppendingPathComponent: name];
     NSDictionary* attrs = @{NSFilePosixPermissions: @(0700)};
     if (![[NSFileManager defaultManager] createDirectoryAtPath: tempDir
                                    withIntermediateDirectories: YES
                                                     attributes: attrs
-                                                         error: outError])
-        return nil;
-    return tempDir;
+                                                         error: &error])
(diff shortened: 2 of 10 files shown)


Source/CBL_BlobStore+Internal.h
```diff
@@ -0,0 +1,25 @@
...
+
+@interface CBL_BlobStore ()
+
+- (NSString*) rawPathForKey: (CBLBlobKey)key;
+@property (readonly, nonatomic) NSString* tempDir;
+@property (readonly) CBLSymmetricKey* encryptionKey;
+
+#if DEBUG
+- (BOOL) markEncrypted: (BOOL)encrypted error: (NSError**)outError; // exposed for testing ONLY
+#endif
```
pasin: I saw -markEncrypted: was used in regular codes to indicate encryption type in the blob store folder. I'm confused with #if DEBUG and the comment 'exposed for testing ONLY'. Should #if DEBUG be removed?

Source/CBL_BlobStore.m
```diff
@@ -263,26 +323,122 @@ - (NSInteger) deleteBlobsExceptMatching: (BOOL(^)(CBLBlobKey))predicate
...
-        NSURL* tempDirURL = [[NSFileManager defaultManager] 
-                                                 URLForDirectory: NSItemReplacementDirectory
-                                                 inDomain: NSUserDomainMask
-                                                 appropriateForURL: parentURL
-                                                 create: YES error: &error];
-        _tempDir = [tempDirURL.path copy];
-        Log(@"CBL_BlobStore %@ created tempDir %@", _path, _tempDir);
-        if (!_tempDir)
-            Warn(@"CBL_BlobStore: Unable to create temp dir: %@", error);
+    NSError* error;
```
pasin: I think this variable can be removed.

Source/CBL_BlobStore.m
```diff
@@ -263,26 +323,122 @@ - (NSInteger) deleteBlobsExceptMatching: (BOOL(^)(CBLBlobKey))predicate
...
+    NSURL* parentURL = [NSURL fileURLWithPath: _path isDirectory: YES];
+    NSURL* tempDirURL = [[NSFileManager defaultManager] URLForDirectory: NSItemReplacementDirectory
+                                                               inDomain: NSUserDomainMask
+                                                      appropriateForURL: parentURL
+                                                                 create: YES
+                                                                  error: outError];
+    if (!tempDirURL)
+        Warn(@"CBL_BlobStore: Unable to create temp dir: %@", error);
```
pasin: Should be outError instead?

pasin:
I'm done reviewing. Just have a couple minor comments.

Source/CBL_BlobStore+Internal.h
snej: It's a private method of CBL_BlobStore. _Outside_ that class, it should only be used for testing. It's only the declaration of the method in a header that's ifdef'd.

Source/CBL_BlobStore.m
snej: outError might be NULL according to Obj-C calling conventions, so I added `error` so I could always report the error. But then I forgot to use `&error` in the calls above. I'll fix that.

PR #907, opened 2015-09-08
add UnitTest that crashes CBLIS when fetch groupby

mastohhh:
I think the problem is : `CBLIncrementalStore` handles `propertiesToFetch` as an array of `NSString`, but Apple says that it can be an array of `NSPropertyDescriptions`.

```objc
/* Specifies a collection of either NSPropertyDescriptions or NSString property names
that should be fetched. The collection may represent attributes, to-one relationships, 
or NSExpressionDescription.  If NSDictionaryResultType is set, the results of the fetch
will be dictionaries containing key/value pairs where the key is the name of the 
specified property description.  If NSManagedObjectResultType is set, then 
NSExpressionDescription cannot be used, and the results are managed object faults 
partially pre-populated with the named properties */
@property (nonatomic, copy) NSArray *propertiesToFetch NS_AVAILABLE(10_6, 3_0);
```

### DIFF
--- a/Unit-Tests/IncrementalStore_Tests.m
+++ b/Unit-Tests/IncrementalStore_Tests.m
@@ -1720,6 +1720,49 @@ static NSArray *CBLISTestInsertEntriesWithProperties(NSManagedObjectContext *con
     }];
 }
 
+- (void) test_FetchWithGroupBy {
+    NSError *error;
+
+    NSDictionary *entry1 = @{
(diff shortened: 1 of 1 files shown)


pasin:
Beside the crash issue, currently CBLIS doesn't support group-by. Is this the feature that you currently need?

pasin:
I have fixed the crash issue but the group by feature is still left unsupported.

PR #936, opened 2015-10-05
ForestDB auto-compact and better encryption (#922)

### LINKED ISSUE #922: No auto-compact (db.forest getting very large)
App configured with SQLite:

cblite files gets to 10MB
(if the file size gets excessive, database compact ensures it becomes small again)

App configured with ForestDB:

db.forest gets to 200MB+
calling database compact function results in db.forest.1, db.forest.2 copies that don't get removed automatically

Note: because the app only syncs one way (server to app with own json api), the max revision depth tree is set to 1.

snej:
This enables auto-compaction for ForestDB databases, and switches to the new ForestDB encryption API (as opposed to the one I implemented outside ForestDB as a plug-in.)

Please see the commit notes for the individual commits. It's probably best to review the two commits separately, as they do different things. (They're dependent on each other, which is why I'm submitting them as one PR.)

Fixes #922


snej:
I don't think there's anything we can do to abort a compact, but I'll ask the ForestDB guys.

My understanding is that, if the process quits/crashes before a compact is complete, ForestDB will delete the partially-written compacted file next time it opens the DB.

pasin:
I'm merging it now. If ForestDB requires extra care for a partial compacted file, we could enhance it later.

Issue #938, opened 2015-10-06
CBLIS - Very poor performances from 5000 documents locally


mastohhh:
Despite the optimizations @pasin did here : https://github.com/couchbase/couchbase-lite-ios/issues/793, I still encounter performance issues.

I really don't know what's the best practice to use both `CBLManager` and `NSManagedObjectContext`. Here's what I tried : 

## 1 - Parent Context - Private context's queue : 

- A main context that has a parent context
- A private context parent of the main, and with a coordinator.
- `dispatchQueue` of `CBLManager` is the same as the private context's queue. 

This is the one I actually use.

The problem with this one is `dispatch_get_current_queue();` will be deprecated. And it's not really fast.

## 2 - Parent Context - Background Queue

- A main context that has a parent context
- A private context parent of the main, and with a coordinator.
- `dispatchQueue` of `CBLManager` is `dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_BACKGROUND, 0)`.

It gives sometimes crashes like :  `The FMDatabase <CBL_FMDatabase: 0x7ae753c0> is currently in use.`
`*** Assertion failure in -[CBL_FMDatabase warnInUse], FMDatabase.m:360`

## 3 - Main Context - Main Thread : 

- Only one context in the main thread with the coordinator.
- `dispatchQueue` of `CBLManager` is by default, main thread.

This is very slow

## 4 - Main Context - Background Queue : 

- Only one context in the main thread with the coordinator.
- `dispatchQueue` of `CBLManager` is `dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_BACKGROUND, 0)`.

Every calls to `performBlock:` fails.
Slow and crashes like :  `The FMDatabase <CBL_FMDatabase: 0x7ae753c0> is currently in use.`

## 5 - Confinment Context - Background Queue

- `dispatchQueue` of `CBLManager` is `dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_BACKGROUND, 0)`.
- Only one context `NSConfinementConcurrencyType` (deprecated) in the `dispatchQueue` of `CBLManager`.

Error message `Can only use -performBlock: on an NSManagedObjectContext that was created with a queue.`

## Conclusion

Performance goes down as the global number of documents increase.

[...]

PR #953, opened 2015-10-19
Add CBLIS Performance UnitTests for save-fetch-inform

mastohhh:
Related to this issue : https://github.com/couchbase/couchbase-lite-ios/issues/938

@pasin : If you don't already did it : here are some examples of performance tests. I hope it will help.

### DIFF
--- a/Unit-Tests/IncrementalStore_Tests.m
+++ b/Unit-Tests/IncrementalStore_Tests.m
@@ -1944,6 +1944,178 @@ static NSArray *CBLISTestInsertEntriesWithProperties(NSManagedObjectContext *con
     AssertEq((int)result.count, 0);
 }
 
+#pragma mark - Performance
+
+- (void) test_PerformanceSave {
+    if (NO == self.isSQLiteDB) return;
+
+    [self measureMetrics:[[self class] defaultPerformanceMetrics] automaticallyStartMeasuring:NO forBlock:^{
+
+        for (NSInteger i = 1; i < 1000; i++) {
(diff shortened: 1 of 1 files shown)


pasin:
Now I'm testing several things on my little experimental app. But this is very helpful. Thanks a lot.

mastohhh:
This is my firsts performance tests. I didn't find where we can profile while testing. Too bad. I don't test with an Xcode server, maybe it warns when performance goes down.

snej:
XCTest can measure performance by timing how long a block takes to run. Xcode can track that performance over time. See the XCTest documentation, or look at -[JSON_Tests test_CBJSONEncoderBenchmark] as a simple example.

pasin:
Just read the difference b/w measureBlock and measureMetrics. They both are doing the same thing which is running the block 10 times to measure the performance. MeasureMetrics provides more fine grains when capturing with performance with startMeasuring: and stopMeasuring:.

pasin:
I'm going to pull your changes in and modify it a little bit. Also I think it's better to disable the performance test by default as it slows down the overall unit test running.

mastohhh:
MeasureMetrics also add ability to monitor asynchronous work, as you can see in the performanceInform test.

mastohhh:
You can also reduce the number of objects.

I think these tests are good just to be sure of improvements we made. Not just a feeling.

pasin:
I have just manually merged your changes and made some minor modifications.

https://github.com/couchbase/couchbase-lite-ios/pull/956

pasin:
Anyway, thanks a lot for your contribution.

PR #959, opened 2015-10-21
Fixes #957

### LINKED ISSUE #957: test09_Rekey Invalid
Part of the test appears to check if a query survives a rekey, but it is actually testing nothing.  The view and documents in that view are created in `db`, which is untouched during the entire test, as opposed to `seekrit` which is the database being rekeyed.  If the test is changed to create the view and documents in `seekrit` then the test fails because as part of closing the database the map function on the view is lost.

borrrden:
Fixes #957 by altering unit test


snej:
Thanks for fixing it!

Issue #966, opened 2015-10-30
CBLIS : Data loss


cflorion:
We have had a problem since we started working with Couchbase-lite-iOS. We have crashes and data loss, but we kind of had saved that for later since we had more important bugs to deal with. But no we have been kind of stuck for the last few weeks/months since we can't put our app in production for two reasons: the performance on the device are not good enough (that's the another problem, the next) and the crashes/data loss mentioned above. We are out of ideas, so maybe someone has some for us :)

(We use CBLIncrementalStore, CBL-iOS (latest commit on the master branch) and CB Sync Gateway 1.1.1).

1. We have a Core Data Model. In the model, we use  Core Data object validation (https://developer.apple.com/library/ios/documentation/Cocoa/Conceptual/CoreData/ObjectValidation.html). We use it the simple way by just unchecking the "optional" box on properties and relationships of Entities in the Model Editor UI in Xcode. This is quite handy since the app crashes (and produces a detailed exception) when one of theses constraints are not fulfilled. But the problem is that we have random crashes telling us that some relationship object is missing. For example we have a Tax entity. We tried to look for problems in our implementation but we realized that in our app we have no way nor code to delete a Tax managed object in our code...

2. Sometimes the app crashes because an object is saved with all its properties being nil. The save is therefore prevented by the Core Data Object Validation. We relaunch the app, and all seems fine again after a replication.

3. One more example. We have another to-many relationship (users) where we forgot to uncheck the "optional" box. So this obviously doesn't produce crashes. But this is via our app UI that we noticed the problem. While debugging several times, once we saw that all users were missing in the database (via a core data fetch request) and another time only one user was missing.

We are trying to find a pattern to reproduce this bug, but this is quite difficult. It looks like it gets worse overtime, but not sure of it... We're digging.

Could this be related to the number of revisions that is too high ?
Have you ever heard something like this ?

Thanks! 
:)

pasin:
> 1. But the problem is that we have random crashes telling us that some relationship object is missing. For example we have a Tax entity. We tried to look for problems in our implementation but we realized that in our app we have no way nor code to delete a Tax managed object in our code...

Would it be possible to scope this down? For example, can you make a small simple app that I can use to reproduce the missing relationship or crash (even though I need to run multiple times to reproduce)?

> 2. Sometimes the app crashes because an object is saved with all its properties being nil. The save is therefore prevented by the Core Data Object Validation. We relaunch the app, and all seems fine again after a replication.

[...]

Issue #958, opened 2015-10-21
CBLIS doesn't handle accesses from multiple dispatch queues


pasin:
When using multiple CoreData contexts (each context can have different dispatch queue), CBLIS could be accessed by different dispatch queue other than the root context's dispatch queue. CBLIS needs a way to handle that. One easy way is to always use root context's dispatch queue to operate.

cflorion:
Hmmm interesting.
I'm trying to understand.

Let's take the common example of a multiple CoreData contexts setup with one mainContext (for the UI, on the main queue) and a parentContext (on a background queue, for importing data).

- When the user touches the "create stuff" button on the screen, a Stuff ManagedObject is created in the mainContext (with a temporary ObjectID at this point). We save the mainContext, the objectID is now permanent, we continue with a save on the rootContext. The rootContext is wired to the persistentStore, so when a NSManagedObject is inserted, the CBLIncrementalStore created a CBLDocument and stores it to disk.

- When data comes from a replication. The data is inserted in the CBLDatabase, the rootContext is informed of the changes by the IncrementalStore. At this point the mainContext is unaware of the changed since in this kind of stack, the changes go from the child to the parent, but not from the parent to the child. So the CBL created a NSNotification in order to inform the mainContext. And now everything is up to date.

I may have misunderstood but in the example I'm giving, only the rootContext talks directly to the CBLDatabase. The CBLIS is only talking to the mainContext when sending the notification (on the main queue). In my opinion only the root context's dispatch queue should be used.

But, again, I might have misunderstood.

pasin:
Yeah, that's what I first understood too. 

However, while trying to fix save operation performance with parent (private queue) and child (main queue) context model, I have noticed that the main context (via NSFetchRemoteController) requests to the CBLIS to resolve the entity properties and relationships - more specifically, the following two methods.

```
- (NSIncrementalStoreNode*) newValuesForObjectWithID: (NSManagedObjectID*)objectID
                                         withContext: (NSManagedObjectContext*)context
                                               error: (NSError**)outError;

- (id) newValueForRelationship: (NSRelationshipDescription*)relationship
               forObjectWithID: (NSManagedObjectID*)objectID
                   withContext: (NSManagedObjectContext*)context
                         error: (NSError**)outError;
```

[...]

Issue #903, opened 2015-09-04
Problem with relationship one-to-many when documents are created simultaneously on multiple devices


mastohhh:
I don't know how UnitTest this, so I'll try to explain my problem here.

I have an `Entry` document that has many `Subentries`.

To display all subentries of an entry, I use a `NSFecthedResultsController` on `Subentry` entity with this predicate : 

```
[NSPredicate predicateWithFormat:@"entry == %@", entry];
```

When the controller triggers delegate with : 

```
- (void)controllerDidChangeContent:(NSFetchedResultsController *)controller
```

I can loop on `controller.fetchedObjects` or `entry.subentries`. This works fine while subentries are created one after an other on each devices.

But when I create them simultaneously,  `controller.fetchedObjects` contains good subentries, but `entry.subentries` is not updated. One subentry is missing on each device.

I'm using `master` branch of CouchbaseLite and NSIncrementalStore.

I would like to UnitTest it for you but I didn't find a similar one that simulates 2 devices.

pasin:
Thanks for reporting the issue. Normally when we need to simulate a behavior that involves syncing between 2 devices, we create 2 different CBL databases replicating to the same remote sync-gateway URL. If you are interested in creating the unit test for this, you could make it that way.

Anyway, I would like to make sure that I understand the issue correctly. 

1. Does this mean that 2 subentries were synced correctly into both devices?

2. I understand that calling `controller.fetchedObjects` will execute a fetch request. Does this mean that fetching with the predicate `@"entry == %@", entry` on the `Subentry` entity yielding the correct results?

3. Does entry.subentries miss an entry created on another device?

mastohhh:
This is it. But in the next runloop everything comes back to normal. 

I think the FRC is triggered just before the relationship is filled. 

If you can't reproduce that, I'll try a UnitTest.

mastohhh:
I have more info for you.

I disabled cache that was introduced with : https://github.com/couchbase/couchbase-lite-ios/commit/4f852bc1d53337ee23a3d312be51cc777eda02af

(I just commented the line `_relationshipCache = [[NSCache alloc] init];`)

It works way better. I think this could explain my problem : when FRC is triggered with new entries, the `entry.subentries` calls cache instead of querying again.

I hope this will help you !

pasin:
Thanks a lot for the info and your patient. I have been working on the other issue, and I will tackle this issue this week.

mastohhh:
[...]

PR #969, opened 2015-11-03
CBLIS do not inform contexts just refresh registered objects

mastohhh:
related to https://github.com/couchbase/couchbase-lite-ios/issues/938

- No need to merge contexts with notification.
- Separate inserted and updated objects to refresh relationships only for inserted objects
- refresh objects only if object is registered

### DIFF
--- a/Source/API/Extras/CBLIncrementalStore.m
+++ b/Source/API/Extras/CBLIncrementalStore.m
@@ -56,10 +56,6 @@ static NSError* CBLISError(NSInteger code, NSString* desc, NSError *parent);
 - (NSString*) couchbaseLiteIDRepresentation;
 @end
 
-@interface CBLView (CBLIncrementalStore)
-@property (readonly) SInt64 lastSequenceChangedAt;
-@end
-
 
 @interface CBLIncrementalStore ()
 
@@ -1628,7 +1624,7 @@ static CBLManager* sCBLManager;
     if (view) {
         NSString* cacheKey = [NSString stringWithFormat:@"%@/%@", view, parentKey];
         CBLQueryEnumerator* result = [_relationshipCache objectForKey: cacheKey];
-        if (result && (SInt64)result.sequenceNumber == view.lastSequenceChangedAt)
+        if (result && (SInt64)result.sequenceNumber == view.database.lastSequenceNumber)
             return [result allObjects];
 
         CBLQuery* query = [view createQuery];
--- a/Source/API/Extras/CBLIncrementalStore.m
+++ b/Source/API/Extras/CBLIncrementalStore.m
@@ -61,8 +61,10 @@ static NSError* CBLISError(NSInteger code, NSString* desc, NSError *parent);
 
 // TODO: check if there is a better way to not hold strong references on these MOCs
 @property (nonatomic, strong) NSHashTable* observingManagedObjectContexts;
+@property (nonatomic, weak, readonly) NSManagedObjectContext* rootContext;
 @property (nonatomic, strong) CBLDatabase* database;
 @property (nonatomic, readonly) NSUInteger maxRelationshipLoadDepth;
+@property (nonatomic) BOOL shouldNotifyLocalDatabaseChangesToContexts;
 
 @end
 
@@ -84,7 +86,9 @@ static NSError* CBLISError(NSInteger code, NSString* desc, NSError *parent);
 @synthesize conflictHandler = _conflictHandler;
 @synthesize customProperties = _customProperties;
 @synthesize observingManagedObjectContexts = _observingManagedObjectContexts;
+@synthesize rootContext = _rootContext;
 @synthesize delegate = _delegate;
+@synthesize shouldNotifyLocalDatabaseChangesToContexts = _shouldNotifyLocalDatabaseChangesToContexts;
 
 static CBLManager* sCBLManager;
 
@@ -322,7 +326,7 @@ static CBLManager* sCBLManager;
 
     // Setup database change notification:
     [[NSNotificationCenter defaultCenter] addObserver: self
-                                             selector: @selector(documentsChanged:)
+                                             selector: @selector(databaseChanged:)
                                                  name: kCBLDatabaseChangeNotification
                                                object: self.database];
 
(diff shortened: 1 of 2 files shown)


mastohhh:
There's good performance and no crash but sometimes contexts are not up to date. I maybe missed something.

mastohhh:
I'll try this code with mergeContext back...

Source/API/Extras/CBLIncrementalStore.m
```diff
@@ -2144,102 +2150,41 @@ - (void) mergeContextsWithUpdatedIDs: (NSArray*)updatedIDs deletedIDs: (NSArray*
...
-    for (NSManagedObjectContext *context in self.observingManagedObjectContexts) {
-        if (context == self.rootContext)
-            continue;
-        
-        [context performBlock:^{
-            NSManagedObject *mObj = [context objectRegisteredForID: moID];
-            if (!mObj) return;
-            
-            INFO(@"firing fault : %@ in %@ on %@", moID, context, [NSThread currentThread]);
-            [mObj willAccessValueForKey: nil];
```
pasin: I understand that calling willAccessValueForKey: with nil will make sure that the object will be firing fault.

From Apple Doc of the willAccessValueForKey: method :
> You can invoke this method with the key value of nil to ensure that a fault has been fired, as illustrated by the following example.
>
> [aManagedObject willAccessValueForKey:nil];

pasin:
I just have a chance to closely look at your PR. It's interesting that merging the context with notifications may not be required. When I experimented with only merging with notification option, it seems to me that all contexts got refreshed correctly but NOT all the cases for NSFetchResultController especially the case when sub-entities were updated. Let me know if you found anything.

Anyway what was the result of using my fix alone? What does work or doesn't work?

mastohhh:
I tried your last commit on that branch. Contexts were not good updated.

So I tried to use `mergeChangesFromContextDidSaveNotification:` only. On both contexts. With to-one relationships added to `uidatedIDs`. Nothing was updated.

So I tried `refreshObject:mergeChanges:` only : and I saw that contexts were updated. (Note that PR is outdated, I already improved my code. For example : here I called `refreshObject:mergeChanges:` on deleted objects that causes bugs)

I only use `NSFetchedResultsController` in my app.

I think I misunderstood `fault`. That does not mean : object is live in context or not. I think what we want here is to know if an object is `registered` by the context and needs to be refreshed or not. Objects that are not `registered` by a context is not fetched and does not exists in context's memory, so we don't need to refresh it because it will be fetch with good values later.

My last problem here is : https://github.com/couchbase/couchbase-lite-ios/issues/903 is back.

pasin:
Would it be possible to push the updated?

mastohhh:
As I said #903 is back.

pasin:
Yesterday I chatted with @zgramana about #903 and @zgramana suggested me to look at processPendingChanges method of the NSManagedContext. You may try to call that at the end after refreshing the objects and see if it helps.

Some discussions:
http://theocacao.com/document.page/198
http://stackoverflow.com/questions/5719326/difference-between-save-and-processpendingchanges-in-coredata

mastohhh:
I thought about calling save: ! But that made no sense. I'll try your idea tomorrow. Thanks

mastohhh:
This works way better. I thought #903 was fixed. But unfortunately (after hours of tests) I saw the problem occur.

I push it and I hope you'll have an other idea.

pasin:
How hard to reproduce the issue? Do you know exact steps so I could try to do something similar?

Anyway I will grab your current code, run on my test app, and see if I spot anything.

mastohhh:
As I commented this PR code, I think #903 could come from replications that gives documents by batches, and objects and their relationships can come in separate batches of documents.

A CoreData save operation validates object graph (with optional etc..) before writing data to persistent store. A CoreData save operation creates/updates/deletes a set of documents that "make sense" (for example an entity with all his subentities).

But when documents goes through `CBLReplication` layer, they are just separate documents !

So, when an other device gets the `_changes` feed, how can we be sure we get a batch of data that "made sense" with a CoreData Save operation ?

If I receive the entity first, I refresh objects in contexts so his `subentities` relationship is empty. Then, when I receive "subentities", the couchbase-lite-ios layer doesn't tell us that the previous entity is updated. That's why I tried to get the toOne relationships to refresh them, in this example I get the entity.

As we are in a PR that talks about performance : I think this is necessary only for inserted objects. That's why I commented this code with that question : when we process `kCBLDatabaseChangeNotification` with every `CBLDatabaseChange` how do we know that a document is in insertion ?

pasin:
Let me try to answer your questions first.

> So, when an other device gets the _changes feed, how can we be sure we get a batch of data that "made sense" with a CoreData Save operation ?

If I remembered correctly the max batch size is 100 (optimize memory usage), if pulling more than 100 related documents, then not all the related documents will come on the same batch. 

> when we process kCBLDatabaseChangeNotification with every CBLDatabaseChange how do we know that a document is in insertion ?

I asked the same question to @snej a few week ago. Here is the answer for this:
> You could check whether any of the parent revisions have bodies Or rather, if any other revisions (not just parents). That would be pretty expensive though.

I saw that you tried to detect that by checking if the parent rev is nil or not, which is correct but may not cover all the cases. However if you refresh both update and insert objects, then your logic might be enough.

Source/API/Extras/CBLIncrementalStore.m
```diff
@@ -2144,102 +2150,65 @@ - (void) mergeContextsWithUpdatedIDs: (NSArray*)updatedIDs deletedIDs: (NSArray*
...
-            NSManagedObject* mObj = [strongRootContext objectRegisteredForID: moID];
-            if (mObj) {
-                [updated addObject: mObj];
-            } else {
-                mObj = [strongRootContext objectWithID: moID];
-                [inserted addObject: mObj];
+    NSArray *sortedContextRootBeforeMain = [self.observingManagedObjectContexts.allObjects sortedArrayUsingDescriptors:@[[NSSortDescriptor sortDescriptorWithKey:NSStringFromSelector(@selector(parentContext)) ascending:YES]]];
+
+    for (NSManagedObjectContext *context in sortedContextRootBeforeMain) {
+        [context performBlock:^{
```
pasin: This doesn't guarantee that the code in root context performBlock: will get executed before the main's performBlock: because both of the block will get executed in their own dispatch queue.

Explicitly call performBlock: on the root context and inside the block, at the end looping to the other context and do the performBlock.

pasin:
I have added one more comment to the issue that I have found today while running some tests on your PR. Please check.

mastohhh:
Ok I tried what you said : refresh both updated and inserted objects. I'll tell you if #903 is fixed as soon as my tests ends.

Thanks

pasin:
As I have commented, the root context is not refreshed before the main context. You may have to take care that as well. Please let me know the result. Thanks!

mastohhh:
#903 never gets back.

Refreshing to-one relationship for every inserted and updated objects fixes everything but gives a little more work.

The next optimization will be possible if `kCBLDatabaseChangeNotification` gives us inserted objects separate from updated objects.

Now I'm gong to work on https://github.com/couchbase/couchbase-lite-ios/issues/958 and https://github.com/couchbase/couchbase-lite-ios/issues/966. Those two last bugs doesn't allow us to deliver.

Thanks

pasin:
Glad to hear that. I will review all the code this week and let you know. Thank you very much for all your help and contribution.

Source/API/Extras/CBLIncrementalStore.m
```diff
@@ -2057,103 +2090,122 @@ - (void) processCouchbaseLiteChanges: (NSArray*)changes {
...
+                    if (rel.toMany)
+                        continue;
+
+                    NSRelationshipDescription *invRel = rel.inverseRelationship;
+                    if (!invRel)
+                        continue;
+
+                    NSArray *objIDs = [mObj objectIDsForRelationshipNamed:relName];
+
+                    [refreshedIDs addObjectsFromArray:objIDs];
```
pasin: Based on my testing, this can cause creating a duplicate mObj with a new ID on the destination managed object. I'm not sure why that happens.

Source/API/Extras/CBLIncrementalStore.m
```diff
@@ -2057,103 +2090,122 @@ - (void) processCouchbaseLiteChanges: (NSArray*)changes {
...
-        [self informManagedObjectContext: context
-                              updatedIDs: updatedObjectIDs deletedIDs: deletedObjectIDs];
+    if (updatedIDs.count > 0 || deletedIDs.count > 0) {
+        [self refreshContextsWithUpdatedIDs:updatedIDs deletedIDs: deletedIDs];
+
+        [[NSNotificationCenter defaultCenter]
+         postNotificationName: kCBLISObjectHasBeenChangedInStoreNotification
+         object: self
+         userInfo: @{ NSUpdatedObjectsKey: updatedIDs,
+                      NSDeletedObjectsKey: deletedIDs }];
```
pasin: Notification should be fired after all refresh operations are finished. Note that refreshContextsWithUpdatedIDs: deletedIDs: is actually done in the background.

pasin:
I just have a chance to come back and review the code. I like the idea but currently I have found a couple issues on my simple test app. I have provided my additional comments.

Source/API/Extras/CBLIncrementalStore.m
pasin: Actually the duplicate object is the relationship object. I think NSFetchResultController thinks that the refreshed one is the new one.

pasin: The issue was the the NSFetchResultController gets the new created object right after the main context gets saved and it doesn't update with the object with the new permanent ID after the root context gets saved. Still not sure how to solve the issue. 

@mastohhh have you seen anything like this?

mastohhh: I don't have any FRC problem. I'm using https://github.com/wtmoose/TLIndexPathTools to handle collectionView updates.

pasin:
I did a hand merge, fix conflicts, and issues that I have commented. I squashed multiple @mastohhh 's commits into one.

Issue #988, opened 2015-12-04
Revision reset code sample


tleyden:
## Problem

The scenario is this:

* Couchbase Lite is set to do a pull replication only
* Documents are only changed on Sync Gateway, not on Couchbase Lite
* Somehow the Couchbase Lite database has gotten into a state where it has a much later revision of a document (rev: `1800-xxx`) than the latest revision of that same document as it exists on sync gateway (rev: `1500-xxx`)
* Even though rev `1500-xxx` has been sync'd down from SG -> CBL, CBL considers rev `1800-xxx` as the winning revision and displays rev `1800-xxx` in the UI, rather than the desired behavior of showing rev `1500-xxx`

What needs to happen:

* The document in the Couchbase Lite DB needs to be updated so that `rev-1500` is considered the winning revision for that doc.

## Proposed Fix 

An objective-C code sample that exposes a method:

```
resetRevisions(NSDictionary docIDRevIDPairs)
```

Which would:

* Loop through the `docIDRevIDPairs`, for each pair:
  * For the current docID, for each revision in the document:
    * If the revision generation (eg, 1800) is greater than or equal than the RevID given in `docIDRevIDPairs`, but not identical to the RevID, then delete this revision.

## Additional requirement: make operation one-time-only 

It would be useful to provide a way for this code snippet to be run one time only for the given docIDRevIDPairs, possibly using a LocalDocument to track this.

pasin:
@tleyden Should the revision be removed only when it's in conflict with the given revision?

tleyden:
@pasin I don't think it needs to check whether it's in conflict or not, as long as it has a higher rev id generation than the given rev id generation, we want to remove it.

pasin:
I have implemented a sample code per proposal above. I created a PR for code review (no merge required). 

https://github.com/couchbase/couchbase-lite-ios/pull/990/files

PR #990, opened 2015-12-04
Add resetRevisions sample code in Database_Tests.m

pasin:
Implemented -(BOOL)resetRevisions:inDatabase:error per proposal in #988.

This PR is created only for code review. No merge is required.

NOTE: Per discussion with @tleyden, the app has only pull replication so creating tomstone revisions should be enough (no purge).


pasin:
After looking at the actual sample revisions data, I think this PR is not valid. I'm taking this back.

PR #992, opened 2015-12-05
Rework the solution

pasin:
Based on the revision sample data, creating a tombstone rev on each branch that doesn't contain a selected winning revision.


pasin:
@tleyden Is it correct that we don't need this anymore?

tleyden:
Yes, you can close the PR without merging if you want

PR #1002, opened 2015-12-12
Check for currentRevision.isDeletion state in CBLDocument -isDeleted

mz2:
I noticed a curious case of a deleted document responding to `-isDeleted` with `NO`. Digging deeper into its properties I can see that the current revision indeed has `_deleted = 1`:

```
(lldb) po [document isDeleted]
NO

(lldb) po [document properties]
{
    "_deleted" = 1;
    "_id" = "MPBibliographyItem:10D49A6A-4822-400C-97A2-3A41429B118A";
    "_rev" = "2-29edc04926e670a8e3df99e5b8122175";
}
```

The current definition of `isDeleted` in CBLDocument indeed does not check for the presence of `_deleted`:

```
- (BOOL) isDeleted {
    return self.currentRevision == nil && [self getLeafRevisions: NULL].count > 0;
}
```

This PR is an attempt at addressing the above issue by checking for self.currentRevision.isDeletion as well as the pre-existing conditions. 

This case happens at least when fetching a document with a `CBLQuery` with `kCBLIncludeDeleted` passed in as the `allDocsMode`.

### DIFF
--- a/Source/API/CBLDocument.m
+++ b/Source/API/CBLDocument.m
@@ -96,7 +96,8 @@ NSString* const kCBLDocumentChangeNotification = @"CBLDocumentChange";
 
 
 - (BOOL) isDeleted {
-    return self.currentRevision.isDeletion || (self.currentRevision == nil && [self getLeafRevisions: NULL].count > 0);
+    CBLRevision *currentRev = self.currentRevision;
+    return currentRev.isDeletion || ((currentRev == nil) && ([self getLeafRevisions: NULL].count > 0));
 }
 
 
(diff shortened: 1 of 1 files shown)


mz2:
Ah, reading that again, "fetching by its ID" is not really explaining the context well. The "identifier" has no relevance, it's setting `allDocsMode` to `kCBLIncludeDeleted` on a CBLQuery which lead me to this scenario where I receive a CBLDocument which is deleted but returns `NO` when it's sent `-isDeleted`. 

I shall edit the description to clarify.

snej:
In general a document that's been deleted will have a nil currentRevision, but you're right that when being enumerated with `kCBLIncludeDeleted` the tombstone revision will be visible.

The logic of the fix looks good, but please
- [ ] put `self.currentRevision` in a local variable instead of calling it twice
- [ ] Use parentheses to make explicit the ordering of `&&` and `||`.

Thanks!

mz2:
Ok – done.

PR #1060, opened 2016-01-25
Added CBLIS sample app

cflorion:
Added Sample app with Core Data model using CBLIncrementalStore.
The app shows a list of Teams with the count of players for each team.

The Add button inserts a Team and random number of players.


snej:
Having a sample app is great! Thanks for doing this. We'd rather have it as a separate repository, though, like our other sample apps. Could you create a Github repo for it, send us the URL, and then we can fork it into `couchbase` or `couchbaselabs`?

(The approach we use in our sample apps is to put a copy of `CouchbaseLite.framework` into a top-level `Frameworks` directory, and have the project link with that. But don't check the framework in; instead add a gitignore rule to ignore it, and put a README.md in the Frameworks dir telling the user to copy the framework there.)

snej:
I just read #1061 and it sounds like this app is more for isolating a bug, not for sample code. If that's true, ignore what I said above...

zgramana:
We got it. Thanks again for the repro.

PR #1066, opened 2016-01-27
Add CUSTOM_BUILD_OPTIONS to BuildFatLibrary script

pasin:
Add CUSTOM_BUILD_OPTIONS to BuildFatLibrary script to allow to set any extra build options from the command line.

For example, we need to set CLANG_ENABLE_MODULES=NO from the Jenkins build server to workaround XCode7 warning message issue.

### DIFF
--- a/Source/BuildFatLibrary.sh
+++ b/Source/BuildFatLibrary.sh
@@ -36,6 +36,10 @@ echo "CONFIGURATION_BUILD_DIR = $CONFIGURATION_BUILD_DIR"
 echo "BUILT_PRODUCTS_DIR = $BUILT_PRODUCTS_DIR"
 echo "CONFIGURATION_TEMP_DIR = $CONFIGURATION_TEMP_DIR"
 echo "TARGET_BUILD_DIR = $TARGET_BUILD_DIR"
+
+# Set CUSTOM_BUILD_OPTION env variable to include extra build options
+# Ex: $export CUSTOM_BUILD_OPTIONS="CLANG_ENABLE_MODULES=NO"
+echo "CUSTOM_BUILD_OPTIONS = $CUSTOM_BUILD_OPTIONS"
 fi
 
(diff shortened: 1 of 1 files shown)


snej:
Why do we need to set CLANG_ENABLE_MODULES in the build script? Why can't we just set the value in the target build settings in the project?

pasin:
Let me try that and see if settings only at the top target (e.g. 'CBL iOS') works.

pasin:
Actually it turns out we need both CLANG_ENABLE_MODULES=NO and CLANG_ENABLE_MODULE_DEBUGGING=NO. Missing one of them doesn't solve the warning message issue.

pasin:
To disable at the project settings, we need to turn off CLANG_ENABLE_MODULES and CLANG_ENABLE_MODULE_DEBUGGING in all static library targets including:
- CBL iOS library
- CBL iOS fat library
- CBL SQLite Storage
- BLIP-IOS in vendor/BLIP
- PocketSocket in vendor/BLIP/vendor/PocketSocket

I haven't checked the following target but most likely we need to do the same thing:
- CBL ForestDB Storage
- CBForest static in vendor/CBForest
- Listener iOS Library
- CBLJSViewCompiler

@snej let me know if I should go ahead and make changes to the target settings above instead of doing it in script.

snej:
Yeah, I know it's more work to do it that way, but I don't want a situation where the build script produces functionally-different binaries than Xcode. I can see all sorts of confusion resulting from that, since some developers use our binaries while some build their own, and we sometimes send developers/customers one-off builds from our desktops.

The easiest way to make the change would be to disable CLANG_ENABLE_MODULES and CLANG_ENABLE_MODULE_DEBUGGING at the project level, and then individually re-enable them in the couple of targets that build dynamic libraries/frameworks.

pasin:
I will close this PR then. Thanks for your comment.

PR #1109, opened 2016-02-08
Set CBLDatabaseChange.source for changes made by peer via listener

### LINKED ISSUE #1100: External property on kCBLDatabaseChangeNotification when CBL is remote
Steps to reproduce:

1. Device A starts the Listener to enable CBL to become a remote in a replication
2. Start a push replication from another db to the CBL Listener (the target)
3. The changes being pushed and posted in the `kCBLDatabaseChangeNotification` handler always have a `source` URL that's null and `external` is false

When the replication is CBL <-> Sync Gateway, the `external` field has the expected value but when CBL becomes the remote it's always false on the database event.

snej:
Sets the `source` property of the CLBChangeListener when a revision gets added by a remote peer using the listener (most likely pushing to me.)

The old code only set the URL if the connection had an authenticated username, and the URL was just `user:username`. I've changed it into an HTTP URL that includes the IP address of the connection.

Fixes #1100

### DIFF
--- a/Listener/CBLHTTPConnection.m
+++ b/Listener/CBLHTTPConnection.m
@@ -32,6 +32,8 @@
 @implementation CBLHTTPConnection
 {
     BOOL _hasClientCert;
+    NSURL* _remoteURL;
+    BOOL _builtRemoteURL;
 }
 
 @synthesize username=_username;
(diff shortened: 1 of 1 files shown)


Listener/CBLHTTPConnection.m
```diff
@@ -110,6 +112,26 @@ - (NSArray *)sslIdentityAndCertificates {
 }
 
 
+- (NSURL*) remoteURL {
+    if (!_builtRemoteURL) {
+        NSString* addr = asyncSocket.connectedHost;
+        if (![addr isEqualToString: @"127.0.0.1"]  && ![addr isEqualToString: @"::1"]) {
```
pasin: I'm curious why we don't need to set source for localhost (for phonegap?)? If I run the listener on the simulator and request to the listener simulator by using real IP or hostname, I will get source but if I use 127.0.0.1, I will not get source.

snej: `source` is supposed to be set when a revision comes from an external source, generally a replication. If it's coming from localhost, it's presumably the same app, via PhoneGap or something like that.

Issue #1085, opened 2016-02-01
CBLIS : NSIncrementalStore accessed by multiple context dispatch queues


pasin:
Reported by @cflorion in #951, NSFetchResultController called to newValuesForObjectWithID:withContext:error: from the main context's dispatch queue instead of the root's context dispatch queue.

```
thread #1: tid = 0x20750c, 0x00000001818d44c4 libobjc.A.dylib`protocols() + 4, queue = 'com.apple.main-thread'
    frame #0: 0x00000001818d44c4 libobjc.A.dylib`protocols() + 4
    frame #1: 0x00000001818d6670 libobjc.A.dylib`getProtocol(char const*) + 20
    frame #2: 0x00000001818d8ab8 libobjc.A.dylib`class_conformsToProtocol + 148
    frame #3: 0x00000001818ea2b8 libobjc.A.dylib`-[NSObject conformsToProtocol:] + 52
    frame #4: 0x00000001821933e0 CoreFoundation`-[NSCache setObject:forKey:cost:] + 124
    frame #5: 0x00000001002696f0 CBLIS iOS Test App`-[CBL_SQLiteStorage getDocNumericID:](self=0x000000012764b7b0, _cmd="getDocNumericID:", docID=@"B8A3A665-9B2A-4316-9004-0BCEBE253436-3007-00000267C85ACA96") + 384 at CBL_SQLiteStorage.m:753
    frame #6: 0x00000001002687fc CBLIS iOS Test App`-[CBL_SQLiteStorage getDocumentWithID:revisionID:withBody:status:](self=0x000000012764b7b0, _cmd="getDocumentWithID:revisionID:withBody:status:", docID=@"B8A3A665-9B2A-4316-9004-0BCEBE253436-3007-00000267C85ACA96", revID=0x0000000000000000, withBody=YES, outStatus=0x000000016fdd00dc) + 124 at CBL_SQLiteStorage.m:628
    frame #7: 0x0000000100212f84 CBLIS iOS Test App`-[CBLDatabase(self=0x000000012764a420, _cmd="getDocumentWithID:revisionID:withBody:status:", docID=@"B8A3A665-9B2A-4316-9004-0BCEBE253436-3007-00000267C85ACA96", inRevID=0x0000000000000000, withBody=YES, outStatus=0x000000016fdd00dc) getDocumentWithID:revisionID:withBody:status:] + 168 at CBLDatabase+Internal.m:508
[...]

Issue #951, opened 2015-10-18
"FMDatabase is currently in use" error saving, intermittently


ColossalChris:
99% of the time the following code works perfectly. But the other 1% I get a built in crash with assert using the following code:
```
CBLDocument *doc = [_database createDocument];
currentDocID = doc.documentID;
CBLRevision *newRevision = [doc putProperties: myDictionary error: &error];
if (!newRevision) {
                //handle error
                return;
}else{
                //add the image as an attachment
                CBLDocument *docWithAttachment = [_database documentWithID:doc.documentID];
                CBLUnsavedRevision *newRevisionWithAttachment = [docWithAttachment.currentRevision createRevision];
                NSData *imageData = UIImageJPEGRepresentation(imageToTurnIntoAttachment, 0.75);
                [newRevisionWithAttachment setAttachmentNamed: [NSString stringWithFormat:@"p_%@_%lli.jpg",[userDic objectForKey:@"_id"],timeStamp]
                           withContentType: @"image/jpeg"
                                   content: imageData];
                NSError* error;
                assert([newRev save: &error]); //random fail trace shows this line as the cause of the crash of course
 }
```

What can I do to circumvent this failure? Thanks in advance.

snej:
It's impossible to say what the problem is without knowing the error.

I recommend replacing the `assert` call with actual error handling. Assertions are only for conditions that should be impossible, i.e. the result of programmer error. It's perfectly possible for a save to fail for legitimate reasons. (For example the device's storage might be full.)

snej:
Also, not related to the problem, but: the above code is saving the document twice, when it only needs to save it once. You can set the document's properties and attachments in one operation.

ColossalChris:
The error I have been getting is 'FMDatabase is currently in use' error even though the code is run in serial fashion. Any ideas - and yes I will combine the dual saves in to a single save

snej:
What version of CBL is this?

If you're using a dispatch queue, did you set the `CBLManager.dispatchQueue` property?

ColossalChris:
Yes the manager is set up to use that dispatchQueue. Using 1.1

snej:
Any chance you could create a small standalone project that reproduces this bug? It shouldn't be possible to get database-busy errors...

Is your code creating or updating docs at a high rate?

ColossalChris:
[...]

PR #1112, opened 2016-02-09
Don't access database from multiple contexts

mastohhh:
lock NSIncrementalStore methods for values and relationships with a @synchronized block

refers to #1085 and #951


pasin:
Per my https://github.com/couchbase/couchbase-lite-ios/issues/1085#issuecomment-189520459, we need to fix the issue in a different way. I will close the PR.

PR #1113, opened 2016-02-09
Be more careful calling blocks

### LINKED ISSUE #1015: Crash when [CBLRemoteRequest connection:didFailWithError:]
Crash comes from Fabric on production version. Trying to reproduce it, but have no idea when this problem happens.

More details about crash:
https://gist.github.com/Magnat12/0260a0c66e0ff4198adb

snej:
If a block ref is stored in an ivar, and the ivar might change due to
a re-entrant call during a call to the block, then be safe and copy the
block ref to a local variable before calling it. Otherwise the block
might be dealloced while its code is running, which will cause garbage
results or even crashes when it accesses a captured variable.

Fixes #1015 (I think)

### DIFF
--- a/Source/CBLRemoteRequest.m
+++ b/Source/CBLRemoteRequest.m
@@ -176,7 +176,9 @@ typedef enum {
 
 - (void) respondWithResult: (id)result error: (NSError*)error {
     Assert(result || error);
-    _onCompletion(result, error);
+
+    CBLRemoteRequestCompletionBlock onCompletion = _onCompletion;   // keep block alive till return
+    onCompletion(result, error);
     _onCompletion = nil;  // break cycles
 }
 
--- a/Source/CBLBatcher.m
+++ b/Source/CBLBatcher.m
@@ -41,8 +41,6 @@
 }
 
 
-
-
 - (void) unschedule {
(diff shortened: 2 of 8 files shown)


Source/CBLRemoteRequest.m
```diff
@@ -176,7 +176,9 @@ - (NSMutableDictionary*) statusInfo {
 
 - (void) respondWithResult: (id)result error: (NSError*)error {
     Assert(result || error);
-    _onCompletion(result, error);
+
+    CBLRemoteRequestCompletionBlock onCompletion = _onCompletion;   // keep block alive till return
```
pasin: Should this also use `__typeof(_onCompletion) onCompletion = _onCompletion ` like the others?

pasin:
I have just one comment as above.

Source/CBLRemoteRequest.m
snej: It could, but that wouldn't change anything since its type is `CBLRemoteRequestCompletionBlock`.

(I fixed this method first, and then after searching the codebase for more I came up with the `__typeof` shortcut to make it easier to copy and paste the fix.)

PR #1122, opened 2016-02-16
Warn (but don't crash/throw) if map fn calls emit(nil

### LINKED ISSUE #1120: Warn if map function calls emit(nil, ...)
`nil` is not a valid key. If a map function passes a nil key to the emit block, it should log a warning and do nothing.

The code on the master branch will instead emit a JSON `null` key if it's a SQLite database, or crash(!) if it's ForestDB. The crash is a post-1.2 regression.

snej:
Also added a warning when user-provided keys/values can't be converted
to JSON.

Fixes #1120


snej:
FYI @borrrden @hideki

PR #1123, opened 2016-02-17
Fix major listener memory leak

### LINKED ISSUE #1017: Possible p2p memory leak
We are encountering a very large memory usage from Coushbase Lite iOS. We first noticed it when we had four or more continuous p2p connections (all pull). We have seen this in 1.1.1, 1.2(unofficial), and 1.3(unofficial).

The memory in our app will idle around 150M without the p2p. Once we begin turning the other apps on, p2p automatically begins between all of the iPads, but you will see large memory spikes as they connect (anywhere between 30M-50M per connection). Sometimes, after around three minutes, it will lower down closer to normal memory usage, other times it will not.

The issue really begins to manifest itself when you have more than four iPads. We currently have a use case that is happening in our beta right now where they can have up to 25 iPads all using p2p with each other (we are currently implementing a partially connected mesh design to reduce strain).

Replication class:
```
class Connection : NSObject
{
    var ipadOwner: FoundIpad!
    
    var push: CBLReplication!
    var pull: CBLReplication!
    
    var connectionURL: NSURL!
    
    var peerUUID: String!
    
    init(connectionURL: NSURL, owner: FoundIpad!, database: CBLDatabase! = nil, username: String = "", password: String = "", shouldCreatePush: Bool = true, UUID: String! = nil)
    {
        super.init()
        
        ipadOwner = owner
        
        self.connectionURL = connectionURL
        
        if UUID != nil
        {
            peerUUID = UUID
        }
        
        var db = database
        if db == nil
        {
            db = staticVars.database()
        }
        
        if shouldCreatePush
        {
            push = db.createPushReplication(connectionURL)
            push.continuous  = true
            
            if peerUUID != nil
            {
                push.customProperties = [String: AnyObject]()
                push.customProperties!["remoteUUID"] = peerUUID!
            }
        }
    
        pull = db.createPullReplication(connectionURL)
        pull.continuous  = true
        
        if peerUUID != nil
        {
            pull.customProperties = [String: AnyObject]()
            pull.customProperties!["remoteUUID"] = peerUUID!
        }
        
        if username != "" && password != ""
        {
            if shouldCreatePush
            {
            	push.authenticator = CBLAuthenticator.basicAuthenticatorWithName(username, password: password)
            }
            pull.authenticator = CBLAuthenticator.basicAuthenticatorWithName(username, password: password)
        }
        
        print("Push and pull created for url: \(connectionURL)")
        
        if shouldCreatePush
        {
        	NSNotificationCenter.defaultCenter().addObserver(self,
            	selector: "pushChanged", name: kCBLReplicationChangeNotification, object: push)
        }
        
        NSNotificationCenter.defaultCenter().addObserver(self,
            selector: "pullChanged", name: kCBLReplicationChangeNotification, object: pull)
        
        if shouldCreatePush
        {
        	push.start()
        }
        pull.start()
    }
    
    func stopConnection()
    {
        if push != nil
        {
            push.stop()
        }
        if pull != nil
        {
            pull.stop()
        }
        print("Push and pull closed for url: \(connectionURL)")
    }
    
    func pushChanged()
    {
        let pushStatus = CBLReplicationStatus(rawValue: self.push.status.rawValue)!
        
        if staticVars.appDelegate != nil
        {
            staticVars.appDelegate.notifyReplicationChange()
        }
        print("Replication ➡️ Push Change: replication state: \(pushStatus.description()) | ip: \(self.push.remoteURL) | error: \(self.push.lastError))")
        
    }
    
    func pullChanged()
    {
        let pullStatus = CBLReplicationStatus(rawValue: self.pull.status.rawValue)!
        
        if staticVars.appDelegate != nil
        {
            staticVars.appDelegate.notifyReplicationChange()
        }
        print("Replication ⬅️ Pull Change: replication state: \(pullStatus.description()) | ip: \(self.pull.remoteURL) | error: \(self.pull.lastError))")
        
    }
    
    func restart()
    {
        if push != nil
        {
            push.restart()
        }
        if pull != nil
        {
            pull.restart()
        }
    }
    
    enum CBLReplicationStatus: UInt32
    {
        case kCBLReplicationStopped
        case kCBLReplicationOffline
        case kCBLReplicationIdle
        case kCBLReplicationActive
        
        func description () -> String
        {
            switch self
            {
            case kCBLReplicationStopped:
                return "Stopped"
            case kCBLReplicationOffline:
                return "Offline"
            case kCBLReplicationIdle:
                return "Idle"
            case kCBLReplicationActive:
                return "Active"
            }
        }
    } 
}
```

We create one of these for each iPad that is found and should be replicated to.

screen shot 2016-01-06 at 10 20 31 am

We have tried a one shot replication model where we would listen for changes based on other iPad's listeners. This fixed the memory issue, but caused a delay in the propagation of documents, which allows for errors in our app (we require changes to be reflected nearly immediately and continuous does that for us).

We have also implemented a UUID model, inspired by the p2p Grocery Sync example, that stores a unique ID of iPads to prevent unnecessary re-cataloging of other iPads when a port or IP address changes.

If any other information is needed, just let me know and I will get it to you posthaste. Any help or advice is greatly appreciated.

snej:
CBLHTTPConnection objects never got cleaned up, because my override of
-socketDidDisconnect:withError: wasn't calling the (undeclared)
superclass method, which is what removes the connection object from
the server's connection list.
This caused the CBLHTTPConnection, GCDAsyncSocket, and other objects
to stay in memory; not leaked exactly, but kept in the HTTPServer's
connection list.
Fixed by adding the 'super' call.

Fixes #1017

### DIFF
--- a/Listener/CBLHTTPConnection.m
+++ b/Listener/CBLHTTPConnection.m
@@ -29,6 +29,11 @@
 #import "Test.h"
 
 
+// For some reason this is not declared in HTTPConnection's @interface
+@interface HTTPConnection () <GCDAsyncSocketDelegate>
+@end
+
+
 @implementation CBLHTTPConnection
 {
     BOOL _hasClientCert;
@@ -139,6 +144,7 @@ static void evaluate(SecTrustRef trust, SecTrustCallback callback) {
                                  code: GCDAsyncSocketReadTimeoutError]) {
         Warn(@"CBLHTTPConnection: Client disconnected: %@", error);
     }
+    [super socketDidDisconnect: socket withError: error];
 }
 
 


snej:
After this fix, I ran LiteServ in Instruments with the Leaks instrument, and found no growth in the number of CBL objects after sending it individual REST requests or after replicating from it using another copy of LiteServ.

pasin:
@snej do you want this to merge to master or dev? The PR is currently set to `merge to master`.

snej:
(Sorry, wrong target branch — should be dev. Closing.)

PR #1133, opened 2016-02-19
Fix CBLVersion() in official builds of iOS simulator binaries

### LINKED ISSUE #1131: CBLVersion() incorrectly returns "(unofficial)" in official releases
As discovered in [this forum thread](https://forums.couchbase.com/t/is-couchbase-lite-1-2-really-out/6959), the result of calling `CBLVersion()` in the official 1.2 release is `@"1.2 (unofficial)"`. But only if you run it in the iOS simulator — on a device, or in a Mac OS app, you correctly get `@"1.2.0 (build 112)"`.

The cause of this is that the constant `CouchbaseLiteVersionNumber` is 0.0 in the simulator builds. And that's because the build setting `CURRENT_PROJECT_VERSION` has its default value of `0`. The [build script](https://github.com/couchbase/build/blob/master/scripts/jenkins/mobile/build_cblite_ios.sh#L208) overrides it to the correct build number when invoking xcodebuild, but that's not taking effect when the simulator binaries are compiled. And _that_ seems to be due to BuildFatLibrary.sh.

snej:
The build variable CURRENT_PROJECT_VERSION, which gets overridden by
the official server build script, didn't get passed through to the
simulator builds done as part of our fat (device+simulator) build script.
Updated BuildFatLibrary.sh to explicitly pass it, and two other build
variables that also get set during official builds.

Fixes #1131


snej:
This PR is targeted at the master branch, because it fixes a build issue, and the only way we can make sure it's really fixed is to look at an official build.

michaelkwok:
I will verify the master build and delete the branch afterward.

michaelkwok:
Build 589 looks good now

PR #1136, opened 2016-02-22
Reimplemented CBLReplication.pendingDocumentIDs and -isDocumentPending:

### LINKED ISSUE #1132: CBLReplicator.isDocumentPending is inaccurate when replicator is offline
It seems like`isDocumentPending` might not be working as one would expect it to when a device is completely offline.

# Setup to reproduce

* CBL (1.2 release)
* Sync Gateway (1.2 release)
* ToDoLite-iOS ([patched with minimally invasive sync status feature](https://github.com/PaulCapestany/ToDoLite-iOS/commit/5fdaf1d48904ad9a01e8a971c9d668b7bd522e4f) which displays synced text as **black**, and unsynced text as *grey*). Sidenote: relevant [PR](https://github.com/couchbaselabs/ToDoLite-iOS/pull/68).

[Follow these instructions in order to use Apple's "Network Link Conditioner"](http://nshipster.com/network-link-conditioner/) to test different connectivity qualities on both the simulator and on actual physical iOS devices.

In the ToDoLite-iOS project, change the `kSyncGatewayUrl` in `AppDelegate.m` if need be. Also, make sure that the `FacebookAppID` and `URL Scheme` in `Info.plist` is usable/accurate as well.

# Expected behavior

1. When on a good wifi connection, newly added or edited items in any of the tableViews should have **black text**.
2. When switching to the "Very Bad Network" setting in Apple's "Network Link Conditioner", newly added or edited items in any of the tableViews should have *grey text* if they haven't been successfully synced within 1.5 seconds (they should turn **black** again once the replication properly executes however).
3. When putting the testing device in 'Airplane Mode', newly added or edited items in any of the tableViews should have *grey text* since they're unable to sync, and should turn **black** again when back on a network connection.

# Actual behavior

1. Works as expected.
2. Works as expected.
3. Unexpected behavior. The text of newly added or edited items stays **black** even though items have not synced due to being in 'Airplane Mode' (and therefore should be *grey*).

snej:
Now tracking replicator's lastSequence and caching it in an ivar.
-isDocumentPending: just compares the doc's sequence with that.
pendingDocumentIDs does basically what it used to do, but it doesn't
need to do it on a background thread.

Also, the bug-fix portion of this is that pendingDocumentIDs remembers
the database's lastSequenceNumber when it's cached; if that changes,
the cached value is invalidated.

Fixes #1132

### DIFF
--- a/Source/API/CBLReplication.m
+++ b/Source/API/CBLReplication.m
@@ -55,8 +55,10 @@ NSString* CBL_ReplicatorStoppedNotification = @"CBL_ReplicatorStopped";
 
 @implementation CBLReplication
 {
-    NSSet* _pendingDocIDs;
-    bool _started;
+    bool _started;                          // Has replicator been started?
+    SequenceNumber _lastSequencePushed;     // The latest sequence pushed by the replicator
+    NSSet* _pendingDocIDs;                  // Cached set of docIDs awaiting push
+    SequenceNumber _pendingDocIDsSequence;  // DB lastSequenceNumber when _pendingDocIDs was set
 
     // ONLY used on the server thread:
     id<CBL_Replicator> _bg_replicator;
@@ -280,7 +282,7 @@ NSString* CBL_ReplicatorStoppedNotification = @"CBL_ReplicatorStopped";
 
         // Initialize the status to something other than kCBLReplicationStopped:
         [self updateStatus: kCBLReplicationOffline error: nil processed: 0 ofTotal: 0
-                serverCert: NULL];
+             lastSeqPushed: 0 serverCert: NULL];
 
         [_database addReplication: self];
     }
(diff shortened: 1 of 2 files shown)


Source/API/CBLReplication.m
```diff
@@ -403,46 +409,84 @@ - (void) updateStatus: (CBLReplicationStatus)status
...
-            Warn(@"Error getting unpushed revisions : %@", error);
+    CBL_ReplicatorSettings* settings = self.replicatorSettings;
+    if (!settings)
         return nil;
-    }];
+    SequenceNumber lastSequence = self.lastSequencePushed;
+    if (lastSequence < 0)
+        return nil;
+
+    _pendingDocIDsSequence = _database.lastSequenceNumber;
```
pasin: I think it's more safe to move it after `if (!revs)` block.

snej: You're right, the assignment should go there. But reading `_database.lastSequenceNumber` should still be before, to avoid skipping any sequences. I'll make the change and update the PR.

pasin:
@snej the PR cannot be merged. Could you please fix?

PR #1149, opened 2016-02-29
Improve multipart reading performance using a cyclic buffer

robertjpayne:
The CBLMultipartReader is somewhat slow parsing on older devices. This is almost entirely caused by the `[CBLMultipartReader deleteUpThrough:]` API, mostly because this particular NSMutableData API is very slow though even a realloc here wouldn't be that great.

This slowness is easily observable on an older device ( iPhone 4S in my case ) with instruments time profiler.

This PR addresses this by bringing in a more cyclic style buffer that appends data and maintains a offset/bytes pointer and finally compacts itself once it's past a certain amount of data.

This change makes most of the memory operations inside the `CBLMutlipartReader` negligible.

### DIFF
--- a/Source/CBLMultipartBuffer.m
+++ b/Source/CBLMultipartBuffer.m
@@ -32,7 +32,7 @@
 #pragma mark - Actions
 
 - (BOOL)advance:(NSUInteger)amount {
-    if (_offset + amount <= _data.length) {
+    if (_offset + amount >= _data.length) {
         Warn(@"Preventing unsafe buffer overflow", nil);
         return NO;
     }
--- a/Source/CBLMultipartBuffer.m
+++ b/Source/CBLMultipartBuffer.m
@@ -1,0 +1,70 @@
+//
+//  CBLMultipartBuffer.m
+//  CouchbaseLite
+//
+//  Created by Robert Payne on 1/03/16.
+//  Copyright © 2016 Couchbase, Inc. All rights reserved.
+//
+
+#import "CBLMultipartBuffer.h"
+
+@interface CBLMultipartBuffer() {
+    NSMutableData *_data;
+}
+
+@end
+@implementation CBLMultipartBuffer
+
+@synthesize offset=_offset, compactionLength=_compactionLength;
+
+#pragma mark - Initialization
+
+- (instancetype)init {
+    if((self = [super init])) {
+        _data = [NSMutableData dataWithCapacity:1024];
+        _offset = 0;
+        _compactionLength = 4096;
+    }
+    return self;
+}
+
+#pragma mark - Actions
+
+- (BOOL)hasBytesAvailable {
+    return _data.length > _offset;
+}
+- (NSUInteger)bytesAvailable {
+    if(_data.length > _offset) {
+        return _data.length - _offset;
+    }
+    return 0;
+}
+- (void)appendData:(NSData *)data {
+    [_data appendData:data];
+}
+- (void)appendBytes:(const void *)bytes length:(NSUInteger)length {
+    [_data appendBytes:bytes length:length];
+}
+- (void)compact {
+    if(_offset > _compactionLength && _offset > (_data.length >> 1)) {
+        _data = [NSMutableData dataWithBytes:(char *)_data.bytes + _offset
+                                      length:_data.length - _offset];
+        _offset = 0;
+    }
+}
+- (void)reset {
+    _offset = 0;
+    _data.length = 0;
+}
+- (const void *)bytes {
+    return _data.bytes + _offset;
+}
+- (void *)mutableBytes {
+    return _data.mutableBytes + _offset;
+}
+- (NSData *)data {
+    return [_data copy];
+}
+
+
+@end
--- a/Source/CBLMultipartBuffer.m
+++ b/Source/CBLMultipartBuffer.m
@@ -75,7 +75,7 @@
     if (!self.hasBytesAvailable) {
         return nil;
     }
-    return [[NSData dataWithBytes:self.bytes length:self.bytesAvailable freeWhenDone:NO] subdataWithRange:range];
+    return [[NSData dataWithBytesNoCopy:self.mutableBytes length:self.bytesAvailable freeWhenDone:NO] subdataWithRange:range];
 }
 
 @end
--- a/Source/CBLMultipartBuffer.m
+++ b/Source/CBLMultipartBuffer.m
@@ -8,13 +8,10 @@
 
 #import "CBLMultipartBuffer.h"
 
-@interface CBLMultipartBuffer() {
+@implementation CBLMultipartBuffer {
     NSMutableData *_data;
     NSUInteger _offset;
 }
-
-@end
-@implementation CBLMultipartBuffer
 
 @synthesize compactionLength=_compactionLength;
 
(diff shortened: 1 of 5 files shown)


robertjpayne:
P.S. I've run tests locally but I can't really figure out how to run the test rig in the project so apologies if this triggers warnings/errors on your end.

Also did my best to integrate it to the correct targets but could use a second look as the project is quite complex.

snej:
Thanks for this! I skimmed it quickly and while I haven't seen any bugs, I think the API could be cleaner/safer. (For instance, I don't think the `offset` property should be exposed; it breaks the abstraction, and it allows the client to set it to an illegal value which will probably lead to crashes heap corruption.)

I can either post issues inline and you can fix them, or I can check out the branch and edit the code before merging (of course leaving your name in the credits). Do you have a preference?

robertjpayne:
@snej I can tidy them up just let me know, fixed the buffer one ( though testing it still )

Source/CBLMultipartBuffer.m
```diff
@@ -0,0 +1,79 @@
...
+//  CouchbaseLite
+//
+//  Created by Robert Payne on 1/03/16.
+//  Copyright © 2016 Couchbase, Inc. All rights reserved.
+//
+
+#import "CBLMultipartBuffer.h"
+
+@interface CBLMultipartBuffer() {
+    NSMutableData *_data;
```
snej: These can go at the start of the `@implementation`; you don't need to add a category just to declare ivars.

Source/CBLMultipartBuffer.m
```diff
@@ -0,0 +1,79 @@
...
+        _compactionLength = 4096;
+    }
+    return self;
+}
+
+#pragma mark - Actions
+
+- (BOOL)advance:(NSUInteger)amount {
+    if (_offset + amount >= _data.length) {
+        Warn(@"Preventing unsafe buffer overflow", nil);
```
snej: The `nil` parameter isn't needed.

Source/CBLMultipartBuffer.m
```diff
@@ -0,0 +1,79 @@
...
+        return NO;
+    }
+    _offset += amount;
+    return YES;
+}
+- (BOOL)hasBytesAvailable {
+    return _data.length > _offset;
+}
+- (NSUInteger)bytesAvailable {
+    if(_data.length > _offset) {
```
snej: I don't think it's possible for `_data.length` to be less than `_offset`, so you don't need that test.

Source/CBLMultipartBuffer.m
```diff
@@ -0,0 +1,79 @@
...
+}
+- (void)appendData:(NSData *)data {
+    [_data appendData:data];
+}
+- (void)appendBytes:(const void *)bytes length:(NSUInteger)length {
+    [_data appendBytes:bytes length:length];
+}
+- (void)compact {
+    if(_offset > _compactionLength && _offset > (_data.length >> 1)) {
+        _data = [NSMutableData dataWithBytes:(char *)_data.bytes + _offset
```
snej: It would be more efficient to `memmove` the remaining bytes down to `_data.mutableBytes` and then lower `_data.length`. This avoids memory allocation, and leaves `_data` at its current capacity, so it won't have to grow immediately.

Source/CBLMultipartBuffer.m
robertjpayne: Probably not, it was a safety check before the hiding of _offset

Source/CBLMultipartBuffer.m
```diff
@@ -0,0 +1,79 @@
...
+    _data.length = 0;
+}
+- (const void *)bytes {
+    return _data.bytes + _offset;
+}
+- (void *)mutableBytes {
+    return _data.mutableBytes + _offset;
+}
+- (NSData *)data {
+    return [_data copy];
```
snej: This includes the already-consumed bytes. Seems like it should return only the unread bytes.

robertjpayne: To be honest should probably just remove the copy method all together, it's not really that useful in this context.

Source/CBLMultipartBuffer.m
robertjpayne: I don't think the optimisation is worth changing it here, it's a minute optimisation for how often the compaction runs.

The reduction of the capacity is actually intentional to ensure memory stays as compact as possible.

robertjpayne:
@snej fixed up most of the APIs and also brought in a `subdataWithRange` API to DRY up the use in the multipart reader.

snej:
BTW, before I forget, could you please [sign the contributor license agreement](https://github.com/couchbase/couchbase-lite-ios/wiki/Sign-the-contributor-license-agreement)? We need that for legal reasons before we can accept a patch from you. Thanks!

Source/CBLMultipartReader.m
```diff
@@ -137,17 +138,20 @@ - (BOOL) parseHeaders: (NSString*)headersStr {
 
 
 - (NSRange) searchFor: (NSData*)pattern from: (NSUInteger)start {
-    return [_buffer rangeOfData: pattern
+    if (!_buffer.hasBytesAvailable)
+        return NSMakeRange(NSNotFound, 0);
+    NSData *data = [NSData dataWithBytesNoCopy:_buffer.mutableBytes length:_buffer.bytesAvailable freeWhenDone:NO];
```
snej: It'd be better to implement -rangeOfData:... directly on CBLMultipartBuffer. Then it can search without creating a new NSData object since it knows the range to search in.
(This does probably make a difference, since this method is called on every read to find the multipart boundary string.)

snej:
Have you run with the Address Sanitizer enabled? That tool is really good at detecting buffer overruns.

Source/CBLMultipartReader.m
robertjpayne: Yea it could be implemented directly there. To be honest I copied my cyclic buffer from `PocketSocket` and just modified it a bit to suite your comments. I don't think there will be a huge performance difference here as it doesn't copy the bytes.

I was trying to not wholesale copy all the NSData API's over but rather just make a more workable `void *`.

robertjpayne:
@snej totally happy to sign a CLA buuut do you have any way to do it without a Yahoo account?

snej:
The site supports any OpenID provider, so if you have a StackOverflow/StackExchange account, you can log in by entering the URL `https://openid.stackexchange.com`. WordPress.com accounts work too, apparently; just enter the URL of your subdomain.

robertjpayne:
@snej it really doesn't work, I just tried and after logging in with Stack Exchange it just says "Not Found"

robertjpayne:
Ok so if you login first then go through OpenID it gets a bit further but fails once redirecting back to the Gerrit page:

`http://review.couchbase.org/#SignInFailure,SIGN_IN,Discovered+information+verification+failed.`

```
Not Found
 	 
The page you requested was not found, or you do not have permission to view this page.
```

snej:
Sorry about that! (The site admins keep saying they're about to get Github logins working, but it hasn't happened yet.) I will ask around and get you some help.

snej:
Sorry to leave you hanging, Robert — I've just escalated this and am asking for help with the login. We shouldn't be stuck just because you don't have a Yahoo account. Hopefully we can get this cleared up soon.

snej:
At last! I have a PDF of the new contributor agreement. Please review, sign, and either email us a scan or send the paper by snail-mail (addresses are in the document.)

[Couchbase New Contributor Agreement.pdf](https://github.com/couchbase/couchbase-lite-ios/files/215636/Couchbase.New.Contributor.Agreement.pdf)

snej:
I ran the unit tests (and found one bug), cleaned up the style a bit, and optimized -compact further. It's merged to dev now. Thanks!

PR #1155, opened 2016-03-03
Moved customHTTPHeaders initialization to the common initializer

lucatorella:
`customHTTPHeaders` was initialized in the `init` method but not by the public method `initWithDirectory:options:error:`. So it is legally possible to create an instance of `CBLManager` with a `nil` `customHTTPHeaders`.
To fix this, I moved `customHTTPHeaders` initialization to the custom initialization method.


snej:
Thanks!

PR #1166, opened 2016-03-13
SQLite: Prune deepest revisions when new ones are inserted

### LINKED ISSUE #1165: Docs in SQLite dbs never get pruned until entire db is compacted
The SQLite storage engine only prunes documents — deletes metadata of old revisions that are deeper down than the maxRevTreeDepth — when the database's Compact method is called by the app. This means that a doc that's frequently updated will accumulate more and more entries in the `revs` table. 

We have evidence from the field that this eventually slows down database operations on that document to an unacceptable degree (albeit in an extreme scenario where a doc is updated several times a minute for hours, while multiple replications are running.)

We can fix this by pruning the revision tree periodically (the way the ForestDB storage does) as new revisions are added. `-[CBLSQLiteStorage forceInsert:...]` (called by pull replications) looks like it could be extended to prune old revisions, since it already loads the metadata of every revision of the document. The `-addDocID:...` method (called by local puts) is a bit harder to fix since it doesn’t have the whole document history available. It might be sufficient to check if the new revision’s generation # is a multiple of the maxRevTreeDepth, and if so trigger a prune operation on that document.

snej:
This makes revision pruning incremental, the way it is in ForestDB. A document's revision tree will never get deeper than the database's maxRevTreeDepth.

The -compact method used to prune every doc in the database. This is still needed, _once_, for pre-existing databases that might have a lot of docs that need pruning. I've added a special key to the database's 'info' metadata to indicate that this bulk prune isn't needed; it's set after the prune, or when a new database is created.

Fixes #1165

CC: @borrrden , @hideki

### DIFF
--- a/Source/CBL_SQLiteStorage.m
+++ b/Source/CBL_SQLiteStorage.m
@@ -461,6 +461,9 @@ static void CBLComputeFTSRank(sqlite3_context *pCtx, int nVal, sqlite3_value **a
     if (isNew && ![self initialize: @"END TRANSACTION" error: outError])
         return NO;
 
+    if (isNew)
+        [self setInfo: @"true" forKey: @"pruned"];  // See -compact: for explanation
+
     if (!isNew)
         [self optimizeSQLIndexes];          // runs ANALYZE query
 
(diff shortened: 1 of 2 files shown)


pasin:
Sorry for keep this review pending. I will review all PRs by end of this week. Thanks!

PR #1184, opened 2016-03-25
Improved CBLReachability (proxy handling, intermittent connections)

### LINKED ISSUE #1005: Replication fails without starting on wifi if network depends on a proxy
If on a wifi connection within a network that uses a proxy, replication never really starts because it's using 
SCNetworkReachabilityCreateWithName under the covers to verify the host is reachable before attempting replication. Since SCNetworkReachabilityCreateWithName does not respect http proxy settings (see: http://stackoverflow.com/questions/7286329/does-scnetworkreachability-respect-http-proxy-settings for a discussion of this issue) this means replication within a network that uses a proxy never works.
If however, instead of SCNetworkReachabilityCreateWithName, the code used SCNetworkReachabilityCreateWithAddress with zero address (like Apple's Reachability reachabilityForInternetConnection sample code) the existence of an internet connection could be verified without the specific host check. If then the network connection failed you could examine the failure reason to determine it was because the host could not be reached.

Log snippet:
16:15:26.960‖ Sync: CBL_Puller[https://917xxx.run.aws-usw02-pr.myserver.com/pg/data/pm]: Reachability state = <917530.run.aws-usw02-pr.ice.predix.io>:unreachable (00), suspended=0
16:15:26.960‖ Sync: CBL_Puller[https://917xxx.run.aws-usw02-pr.myserver.com/pg/data/pm]: Going offline
16:15:26.960‖ Sync: Stopping 1 remote requests
16:15:26.961‖ Sync: CBL_Puller[https://917xxx.run.aws-usw02-pr.myserver.com/pg/data/pm]: Error fetching last sequence: The operation couldn’t be completed. (NSURLErrorDomain error -999.)

snej:
* Now takes a URL, not just a hostname.
* Supports checking for general network connectivity
* Now checks system proxy settings to see if a proxy is required to
  reach the given URL; if so, it switches to checking general
  network connectivity.
* Host is considered reachable even if the flags say 'ConnectionRequired'
  as long as the connection doesn't require user intervention.

Fixes #1005 (I think; don't have a proxied network to test with)

### DIFF
--- a/Source/CBLBlipReplicator.m
+++ b/Source/CBLBlipReplicator.m
@@ -210,7 +210,7 @@
 
     if (!_reachability && CBLIsOfflineError(_error)) {
         // Network seems to be offline, so start a reachability monitor so I can retry ASAP:
-        _reachability = [[CBLReachability alloc] initWithHostName: _settings.remote.host];
+        _reachability = [[CBLReachability alloc] initWithURL: _settings.remote];
         __weak CBLBlipReplicator* weakSelf = self;
         _reachability.onChange = ^{
             [weakSelf reachabilityChanged];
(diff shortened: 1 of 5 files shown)


pasin:
I have created #1213 for testing the replicator with a proxy.

PR #1187, opened 2016-03-30
Fixed incompatibility with CouchDB's _changes feed

### LINKED ISSUE #1139: ChangeTracker POST request are incompatible with CouchDB
* Version:   1.2 Release Branch
* Client OS: iOS 9.2
* Server:    CouchDB 1.6

The ChangeTracker contains a `_usePOST` ivar which by default is set to `YES`. When this is set to `YES` the changes URL is built as simply `_changes` and all of the request parameters are pushed into a JSON body for the post request.

This is however invalid as CouchDB still expects the majority of parameters to come through the URL query params not the post body.

You can reproduce this by targeting any CouchDB database and running:

```
POST /_changes
Content-Type: application/json

{"limit": 5}
```

You will notice this does not actually limit the result set at all. Where as if you run:

```
POST /_changes?limit=5
```

It will correctly capture the parameters. As far as I can tell the only valid parameter to have in the POST body is doc_ids and all others should remain in the query params of the URL

snej:
CouchDB also supports POST to _changes, but expects most of the parameters
to still be in the URL, not the request body.

Fixes #1139

### DIFF
--- a/Source/ChangeTracker/CBLChangeTracker.m
+++ b/Source/ChangeTracker/CBLChangeTracker.m
@@ -96,9 +96,10 @@ DefineLogDomain(ChangeTracker);
 }
 
 - (NSString*) changesFeedPath {
-    if (_usePOST)
-        return @"_changes";
-    
+    // We add the basic query params to the URL even if we'll send a POST request. Yes, this is
+    // redundant, since those params are in the JSON body too. This is for CouchDB compatibility:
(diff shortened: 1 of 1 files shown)


danthi:
Does this fix also apply to Android?  We seem to be having this same issue when replicating CBL with CouchDB 2.0.

snej:
@hideki?

PR #1193, opened 2016-04-01
Speculative fixes for assertion failures caused by invalid JSON

### LINKED ISSUE #1179: Assertion failure in CBL_MutableRevision, pushing document
Hi all, 

Reposting from forum.

This is probably a long shot, but wanted to see if anyone here experienced the same issue.  Whenever we push out a new update on our app in the app store, some user is reporting a crash on boot and there is no way from them to come back to the app unless they do a full reinstall.

I managed to ask one of them to send us their crash log file in their own device, and [here](https://gist.github.com/hermanccw/d119fc1067a1b4b77cf5) it is.

Which doesn't mean anything since it is not symbolized, however, I am getting the similar crash log from Crashlytics that looks like [this](https://gist.github.com/hermanccw/ca17a10b43e3449f013b)

Which seems to suggest it is failing on the CouchbaseLite thread and it's something regarding an assertFail on "!_changeTracker"

Looking at the log message before the crash in Crashlytics, it seems to indicate that the app is going in and out of network connection before the crash.  Could that be related?

Also want to note that we are using CBL with multiple dispatch queues, I've done a clean up in my code to make sure no cross-thread access to CBL* objects, but not sure if this is the cause of such crash.

It's especially bad since users who's getting these crashes seems to be stuck on a crash on boot cycle and never get into the app.

--------
* Version:   1.1.0 at commit 8880a96
* Client OS: iOS 8/9
* Server:   CouchDB 1.3.1

snej:
In diagnosing @hermanccw's crash (#1179) I read through some code and found a few ways that invalid JSON data could trigger assertion failures. I still don't know exactly how the bad data could have happened, but these are worth fixing and should prevent it in the future.

Fixes #1179

### DIFF
--- a/Source/CBLRestPusher.m
+++ b/Source/CBLRestPusher.m
@@ -282,10 +282,14 @@
                 CBLStatus status;
                 CBL_Revision* loadedRev = [db revisionByLoadingBody: rev
                                                              status: &status];
+                if (loadedRev && !loadedRev.properties)
+                    status = kCBLStatusBadJSON;
                 if (status >= 300) {
-                    Warn(@"%@: Couldn't get local contents of %@", self, rev);
-                    [self revisionFailed];
-                    continue;
+                    Warn(@"%@: Couldn't get local contents of %@ (status=%d)", self, rev, status);
+                    if (status < 500)
+                        [self removePending: rev];
+                    else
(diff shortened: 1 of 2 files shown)


pasin:
Could you fix the PR?

snej:
Fixed!

PR #1204, opened 2016-04-05
Added missing CBLDynamicObject.h header to CBL iOS target's Public headers…

brendand:
Without this, the built CouchbaseLite.framework would display a CBLDynamicObject.h not found error when trying to build your project.


snej:
Thanks!

PR #1208, opened 2016-04-06
Added 'attachments:' parameter to -[CBLDocument putExistingRevision...]

### LINKED ISSUE #1195: Need API to add attachments with -putExistingRevisionWithProperties:... method
The new-for-1.3 method `-[CBLDocument putExistingRevisionWithProperties: revisionHistory: fromURL: error]` doesn't have a good way to add attachments. Currently they have to be inline, i.e. with every attachment having a `body` property whose value is a base64 string.

A couple of ideas:

* Allow the `body` property value to be NSData or CBLAttachment
* Add an `attachments:` parameter that's an NSDictionary mapping attachment names to NSData or CBLAttachment. This would be nice in that you wouldn't have to mess with the revision body to splice the bodies into it.

snej:
Allows revisions with attachments to be added.

Fixes #1195

### DIFF
--- a/Source/API/CBLDocument.h
+++ b/Source/API/CBLDocument.h
@@ -116,6 +116,9 @@ NS_ASSUME_NONNULL_BEGIN
     implementing some new kind of replicator that transfers revisions from another database.
     @param properties  The properties of the revision (_id and _rev will be ignored, but _deleted
                     and _attachments are recognized.)
+    @param attachments  A dictionary providing attachment bodies. The keys are the attachment
+                    names (matching the keys in the properties' `_attachments` dictionary) and
+                    the values are the attachment bodies as NSData or NSURL.
     @param revIDs  The revision history in the form of an array of revision-ID strings, in
                     reverse chronological order. The first item must be the new revision's ID.
                     Following items are its parent's ID, etc.
@@ -124,7 +127,8 @@ NS_ASSUME_NONNULL_BEGIN
                     decide whether the change is local or not.)
     @param outError  Error information will be stored here if the insertion fails.
     @return  YES on success, NO on failure. */
-- (BOOL) putExistingRevisionWithProperties: (NSDictionary*)properties
+- (BOOL) putExistingRevisionWithProperties: (CBLJSONDict*)properties
+                               attachments: (nullable NSDictionary*)attachments
                            revisionHistory: (CBLArrayOf(NSString*)*)revIDs
                                    fromURL: (nullable NSURL*)sourceURL
                                      error: (NSError**)outError;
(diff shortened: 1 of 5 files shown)


snej:
@brendand -- since you're using this API, could you try out this commit and see if it works for you?

brendand:
Looks great from browsing the code. I'll give it a try now.

brendand:
So I'm trying out the new code with a URL to a local file, but I'm getting errors.

Here's a gist of my code:

[https://gist.github.com/brendand/aea42c302da1456e5b09fab5a3260678](url)

and sample JSON data:

[https://gist.github.com/brendand/f229551cafe57f2733713cdf730f6c20](url)

The errors I'm getting are:

```
WARNING: Don't have original attachment for stub 'icon' in {frm-1146F8F3797F4E7698167EF1980DB153261080000094C59BE9234 #30-8098f81efdc5cd4465cc5786dfd8f613} (missing rev) {at __74-[CBLDatabase(Attachments) processAttachmentsForRevision:ancestry:status:]_block_invoke:426}

Error Domain=CBLHTTP Code=400 "400 Invalid attachment" UserInfo={NSLocalizedFailureReason=Invalid attachment, NSLocalizedDescription=400 Invalid attachment}
```

My attachmentProps dictionary looks like this:

```
{
    icon = "file:///Users/brendan/Desktop/My%20Movie%20Library.tfarc/attachments/FF0C630C0552A4E69FEEA37B496BAA359E0833BF.blob";
}
```

If you need more information, please let me know.

Thanks!

snej:
You'll need to provide data for "967B4D6D-E778-492C-B69E-1B21EB37B5BF.jpg" and "Shawshank.jpg" in your `attachments` dictionary. (There's no attachment named `icon` -- did you mix that up with a different document?)

But I can also see that the code needs to process doc attachments with a `"stub":true` property, if the data is given in the `attachments` dictionary. The `stub` means that when the JSON was generated Couchbase Lite didn't think it needed to include the attachments, but in this case it does because you're creating the document from scratch.

snej:
OK, I've updated the commit. Now any entry in the `attachments` dict will force the corresponding `_attachments` entry into `"follows":true` mode, so the right thing will happen.

brendand:
Yup. That fixed it!

Works perfect now!

Sorry for the mismatch in my sample attachmentProps dictionary. Yes, that was from a different document than the sample JSON I provided. I had edited the comment a bunch of times and missed updating that.

My code works perfect now with your latest update. I'd say it's ready to commit to master :+1:

snej:
Thanks!

PR #1211, opened 2016-04-11
Allow incoming attachments with revpos:0, for PouchDB compatibility

### LINKED ISSUE #1200: Replication rejects attachments generated by PouchDB with revpos:0
```
16:31:20.223‖ WARNING: CBLRestPuller[https://server:443/user-62356431623932312d323233652d343935612d623834622d663439666462373234356336]: Revision {section_image-7d5198c2-5ddf-5405-9835-e980b634535c #2-8145a6217ced16862b43577c204d2ac8} has invalid attachment metadata: {
    "image.png" =     {
        "content_type" = "image/png";
        digest = "md5-Hke94Wxx368Us7xYprTmIw==";
        follows = 1;
        length = 101033;
        revpos = 0;
    };
}
```

The issue seems to be revpos not having a proper value, value is 0 and should be 2 in this case.
But anyway i would expect this doc to be synced.

--------
* Version:  1.2.0
* Client OS: iOS 9.3
* Server:  Apache CouchDB 1.5.0

snej:
Fixes #1200

### DIFF
--- a/Source/CBL_Attachment.m
+++ b/Source/CBL_Attachment.m
@@ -101,8 +101,13 @@ static NSString* blobKeyToDigest(CBLBlobKey key) {
             // Get the revpos:
             id revPosObj = attachInfo[@"revpos"];
             if (revPosObj) {
-                int revPos = [$castIf(NSNumber, revPosObj) intValue];
-                if (revPos <= 0) {
+                if (![revPosObj isKindOfClass: [NSNumber class]]) {
+                    *outStatus = kCBLStatusBadAttachment;
+                    return nil;
+                }
+                int revPos = [revPosObj intValue];
+                // PouchDB has a bug that generates "revpos":0; allow this (#1200)
+                if (revPos < 0) {
                     *outStatus = kCBLStatusBadAttachment;
                     return nil;
                 }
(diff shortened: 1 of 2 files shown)


Source/CBL_Attachment.m
```diff
@@ -101,8 +101,13 @@ - (instancetype) initWithName: (NSString*)name
...
             if (revPosObj) {
-                int revPos = [$castIf(NSNumber, revPosObj) intValue];
-                if (revPos <= 0) {
+                if (![revPosObj isKindOfClass: [NSNumber class]]) {
+                    *outStatus = kCBLStatusBadAttachment;
+                    return nil;
+                }
+                int revPos = [revPosObj intValue];
+                // PouchDB has a bug that generates "revpos":0; allow this (#1200)
+                if (revPos < 0) {
```
pasin: I always has a problem remembering the logic around stub and follows but when we allows revpos = 0, do we also need to change the logic in the line below as well?

https://github.com/couchbase/couchbase-lite-ios/blob/0705423d3d030b208e87f48ee63725c7b553f7ca/Source/CBLDatabase%2BAttachments.m#L257

snej: It's already possible for an incoming doc attachment to end up with a revpos of 0, if its attachment dict doesn't have a `revpos` property at all. The error case I'm fixing is one where there _is_ a `revpos` but its value is 0. So I don't think this patch allows any new conditions that we didn't get before.

Issue #1074, opened 2016-01-28
Couldn't unzip compressed message


cflorion:
As I was testing my app once, at one point it stopped syncing. I looked at the logs and saw the following message:
`
2016-01-28 11:47:07.168 CBLIS iOS Test App[2937:1179248] WARNING: CBLWebSocketChangeTracker: Couldn't unzip compressed message; status=-3
`

 This line kept being logged over and over again as I was using the app. I had to relaunch the app in order for it to work again. I haven't succeeded to reproduce this bug yet.
I am posting this message to know if you ever encountered this and if this is something that can happen and that is not that serious ?

Thanks :)

zgramana:
@cflorion can you provide full logs, including the version of CBL that you're running? And which version of SG are you running against?

snej:
That warning is at CBLWebSocketChangeTracker:187 (in -webSocket:didReceiveMessage:.) It indicates gzip decompression ("inflate") got a zlib error -3, which is Z_DATA_ERROR; interpretation is that "the input data was corrupted (input stream not conforming to the zlib format or incorrect check value)" (see zlib.h.)

snej:
When that error occurs, the change tracker closes the WebSocket, because once you're out of sync reading the gzip stream you can't recover. So if the warning keeps being logged, it means the error keeps happening on successive WebSocket connections.

It'd be helpful to see the SG logs from that time period too.

cflorion:
@zgramana The version of the CBL is the current master branch (#1060). The SG is the current master branch of a week ago.

Unfortunately I have not been able to reproduce this bug yet, once I do I'll post the logs.

Would it be a good idea to have like a method that reset and retry the connection in order to have a clean state ? It's just strange that the error kept happening over and over again...
Anyways, I'll keep you posted :)

Cheers.

snej:
It _does_ close the connection and reconnect, after this error occurs. That's what makes it especially weird that it happened multiple times…

snej:
A forum user ran into this and was able to [capture some logs](https://forums.couchbase.com/t/14-31-48-705-warning-cblwebsocketchangetracker-couldnt-unzip-compressed-message-status-3/7454/5).

K-Leon:
I stumbled over this bug also. The logs are equal to them the noted forum user collected, but i managed to get a little bit background to what happened and i'm convinced that this bug is related to filesize. 

[...]

PR #1212, opened 2016-04-12
Partially fixed CBLWebSocketChangeTracker unzip errors

snej:
I still don't know why the first error occurs. But after that, the
ChangeTracker wouldn't be able to recover, because it would reconnect
but keep using the same GZip decoder, which wouldn't be in an initial
state. So it would keep getting gzip errors every time. I fixed this
by nil-ing out _gzip in the -start method.

I also improved the flow of control so, after it gets an error and
decides to close the WebSocket, it will ignore subsequent queued
messages to avoid getting confused by them and logging more errors.

For #1074


snej:
Note: Github's diff view makes it look like lots of lines have changed, but most of the line changes are just indentation.

pasin:
I believe this is the issue that I also encountered when testing the tutorial app but the thing is I couldn't reproduce the issue anymore.

PR #1254, opened 2016-05-21
Fixed: doc._conflicts passed to map function may include wrong revID

### LINKED ISSUE #1252: doc._conflicts passed to map function may include wrong revID
There's a case [reported](https://groups.google.com/d/msgid/mobile-couchbase/CACfBfNv1%3DDoXHoMx14T0hgBc%2Bn6jNTQFuQRSdnU-sx343L%3DGuA%40mail.gmail.com) where the `_conflicts` property passed to the map function is a 1-item array containing the same revID as the `_rev` property. That's invalid; it should have a different revID.

I think the problem is in -[CBL_SQLiteViewStorage.m updateIndexes:]. Currently starting at line 445:
```objc
                        if (deleted || [oldRevID compare: revID] > 0) {
                            // It still 'wins' the conflict, so it's the one that
                            // should be mapped [again], not the current revision!
                            revID = oldRevID;
                            deleted = NO;
                            sequence = oldSequence;
                            json = [fmdb dataForQuery: @"SELECT json FROM revs WHERE sequence=?",
                                    @(sequence)];
                        }
                        if (!deleted) {
                            // Conflict revisions:
                            if (!conflicts)
                                conflicts = $marray();
                            [conflicts addObject: oldRevID.asString];
                            while ([r2 next])
                                [conflicts addObject: [r2 stringForColumnIndex:0]];
                        }
```
If the first `if` statement passes, then `revID` will be assigned the value of `oldRevID`. But then if the second `if` also passes, `oldRevID` will be added to the `conflicts` array, even though it's the same as `revID`.

I think what needs to happen instead in this case is that the original `revID` (before being overwritten with `oldRevID` needs to be added to `conflicts`.

snej:
When a new revision of a doc isn't the winner (i.e. it's on a shallower
branch), the view indexing code re-indexes the existing winning revision
instead. It should add the new losing revision to the _conflicts array,
but it was getting mixed up and adding the winning rev. Fixed.

Fixes #1252

### DIFF
--- a/Source/CBL_SQLiteViewStorage.m
+++ b/Source/CBL_SQLiteViewStorage.m
@@ -445,7 +445,9 @@
                         if (deleted || [oldRevID compare: revID] > 0) {
                             // It still 'wins' the conflict, so it's the one that
                             // should be mapped [again], not the current revision!
+                            CBL_RevID* temp = revID;
                             revID = oldRevID;
+                            oldRevID = temp;
                             deleted = NO;
                             sequence = oldSequence;
                             json = [fmdb dataForQuery: @"SELECT json FROM revs WHERE sequence=?",


snej:
@npomfret: Would you be able to try the branch in that pull request and see if it fixes your case? You'd need to check out and build it yourself. If that's too much trouble, let me know, but I want you to have a chance to verify the fix.

npomfret:
I'll figure out how to do it, but I don't know how to reproduce the problem.  I'll let you know if I find anything.

snej:
No problem; I thought you had an easy way to reproduce it.

@pasin, could you review the change? It's very small, but it's in a tricky part of the code...

PR #1272, opened 2016-06-02
Disable App Transport Security for Mac OS X LiteServ

### LINKED ISSUE #1271: Disable App Transport Security on Mac OSX LiteServ
```16:07:32.631‖ CBLRemoteJSONRequest[POST http://192.168.0.111:4985/db/_revs_diff]: Got error NSURLError[-1022, "The resource could not be loaded because the App Transport Security policy requires the use of a secure connection.", <http://192.168.0.111:4985/db/_revs_diff>]<--CFCFNetworkError[-1022]```

Currently 1.3.0+ Mac OSX LiteServ cannot replicated with non ssl enabled endpoints. This need to be disabled for LiteServ builds so the tests can function as before.

Here is how to disable App Transport Security - http://stackoverflow.com/questions/32631184/the-resource-could-not-be-loaded-because-the-app-transport-security-policy-requi

sethrosetter:
Fixes #1271 

This disables App Transport Security by default to enable testing against non-SSL enabled endpoints. I'm not sure about the ramifications for releasing this by default (i.e. Is anyone using Mac OSX LiteServ for anything in production?).

### DIFF
--- a/LiteServ App/LiteServ App-Info.plist
+++ b/LiteServ App/LiteServ App-Info.plist
@@ -30,14 +30,6 @@
 	<true/>
 	<key>NSAppTransportSecurity</key>
 	<dict>
-		<key>NSExceptionDomains</key>
-		<dict>
-			<key>localhost</key>
-			<dict>
-				<key>NSExceptionAllowsInsecureHTTPLoads</key>
-				<true/>
-			</dict>
-		</dict>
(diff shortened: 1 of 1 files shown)


snej:
If you're adding this key you should remove the `NSExceptionDomains` key.

snej:
Sorry, didn't see your new commit!

PR #1330, opened 2016-06-28
Added API to remove OpenID tokens from Keychain

### LINKED ISSUE #1302: OIDC needs an API to clear ID Tokens from the keychain
Without this ability, the OIDC authenticator will continue to use the old ID Token and refresh token so the app cannot switch the user.

snej:
New method is -[CBLReplication removeStoredCredentials:]

Also moved the recently-added .username property from CBLAuthenticator
to CBLReplication -- this makes it consistent that CBLAuthenticator
has no API of its own but has to be accessed through the replication object.

Fixes #1302


snej:
Oops, forgot to target this at `dev` not `master`. I'll merge it into `dev` afterwards.

NOTE: The method name is provisional.

PR #1332, opened 2016-06-29
Newly-pulled docs shouldn't show up in pusher's pendingDocuments

### LINKED ISSUE #1274: pendingDocumentIDs is incorrect
When syncing database for the first time (just pulling), `pendingDocumentIDs` of push replication contains all documents which were pulled.
`changesCount` is actually 0
When stopping/starting replication `pendingDocumentIDs` are cleaned up.

Also, noticed that `pendingDocumentIDs` sometimes don't clean up properly and some documents are never removed

--------
* Version:   1.2.1
* Client OS: iOS 9.3.1
* Server:    4.1

snej:
When the pusher is notified of a new revision, but decides to ignore it
because it's an 'echo' from a pull replication (or because it's rejected
by the filter), it should advance the checkpoint past that rev's
sequence.

This is a good optimization in general, to prevent the next push from
re-scanning the same revisions, but more importantly it tells
CBLReplication.pendingDocuments that those revisions aren't pending.

To fix this I did have to change the timing at which the pusher adds
revs to its internal _pendingSequences set, because it needs to happen
at the same time that skipped sequences are being handled.

Fixes #1274

### DIFF
--- a/Unit-Tests/Replication_Tests.m
+++ b/Unit-Tests/Replication_Tests.m
@@ -606,7 +606,7 @@ static UInt8 sEncryptionIV[kCCBlockSizeAES128];
                                    NSHTTPCookieOriginURL: remoteDbURL,
                                    NSHTTPCookiePath: remoteDbURL.path,
                                    NSHTTPCookieValue: @"logmein",
-                                   NSHTTPCookieExpires: [NSDate dateWithTimeIntervalSinceNow: 10]
+                                   NSHTTPCookieExpires: [NSDate distantFuture]
                                    }];
 
     NSHTTPCookie* cookie2 = [NSHTTPCookie cookieWithProperties:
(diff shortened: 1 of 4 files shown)


Unit-Tests/Replication_Tests.m
```diff
@@ -606,23 +606,23 @@ - (void) test10_ReplicationCookie {
                                    NSHTTPCookieOriginURL: remoteDbURL,
                                    NSHTTPCookiePath: remoteDbURL.path,
                                    NSHTTPCookieValue: @"logmein",
-                                   NSHTTPCookieExpires: [NSDate dateWithTimeIntervalSinceNow: 10]
+                                   NSHTTPCookieExpires: [NSDate distantFuture]
```
snej: Note: Not related to this fix, but I think this will prevent intermittent failures in this unit test. (I got one of those failures just now.)

Issue #1334, opened 2016-06-29
Scope cookie storage by database, not by replication


snej:
Currently every replication has its own cookie store. So if you have a push and a pull replication to the same server, each has independent cookie storage. (This came about as a side effect of using the replicator's checkpoint key as the ID of the cookie store.) 

It would make more sense to make the cookie store per-_database_, so all replications involving the same db would share cookies.

This allows us to put the new "clear stored credentials" method on Database, not Replication, which seems clearer. In the future we'll also be able to move the cookie accessor methods there.

snej:
- [ ] Add a private `cookieStorage` property on CBLDatabase that returns a single instance, lazily initialized. The ivar could be weak, as long as the store is saved before the last strong reference goes away.
- [ ] This should only be called on the background CBLDatabase (the one the CBLReplicator uses), to avoid having to keep multiple instances in sync across multiple threads.
- [ ] The state should be saved to a single database info key with a hardcoded name.
- [ ] On first access, it should look for cookie state in the old location and coalesce it.
- [ ] On first save, it should delete the state from the old location.
- [ ] CBLReplication can be simplified because it won't need to wait to access the cookie store till its `_bg_Replicator` is created — instead it can access the cookie store at any time by running a block on the background thread and getting the bg database's cookieStorage property.

pasin:
@snej From the first 2 bullet points, instead of each database object (of the same name) has its own cookie store object, would it be easier to have one single shared cookie storage kept in the CBL_Shared?

Note that current cookie store loads the cookies into the memory during the initialization and keeps maintain the memory data. If each database (of the same name) has its own cookie store object, we need to change the cookie store.

snej:
Yes, I was thinking that it could be a shared per-database-file object kept by CBL_Shared. Its API would need to be thread-safe, but IIRC it already is?

pasin:
Yes, the cookie storage is thread safe.

PR #1342, opened 2016-07-05
Change cookie storage scope from replicator to database

pasin:
- Added cookieStorages properties to the internal CBLDatabase.
- Cookies are stored in the local database info document under the key named `cookies`.
- Implemented migrateOldCookieStorage to migrate cookies from checkpoint docs to the new local data info document. The old cookies are removed after the migration.
- CBLReplication always accesses the database cookie storage only from the background thread.

#1334

### DIFF
--- a/Source/API/CBLReplication.m
+++ b/Source/API/CBLReplication.m
@@ -30,7 +30,6 @@
 
 
 typedef void (^PendingAction)(id<CBL_Replicator>);
-typedef void (^PendingCookieAction)(CBLCookieStorage*);
 
 
 NSString* const kCBLReplicationChangeNotification = @"CBLReplicationChange";
@@ -54,6 +53,11 @@ NSString* CBL_ReplicatorStoppedNotification = @"CBL_ReplicatorStopped";
 @property (nonatomic, readwrite) NSString* username;
 @end
 
+@interface CBLReplication (Internal)
+// For unit test only:
+@property (nonatomic, readonly) NSArray* cookies;
+@end
+
 
 @implementation CBLReplication
 {
@@ -65,7 +69,6 @@ NSString* CBL_ReplicatorStoppedNotification = @"CBL_ReplicatorStopped";
     // ONLY used on the server thread:
     id<CBL_Replicator> _bg_replicator;
     NSMutableArray<PendingAction>* _bg_pendingActions;
-    NSMutableArray<PendingCookieAction>* _bg_pendingCookieActions;
     NSConditionLock* _bg_stopLock;
 }
 
@@ -211,15 +214,23 @@ NSString* CBL_ReplicatorStoppedNotification = @"CBL_ReplicatorStopped";
 
 - (void) deleteCookieNamed: (NSString*)name {
     Assert(name);
-    [self tellCookieStorage: ^(CBLCookieStorage *storage) {
+    [self tellCookieStorage: ^(CBLCookieStorage* storage) {
         [storage deleteCookiesNamed: name];
     }];
 }
 
 
 - (void) deleteAllCookies {
-    [self tellCookieStorage: ^(CBLCookieStorage *storage) {
+    [self tellCookieStorage: ^(CBLCookieStorage* storage) {
         [storage deleteAllCookies];
+    }];
+}
+
+
+// For unit test only:
+- (NSArray*) cookies {
+    return [self tellCookieStorageAndWait:^id(CBLCookieStorage* storage) {
+        return storage.cookies;
     }];
 }
 
@@ -582,15 +593,17 @@ NSString* CBL_ReplicatorStoppedNotification = @"CBL_ReplicatorStopped";
     }];
 }
 
-- (void) tellCookieStorage: (PendingCookieAction)block {
-    [self tellReplicator: ^(id<CBL_Replicator> bgRepl) {
-        if (bgRepl) {
-            block(bgRepl.cookieStorage);
-        } else {
-            if (!_bg_pendingCookieActions)
-                _bg_pendingCookieActions = [NSMutableArray array];
-            [_bg_pendingCookieActions addObject: block];
-        }
+- (void) tellCookieStorage: (void (^)(CBLCookieStorage*))block {
+    [_database.manager.backgroundServer tellDatabaseNamed: _database.name
+                                                       to: ^(CBLDatabase* bgDb) {
+        block(bgDb.cookieStorage);
+    }];
+}
+
+- (id) tellCookieStorageAndWait: (id (^)(CBLCookieStorage*))block {
+    return [_database.manager.backgroundServer waitForDatabaseNamed: _database.name
+                                                                 to: ^id(CBLDatabase* bgDb) {
+        return block(bgDb.cookieStorage);
     }];
 }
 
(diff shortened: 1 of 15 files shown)


Source/API/CBLReplication.m
```diff
@@ -54,6 +53,11 @@ @interface CBLReplication ()
 @property (nonatomic, readwrite) NSString* username;
 @end
 
+@interface CBLReplication (Internal)
+// For unit test only:
+@property (nonatomic, readonly) NSArray* cookies;
```
snej: Please move this to CouchbaseLitePrivate.h.

Source/CBLDatabase+Internal.m
```diff
@@ -813,4 +817,67 @@ - (void) postNotification: (NSNotification*)notification {
...
+#pragma mark - Database Info Local Document:
+
+
+- (BOOL) putLocalDatabaseInfoWithKey: (NSString*)key
+                               value: (id)value
+                            outError: (NSError**)outError {
+    if (key == nil || value == nil)
+        return NO;
+    
+    NSMutableDictionary* document = [NSMutableDictionary dictionaryWithDictionary:
```
snej: This will crash if the document doesn't already exist.

Source/CBLBlipReplicator.m
```diff
@@ -140,6 +139,12 @@ - (void) databaseClosing {
 @synthesize savingCheckpoint=_savingCheckpoint, active=_active;
 #endif
 
+- (CBLCookieStorage*) cookieStorage {
```
snej: I don't think it's worth caching our own reference — it's easy enough to call `_db.cookieStorage` instead of `self.cookieStorage`.

Source/CBLCookieStorage.m
```diff
@@ -245,13 +248,39 @@ - (BOOL) deleteCookie: (NSHTTPCookie*)aCookie outIndex: (NSUInteger*)outIndex {
...
+                 self, error.my_compactDescription);
+        return success;
+    }
+    return NO;
+}
+
+
 - (void) loadCookies {
-    _cookies = [NSMutableArray array];
+    NSArray* cookies = [_db getLocalDatabaseInfoPropertyValueForKey: kLocalDbInfoCookiesKey];
```
snej: Use `[_db.storage infoForKey:]` instead.

Source/CBLDatabase+Internal.h
```diff
@@ -153,4 +155,22 @@ extern NSArray* CBL_RunloopModes;
 
 - (void) setExpirationDate: (NSDate*)date ofDocument: (NSString*)documentID;
 
+
+// LOCAL DOC DATABASE INFO:
+
+/** Put a property with a given key and value into the local database info document. */
+- (BOOL) putLocalDatabaseInfoWithKey: (NSString*)key
```
snej: We don't need these — can use CBL_Storage's "info" API.

Source/CBLRestReplicator.m
```diff
@@ -82,7 +82,7 @@ @implementation CBLRestReplicator
     CBLCookieStorage* _cookieStorage;
 }
 
-@synthesize db=_db, settings=_settings, cookieStorage=_cookieStorage, serverCert=_serverCert;
+@synthesize db=_db, settings=_settings, serverCert=_serverCert, cookieStorage=_cookieStorage;
```
snej: I don't think we need `_cookieStorage` anymore. Can just access `_db.cookieStorage`.

Unit-Tests/CookieStorage_Tests.m
```diff
@@ -16,6 +18,12 @@ @interface CookieStorage_Tests : CBLTestCaseWithDB
 
 @end
 
+
+@interface CBLCookieStorage (Internal)
+- (NSString*) localDatabaseInfoCookiesKey;
```
snej: Don't redeclare this; it should be imported from CouchbaseLitePrivate.h or CBLInternal.h.

pasin:
@snej Thanks for the code review. I made changes per your comment.

Note that I declare CBLCookieStorage `reset` method in `CBLInternal.h` instead of `CouchbaseLitePrivate.h` as CouchbaseLitePrivate.h doesn't seem to include any declaration for NON-public API interfaces.

Source/CBLCookieStorage.m
```diff
@@ -364,6 +402,17 @@ - (void) notifyCookiesChanged {
                                                       userInfo: nil];
 }
 
+
+- (NSString*) databaseInfoCookiesKey {
```
snej: I don't think this is used anywhere; can it be removed?

Source/CBLDatabase+Internal.m
```diff
@@ -54,6 +55,9 @@
 // How long to wait after a database opens before expiring docs
 #define kHousekeepingDelayAfterOpening 3.0
 
+// Local document for keeping internal database info, e.g. cookies)
+#define kLocalDatabaseInfoDocId @"CBL_DatabaseInfo"
```
snej: This doesn't seem to be used anywhere.

Source/CBLDatabase+Replication.m
```diff
@@ -108,6 +108,26 @@ - (BOOL) putLocalCheckpointDocumentWithKey: (NSString*)key
     return result;
 }
 
+- (BOOL) removeLocalCheckpointDocumentWithKey: (NSString*)key
+                                     outError: (NSError**)outError {
+    if (key == nil)
+        return NO;
+    
+    NSMutableDictionary* document = [NSMutableDictionary dictionaryWithDictionary:
```
snej: This will crash if -getLocalCheckpointDocument returns nil. Instead you can use `[[[self getLocalCheckpointDocument] mutableCopy]`.

Source/CBLInternal.h
```diff
@@ -160,4 +161,9 @@ NSString* CBLKeyPathForQueryRow(NSString* keyPath); // for testing
 - (instancetype) initWithDatabase: (CBLDatabase*)database
                            remote: (NSURL*)remote
                              pull: (BOOL)pull                       __attribute__((nonnull));
+@end
+
+
+@interface CBLCookieStorage ()
+- (void) reset;
```
snej: Why does this method need to be in a separate header? It seems like part of the class's regular API.

pasin: Yes. Actually I can make changes to the existing deleteAllCookies: method to set nil to the database info cookies key.

pasin: Actually I think it's better to rename deleteAllCookies: to reset:.

pasin:
@snej Thanks for the second code review. I removed unused variables methods are removed now and also fixed the potential crash. I changed the exiting deleteAllCookies to reset and included the logic to reset the database info cookies key as well. Please rereview when you have time.

pasin:
Rebase push as one fix didn't go through.

pasin:
Unit tests failed due to the change to the put / remove local checkpoint. I'm checking it.

pasin:
OK, I rebased the fixes. It's ready for re-review now. Thanks!

snej:
Great!

Source/CBLCookieStorage.h
```diff
@@ -25,7 +25,7 @@ extern NSString* const CBLCookieStorageCookiesChangedNotification;
 
 /** Creates a cookie storage that will store cookies inside the given database as a local document 
     referenced by a unique storage key. */
-- (instancetype) initWithDB: (CBLDatabase*)db storageKey: (NSString*)storageKey;
+- (instancetype) initWithDB: (CBLDatabase*)db;
```
borrrden: This is a change to the public API!

snej: CBLCookieStorage is not a public class.

borrrden: Oops, my bad.  For some reason it is in .NET.  I forgot Obj-C classes are in the API folder.

PR #1415, opened 2016-08-23
Enable Application Extension API Only in Couchbase Lite framework target

pitiphong-p:
Couchbase Lite is a database framework which can be embeded in Extensions. So I think Couchbase List should use only Extension APIs.
After testing I found that Couchbase List doesn't use any Application only APIs so that I enable the `APPLICATION_EXTENSION_API_ONLY` build setting and submit it as this PR.

### DIFF
--- a/CouchbaseLite.xcodeproj/project.pbxproj
+++ b/CouchbaseLite.xcodeproj/project.pbxproj
@@ -5946,7 +5946,6 @@
 			isa = XCBuildConfiguration;
 			baseConfigurationReference = 27FFFF151C973FB5001A70FC /* iOS framework_Debug.xcconfig */;
 			buildSettings = {
-				APPLICATION_EXTENSION_API_ONLY = YES;
 			};
 			name = Debug;
 		};
@@ -5954,7 +5953,6 @@
 			isa = XCBuildConfiguration;
 			baseConfigurationReference = 27FFFF161C973FB5001A70FC /* iOS framework_Release.xcconfig */;
 			buildSettings = {
-				APPLICATION_EXTENSION_API_ONLY = YES;
 			};
 			name = Release;
 		};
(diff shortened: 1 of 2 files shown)


pasin:
@pitiphong-p, Thanks for the PR. There are a few things.

1. couchbase-lite-ios uses xcconfig files for project/target build settings. Any build settings should be written into the xcconfig files under the`xcconfigs` folder.

2. I have never written an app extension before. Without `APPLICATION_EXTENSION_API_ONLY` set to true, is couchbase-lite-ios not able to use to build the extension?

3. There are some API limitations listed in this [documentation](https://developer.apple.com/library/ios/documentation/General/Conceptual/ExtensibilityPG/ExtensionOverview.html#//apple_ref/doc/uid/TP40014214-CH2-SW2), but I couldn't find the full list. What method did you use to check whether CBL doesn't use any non Extention APIs?

@snej what is your opinion about setting APPLICATION_EXTENSION_API_ONLY to true?

pitiphong-p:
@pasin  Thank you for the respond, here are the answers for your questions.

1. I will update my PR to move these setting into xcconfig files.
1. With `APPLICATION_EXTENSION_API_ONLY` set to TRUE, compiler will help us validate that we didn't use any non Extension compatible APIs. Also if you link any Extension compatible target to the target that didn't set this option to true, the linker will warn you with `Linking against a dylib which is not safe for use in application extensions` warning message
1. The compiler checked that for me, we don't have to validate every lines of code by ourself.

Hope these answers and this PR help :)

snej:
Looks good, as long as we first verify that everything builds with this setting turned on.
Also, it may need to be in multiple xcconfig files, because we have multiple iOS targets.

pitiphong-p:
I put that setting in the common `iOS Framework` xcconfig file which should already be included in every iOS targets.

snej:
We can't add `APPLICATION_EXTENSION_API_ONLY` to our targets, because it disables important replicator functionality (detecting when the app goes online/offline, and requesting a background session to get more time.)
You can modify the xcconfig file yourself to add this setting, then build the framework.

pasin:
Per Jens's comment, I'm closing the PR. Thanks.

pitiphong-p:
OK, thank you

Issue #1424, opened 2016-08-30
SQLite database file corruption


brendand:
I'm getting some complaints from customers that their SQLite database files are becoming corrupted somehow. It sometimes manifests itself with missing views or errors about map blocks. 

I'm pretty sure I'm being very careful to access the database on the main thread in most places unless I'm doing some sort of batch operation in which case I call backgroundTellDatabaseName and create a new document and CBLModel object from a passed in document ID within the block. 

Here's an example screenshot showing one issue. Notice the duplicate `rowid` value? And the null key and null sequence values.

<img width="888" alt="screen shot 2016-08-30 at 3 09 35 pm" src="https://cloud.githubusercontent.com/assets/586075/18107283/9105dff0-6ec4-11e6-954d-b37a56e6ef4d.png">

What could be causing this and how do I prevent it from happening?

I do have `autosave` on my `CBLModel` subclass turned on, so maybe that's causing an issue since I do explicit saves in a bunch of places too. I've turned that feature off for my next update, just in case it's causing some conflict. Or maybe it's being caused by the database access on the sync thread?

The above screenshot came from a database who's log journal mode was "WAL". 

The end result of things like this are that the data doesn't show in the app.

Any help or things I should be looking out for would be very welcome.

--------
* Version:   1.3 GM
* Client OS: OS X 10.11
* Server:    n/a

pasin:
I'm not sure how the NULL key can happen. According to the [code](https://github.com/couchbase/couchbase-lite-ios/blob/master/Source/CBL_SQLiteViewStorage.m#L576-L579), the null key should be ignored.

Duplicated rows seem to be a threading issue. You may check if the app can query the same view created by the same database object on 2 different threads.

> I do have autosave on my CBLModel subclass turned on, so maybe that's causing an issue since I do explicit saves in a bunch of places too.

I have found [this comment](https://github.com/couchbase/couchbase-lite-ios/blob/1679c4436699aa837509c5706ca49a05e744cd04/Source/CBL_SQLiteViewStorage.m#L574-L577) in the code. @snej has this been fixed?

brendand:
It often just manifests itself with a "malformed disk image" error too.

I do have a central manager class (a subclass of NSDocument & UIDocument) where I create some views that are used to fetch and cache a bunch of CBLModel objects that I use throughout the app. Whenever a sync happens I flush and reload those model objects, but again, on the main thread.

[...]

PR #1430, opened 2016-09-07
Updated FMDB to master

### LINKED ISSUE #1389: Allow SQLite reads concurrently with writer
The locking code I added to FMDatabase explicitly blocks readers while there is a writer. This matches the default behavior of SQLite, but I just noticed that when the WAL is enabled, SQLite does allow reads concurrently with writes. The readers will see a snapshot of the database from before the in-progress write transaction:

>Writers merely append new content to the end of the WAL file. Because writers do nothing that would interfere with the actions of readers, writers and readers can run at the same time. However, since there is only one WAL file, there can only be one writer at a time. [*](https://www.sqlite.org/wal.html)

Changing the FMDatabase locking accordingly would be a big performance/availability win. For example, currently the database is entirely unavailable while a view index is being updated (any reads will block.) With this change, reads will work normally during indexing. The same goes for large inserts that might happen during a pull replication.

snej:
Adds 3 changes:
* Allow threads to read while another thread is writing (fixes #1389) — https://github.com/couchbaselabs/fmdb/commit/943213af3e676d23e75706333f246bd105de31ff
* Fixed a memory leak of FMDatabase.databaseLock when closing db — https://github.com/couchbaselabs/fmdb/commit/8b5814fc55c09fef227cb7e28c64a457c19e74e5
* More rigorous thread-safety checks (for #1424) — https://github.com/couchbaselabs/fmdb/commit/b859ca3796090f1a2dc43046be84a55b4403889a

### DIFF
--- a/README.md
+++ b/README.md
@@ -96,10 +96,10 @@ Can be included in any iOS project via [Carthage](https://github.com/carthage/ca
 Add the following line to your Cartfile:
    
 ```
-github "couchbase/couchbase-lite-ios" "release/1.2.0"
+github "couchbase/couchbase-lite-ios" "release/1.3.0"
 ```
 
-Run `carthage update --platform ios`, and you should now have the latest version of Couchbase Lite iOS in your Carthage folder.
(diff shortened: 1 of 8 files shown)


snej:
This got messed up when I tried to change the target branch to dev. I'll open a new one,

PR #1435, opened 2016-09-12
Patch 1

pabitranjan:


pasin:
@pabitranjan 

Could you please explain what the issue you are encountering so that the fix is needed? Could you please also do rebase instead of merge to make the PR cleaned?

pabitranjan:
@pasin 

Thanks for reviewing this. I added changed file with description in this pull request "#1438".

PR #1438, opened 2016-09-13
Fixed the "Memory leak" issue of CBLRemoteRequest delegate object in CBL iOS 1.3

pabitranjan:
**Issue:-**
- I encounter the memory leak while debugging in instrumentation tool and that is because, strong ownership of CBLRemoteRequestDelegate object.
- In below image, we have encountered the root cause for leak.

screen shot 2016-09-14 at 2 19 21 pm

screen shot 2016-09-14 at 2 19 51 pm

**Solution:-**
- To avoid the "memory leak" for **retain cycle(Strong ownership)** for CBLRemoteRequest delegate object.

### DIFF
--- a/Source/CBLRemoteRequest.h
+++ b/Source/CBLRemoteRequest.h
@@ -30,7 +30,7 @@ void CBLWarnUntrustedCert(NSString* host, SecTrustRef trust);
     NSMutableURLRequest* _request;
     id<CBLAuthorizer> _authorizer;
     CBLCookieStorage* _cookieStorage;
-    id<CBLRemoteRequestDelegate> _delegate;
+    __weak id<CBLRemoteRequestDelegate> _delegate;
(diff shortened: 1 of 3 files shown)


snej:
Have you actually found leaks involving these objects, or are they just reported as a cycle while they're in use? We have some code where objects have cycles, but the cycles are broken when they're closed / cleaned up.

pabitranjan:
@snej, Thanks for reviewing this. 
Yes, I am getting leak because this retain cycles even after releasing(Closing/cleaning up) the controller from memory.
In above image , we can see only CBL prefix object in the memory leak.

So after making the **weak** reference to the **CBLRemoteRequestDelegate** object, there is no memory leak.

snej:
OK. I'll need to check out your commit and run our unit test suite to make sure it doesn't break anything.

Issue #1500, opened 2016-10-31
Bulk-fetching hammers servers that are slow to reply


avishnyak:
I have noticed the following behavior during first-time replication:

CBL fetches the list of changes and queues 12 bulk-fetch requests of 100 documents each.  If those replies take over 1 second to respond (on my test machine), the rest of the documents to be fetched queue up into the 13th bulk-fetch sending a request for THOUSANDS of documents at once to an already busy sync server.

I have isolated the problem down to this line: https://github.com/couchbase/couchbase-lite-ios/blob/master/Source/CBLRestPuller.m#L391

Instead of throwing ALL of the queued revs into the request there should be some sane max limit there as well (e.g. 100).

Here is a snippet from the logger showing the issue:

```
... snip 12 previous requests for 100 items ...
16:28:24.057‖ Sync: CBLRestPuller[http://*****]: postProgressChanged (191/8549, active=1 (batch=77, net=39), lastSeq=455029, online=1, error=(null))
16:28:24.058‖ Sync: CBLRestPuller[http://*****] bulk-fetching 7300 remote revisions...
16:28:24.095‖ Sync: CBLBulkDownloader[(null)]: POST _bulk_get?revs=true&attachments=true  [{"id":"container_365180505a914b52b480be26c12b39cf_AuditLog","rev":"1-55E21DAA6DE150AF1DD924AC2029D587"},
... snip...
```

--------
* Version:   1.3
* Client OS: Couchbase Lite on any version of iOS
* Server:    Custom

snej:
> sending a request for THOUSANDS of documents at once to an already busy sync server.

This should actually be _less_ work on the server than getting tens of requests for 100 docs each; it's the same number of database calls, docs to process and bytes to respond with, but doesn't have the overhead of all the extra HTTP requests.

But that assumes that SG will stream the response instead of assembling it in memory as one giant array and then serializing it when it's done — if it does the latter, it's going to be bad for memory usage and latency. @adamcfraser, @ajres, do you know whether the `_bulk_get` handler is optimized for this or not?

snej:
Sorry, just realized that the 2nd paragraph above is wrong. CBL requests the `_bulk_get` response as MIME multipart, not JSON. That means it should be streamed out by SG.

@avishnyak, can you describe what actual problem you observe as a result of the existing behavior?

avishnyak:
[...]

PR #1502, opened 2016-11-02
Fix/bulk size incorrect

Cyclic:
See #1500


snej:
I made a basically identical change, just renaming the constant for clarity.

Issue #1461, opened 2016-09-26
ForestDB: Assertion failure in -[MYBackgroundMonitor beginBackgroundTaskNamed:]() MYBackgroundMonitor.m:66


pahmed:
This crash happens with me a lot sometimes after sending the app to background

```
Assertion failure in -[MYBackgroundMonitor beginBackgroundTaskNamed:](), /Users/jenkins/jenkins/workspace/couchbase-lite-ios-builds/couchbase-lite-ios-community/vendor/MYUtilities/MYBackgroundMonitor.m:66
```

The backtrace is attached

[forestdb_bg_crash.txt](https://github.com/couchbase/couchbase-lite-ios/files/493306/forestdb_bg_crash.txt)

---
- Version:   1.3.1 (ForestDB)
- Client OS: iOS 9

snej:
Hm, this was supposed to be fixed in 1.3; see #1152 and #1368.

snej:
```
* thread #1: tid = 0x62116, 0x232cfc5c libsystem_kernel.dylib`__pthread_kill + 8, queue = 'com.apple.main-thread', stop reason = signal SIGABRT
    frame #0: 0x232cfc5c libsystem_kernel.dylib`__pthread_kill + 8
    frame #1: 0x23375b46 libsystem_pthread.dylib`pthread_kill + 62
    frame #2: 0x232640c4 libsystem_c.dylib`abort + 108
    frame #3: 0x00928af6 xxxx`uncaught_exception_handler + 26
    frame #4: 0x23641e48 CoreFoundation`__handleUncaughtException + 644
    frame #5: 0x22dff086 libobjc.A.dylib`_objc_terminate() + 174
    frame #6: 0x22df0e16 libc++abi.dylib`std::__terminate(void (*)()) + 78
    frame #7: 0x22df08f6 libc++abi.dylib`__cxa_rethrow + 102
    frame #8: 0x22dfef46 libobjc.A.dylib`objc_exception_rethrow + 42
    frame #9: 0x2355024e CoreFoundation`CFRunLoopRunSpecific + 650
    frame #10: 0x2354ffbc CoreFoundation`CFRunLoopRunInMode + 108
    frame #11: 0x24b6caf8 GraphicsServices`GSEventRunModal + 160
    frame #12: 0x27c8a434 UIKit`UIApplicationMain + 144
[...]

Issue #1443, opened 2016-09-16
Handle iOS file protection gracefully when in background


snej:
Detect when the database files are not accessible (i.e. app is in background / device is locked) and don't perform operations that would open db files.

To some degree it may be cleaner to just try and fail; we'll have to evaluate that. (There can certainly be race conditions involved with preflighting.)

snej:
See #1433, #1422 for side effect bugs caused by accessing files when they're un-openable.

brendand:
Not sure if this is relevant to this issue, but my Apple Watch app requires access to the database when the app is running in the background. In fact, the Apple Watch app will launch the iOS app in the background if it's not already running and will establish a background database connection on its own dispatch_queue. So if whatever change you're making won't allow background access to the database then that could be an issue. Sorry if I've misunderstood or don't know all the details about iOS file protection.

basememara:
If your app requires background access before and you have your database protection level set to 'NSFileProtectionCompleteUnlessOpen' (the Couchbase default), then it almost never worked for you and you should see existing crash reports. This change should not make it crash but silently ignore if the database cannot be opened. This is important because if your app crashes in the background, it will not be woken up by iOS anymore until he user explicitly opens the app again.

brendand:
I haven't noticed any crashing when my device is locked and I access my app from my Apple Watch. It just works (so far).

snej:
I don't know the details, but I _think_ that if the OS explicitly wakes up your app to do work, it first gives it access to its files. The problems I know of stem from the app already running in the background when the user locks the device, which causes file access to be lost.

jamiltz:
@brendand I'm definitely

jamiltz:
Oops sorry, I pressed the wrong button.

@brendand The remote access to documents from the apple watch sounds great!

npomfret:
I get a 500 error on mine sometimes during start up.  I think it might be when the app is being started in the background (though I don't know what causes it to start in the background). Is this likely to cause it?

I'm starting the CBLManager with `CBLManagerOptions options = {NO, NSDataWritingFileProtectionCompleteUntilFirstUserAuthentication}`.

npomfret:
Has there been any progress on this?

[...]

Issue #1422, opened 2016-08-29
ForestDB crashes frequently in the background because of protection data level


basememara:
We came from SQLite `storageType` and things seemed to be more under control while we sprinkle `UIApplication.sharedApplication().protectedDataAvailable` throughout our app so things don't crash while the protected data is unavailable in the background.

Switching to `ForestDB`, things were noticeably fast, but noticing more frequent crashes while in the background. My thought is because `ForestDB` tries to do a lot of unattended maintenance in the background like auto-compaction and other things automatically.

However, it never checks to compare the protection data availability with its protection level set on the file, which is `NSFileProtectionCompleteUnlessOpen` by default. This I believe is an oversight by Couchbase Lite for iOS in general, the core code should constantly be checking the `protectedDataAvailable` otherwise Couchbase crashes the app in the background for the majority of test cases.

Any ideas how to suspend ForestDB in the background or turn off auto-compaction? Anything else I can do to make Couchbase Lite more robust and conscious of the protect data state while in the background? 

Here are some of the crash reports:

```
0  MyApp                        0x1001b7a04 CLSProcessRecordAllThreads + 1572
1  MyApp                        0x1001b7a04 CLSProcessRecordAllThreads + 1572
2  MyApp                        0x1001b78c0 CLSProcessRecordAllThreads + 1248
3  MyApp                        0x1001a8070 CLSHandler + 48
4  MyApp                        0x1001b5ac8 __CLSExceptionRecord_block_invoke + 92
5  libdispatch.dylib              0x18110947c _dispatch_client_callout
6  libdispatch.dylib              0x181114728 _dispatch_barrier_sync_f_invoke
7  MyApp                        0x1001b556c CLSExceptionRecord + 212
8  MyApp                        0x1001b509c CLSTerminateHandler() + 616
9  libc++abi.dylib                0x180d16f44 std::__terminate(void (*)()) + 16
10 libc++abi.dylib                0x180d16fd0 std::set_new_handler(void (*)()) + 90
11 MyApp                        0x10073d8a8 __clang_call_terminate + 14
12 MyApp                        0x10076eec0 c4View::c4View(c4Database*, cbforest::slice, cbforest::slice, fdb_config const&, cbforest::slice) + 432
13 MyApp                        0x10076d010 c4view_open + 156
14 MyApp                        0x10074443c -[CBL_ForestDBViewStorage openIndexWithOptions:status:] + 368
[...]

Issue #1405, opened 2016-08-17
500 database error (error opening!: 23) when saving a new document in app pause state


HanzhiDou:
Hi,
- Version:   1.2.1 by Couchbase Lite Phonegap plugin
- Client OS: ios 9.3 (it happens also in earlier versions)

My ios cordova app uses the couchbase lite phonegap plugin, sometimes it can not save documents in app pause state(app is running in background),  the restful save request get error response as below:

`{ ... "status":500,"statusText":"Internal Server Error","responseText":"{\"status\":500,\"error\":\"Database error!\"}","responseXML":null,"responseBytes":null}`

There is also one log: 

`error opening!: 23`

However the save operation can succeed once app resumes from background mode,  so just some documents are lost from user point of view.  After further investigation and many tests,  I observed the error occurred only when:
- app in background mode
- the database has never been read before the save document operation

Workaround: I add one query view operation just after the database ready in app startup phase, the error disappeared. 

Though the workaround works, it's better to find the root cause and fix it.

Thanks!

HanzhiDou:
The error in console log:
`error opening!: 23
`
The app uses background location updates, once new location available, it save the info to cbl.

background task is used to make the app running in background (beginBackgroundTaskWithExpirationHandler).

There is no attachment in this database, and the save operation is happened only after cbl ready and location updates are manual started.

snej:
We don't have a REST API for changing the file protection settings of the database files. I think we'd need to add one, for this case. Then you can use a more lenient setting that allows you to access files while in the background.

snej:
Note: We'd implement this along with the other proposed REST API extension to specify the db storage type, encryption key, etc.

npomfret:
I'm also seeing thousands of these messages in the log

`<Notice>: error opening!: 23
`
Is there a workaround?

snej:
> On Sep 27, 2016, at 1:47 AM, Nick Pomfret <notifications@github.com> wrote:
> 
> I'm also seeing thousands of these messages in the log
> 
> <Notice>: error opening!: 23
> 
> Is there a workaround?
> 

Yes, but it’s not available to PhoneGap apps unless you modify the native plugin code. You need to change the file protection mode of the CBLManager to one of the modes that allows file access after the app has been backgrounded. (Sorry, I don’t have the long names of those enums memorized!) We’re aware that this should be accessible from the REST API and have an issue open on that.

[...]

PR #1522, opened 2016-11-29
Fix/file protection

snej:
#### Better support for iOS file protection and backgrounding:

* Enabled file protection in the iOS test app (note: only takes effect on a real device; simulator does not implement file protection!)
* The test app logs when app state changes in various ways, like going into/out of the background or files becoming inaccessible.
* Made MYBackgroundMonitor work better if the app is already backgrounded and if the background task expires. (See MYUtilities commit)
* Added CBLDatabase.fileProtection property
* REST replicator monitors when files become inaccessible and suspends the replicator for the duration.
* Thread-safety improvements in CBLRestReplicator+Backgrounding: make sure UIApplication is only called on the main thread.
* Defined new internal status kCBLStatusFilesystemLocked for when we know that file I/O failed due to file protection. It gets reported as a 500 status with message “Device locked”.

#### Replicator should save checkpoint _before_ going idle

Otherwise it doesn’t save the checkpoint if it’s finishing a replication in a background task (the process will be suspended before the timer goes off.)

#### Replicator: Fixed some logic in -reachabilityChanged handling _suspended

When _suspended is true it should always call goOffline regardless of trustReachability.

### DIFF
--- a/CBL Test.entitlements
+++ b/CBL Test.entitlements
@@ -1,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<dict>
+	<key>com.apple.developer.default-data-protection</key>
+	<string>NSFileProtectionComplete</string>
+</dict>
+</plist>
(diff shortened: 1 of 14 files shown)


snej:
Related to #1461 (background monitor); #1443 #1422 #1405 (file protection).

snej:
Rebased, added some issue links in the commits, pushed the latest MYUtilities. Ready for you, @pasin.

pasin:
@snej MYUtilities link is still not valid.

PR #1534, opened 2016-12-21
Fix #1429 crash in CBLRemoteSession

### LINKED ISSUE #1429: 1.3.1-6: crashed EXC_BAD_ACCESS KERN_INVALID_ADDRESS 0x0000000000000000
Here's what got reported in crashlytics

```
Crashed: CouchbaseLite
0  CoreFoundation                 0x18296c480 -[__NSSetM removeObject:] + 144
1  pomochat                       0x10018bf50 __57-[CBLRemoteSession URLSession:task:didCompleteWithError:]_block_invoke (CBLRemoteSession.m:321)
2  Foundation                     0x18346002c __NSThreadPerformPerform + 340
3  CoreFoundation                 0x182a4109c __CFRUNLOOP_IS_CALLING_OUT_TO_A_SOURCE0_PERFORM_FUNCTION__ + 24
4  CoreFoundation                 0x182a40b30 __CFRunLoopDoSources0 + 540
5  CoreFoundation                 0x182a3e830 __CFRunLoopRun + 724
6  CoreFoundation                 0x182968c50 CFRunLoopRunSpecific + 384
7  Foundation                     0x183378cfc -[NSRunLoop(NSRunLoop) runMode:beforeDate:] + 308
8  pomochat                       0x100143340 -[CBL_RunLoopServer runServerThread] (CBL_Server.m:193)
9  Foundation                     0x18345fe4c __NSThread__start__ + 1000
10 libsystem_pthread.dylib        0x1826efb28 _pthread_body + 156
11 libsystem_pthread.dylib        0x1826efa8c _pthread_body + 154
12 libsystem_pthread.dylib        0x1826ed028 thread_start + 4
```

pasin:
I couldn't reproduce the issue myself but from the crash log, it seems like _allRequests is gone while it's used. Applied the technique to ensure that _allRequests is retained inside the block.

#1429

### DIFF
--- a/Source/CBLRemoteSession.m
+++ b/Source/CBLRemoteSession.m
@@ -216,8 +216,12 @@ UsingLogDomain(Sync);
     if (_requestIDs.count > 0)
         Warn(@"CBLRemoteSession closed but has leftover tasks: %@", _requestIDs.allValues);
     _session = nil;
-    _allRequests = nil;
     _requestIDs = nil;
+    
+    // Reset _allRequests on the API thread:
+    [self doAsync: ^{
+        _allRequests = nil;
+    }];
 }
 
 
@@ -318,8 +322,7 @@ UsingLogDomain(Sync);
         didCompleteWithError:(nullable NSError *)error
 {
     [self requestForTask: task do: ^(CBLRemoteRequest *request) {
-        NSMutableSet<CBLRemoteRequest*>* allRequests = _allRequests;
-        [allRequests removeObject: request];
+        [_allRequests removeObject: request];
         if (error)
             [request didFailWithError: error];
         else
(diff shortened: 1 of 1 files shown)


snej:
That fix won't work. It does fix the problem of `_allRequests` being dealloced _during_ the block, but if it's been dealloced just _before_ the block enters, retaining it will crash.

I think the actual problem is on line 219, in -URLSession:didBecomeInvalidWithError:, where it sets `_allRequests` to nil. That method runs on the delegate queue, but `_allRequests` is supposed to be used only on the replicator thread (see its declaration.)

I think the fix is to put `_allRequests = nil` in a -doAsync: block, and move that to the end of the method.

pasin:
Thanks. I understand it now. There is race condition between -URLSession:didBecomeInvalidWithError: and the task running with doAsync:.

pasin:
I fixed the PR. Could you please re-review? Thanks.

PR #1536, opened 2016-12-22
Fixed #1236 Implement longpoll and continuous _changes timeout

### LINKED ISSUE #1236: Listener not respecting timeout on longpoll changes
---
- Version:   1.2.1
- Client OS: Mac OSX
- Server:    NA

I issue the following query and there are no new changes so the request remains open listening for new changes. `http://localhost:59840/ls_db1/_changes?feed=longpoll&since=500`. According to the documentation http://developer.couchbase.com/documentation/mobile/current/develop/references/couchbase-lite/rest-api/database/get-changes/index.html, I would expect the request to timeout after 60 second. 

Also, I talked to @pasin and he was fairly certain that iOS does not support timeout. 

Expected: No new _changes, the longpoll changes request would timeout after 60 seconds
Actual: The request remains open

pasin:
* Implement _changes timeout for longpoll and continuous feeds

* The default timeout of 60 seconds applied will be applied if there is no timeout as well as heartbeat param set.

#1236

### DIFF
--- a/Source/CBLRouter+Changes.m
+++ b/Source/CBLRouter+Changes.m
@@ -282,16 +282,15 @@ NSTimeInterval kDefaultChangesTimeout = 60.0;
 - (void) startTimeout {
     assert(_changesTimeout > 0);
     [_changesTimeoutTimer invalidate];
-    CBL_Router* weakSelf = self;
+    __weak CBL_Router* weakSelf = self;
     _changesTimeoutTimer = [NSTimer scheduledTimerWithTimeInterval: _changesTimeout
                                                            repeats: NO block: ^(NSTimer *timer) {
         CBL_Router* strongSelf = weakSelf;
-        if (_changesMode == kLongPollFeed) {
-            [strongSelf sendLongpollResponseForChanges: @[] since: _changesSince];
-        } else {
-            [strongSelf sendContinuousLine: $dict({@"last_seq", @(_changesSince)})];
+        if (_changesMode == kLongPollFeed)
+            [strongSelf sendLongpollResponseForChanges: @[] since: strongSelf->_changesSince];
+        else
+            [strongSelf sendContinuousLine: $dict({@"last_seq", @(strongSelf->_changesSince)})];
             [strongSelf finished];
-        }
     }];
 }
 
--- a/Source/CBLRouter+Changes.m
+++ b/Source/CBLRouter+Changes.m
@@ -23,13 +23,13 @@
 #if DEBUG
 // Make this configurable for testing purposes
 NSTimeInterval kMinHeartbeat = 5.0;
+NSTimeInterval kDefaultChangesTimeout = 60.0;
 #else
 #define kMinHeartbeat 5.0
+#define kDefaultChangesTimeout 60.0
 #endif
 
 @implementation CBL_Router (Changes)
-
-
 
 
 - (CBLStatus) do_POST_changes: (CBLDatabase*)db {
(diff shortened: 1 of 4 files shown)


Source/CBLRouter+Changes.m
```diff
@@ -127,6 +138,15 @@ - (CBLStatus) doChanges: (CBLDatabase*)db {
...
             NSString* heartbeatResponse = (_changesMode == kEventSourceFeed) ? @":\n\n" : @"\r\n";
             [self startHeartbeat: heartbeatResponse interval: heartbeat];
+        } else {
+            // Apply default timeout when heartbeat is not specified:
+            if (_changesTimeout == 0)
+                _changesTimeout = kDefaultChangesTimeout;
+        }
+        
+        if (_changesTimeout > 0) {
+            _changesTimeoutSince = since;
```
snej: I think it would be better to always store the 'since' value in an ivar instead of sometimes in a local. Otherwise if we ever forget to set `_changesTimeoutSince` before calling -startTimeout, it causes a bug.

So basically remove the local var `since` and replace it with the ivar `_changesTimeoutSince` ... actually rename that `_changesSince` because it's now used even without a timeout.

Source/CBLRouter+Changes.m
```diff
@@ -235,21 +257,50 @@ - (void) dbChanged: (NSNotification*)n {
...
+    NSDictionary* body = [self responseBodyForChanges: changes since: since];
+    _response.body = [CBL_Body bodyWithProperties: body];
+    [self sendResponseBodyAndFinish: YES];
+}
+
+
+- (void) startTimeout {
+    assert(_changesTimeout > 0);
+    [_changesTimeoutTimer invalidate];
+    _changesTimeoutTimer = [NSTimer scheduledTimerWithTimeInterval: _changesTimeout
```
snej: I think this creates a ref cycle because self retains the timer, but the timer's block retains self. So if the timeout isn't stopped, it creates a leak.
You could probably make `_changesTimeoutTimer` weak, because the runloop will hold onto it until the timer fires.

pasin:
Thanks. I will fix the PR.

Source/CBLRouter+Changes.m
pasin: I think could use weakSelf and strongSelf technique to fix this.

pasin:
I push up the fix. Could you please re-review? Thanks.

Source/CBLRouter+Changes.m
```diff
@@ -235,21 +256,49 @@ - (void) dbChanged: (NSNotification*)n {
...
+    NSDictionary* body = [self responseBodyForChanges: changes since: since];
+    _response.body = [CBL_Body bodyWithProperties: body];
+    [self sendResponseBodyAndFinish: YES];
+}
+
+
+- (void) startTimeout {
+    assert(_changesTimeout > 0);
+    [_changesTimeoutTimer invalidate];
+    CBL_Router* weakSelf = self;
```
snej: Shouldn't that have `__weak` before it?

Source/CBLRouter+Changes.m
```diff
@@ -235,21 +256,49 @@ - (void) dbChanged: (NSNotification*)n {
...
+    _changesTimeoutTimer = [NSTimer scheduledTimerWithTimeInterval: _changesTimeout
+                                                           repeats: NO block: ^(NSTimer *timer) {
+        CBL_Router* strongSelf = weakSelf;
+        if (_changesMode == kLongPollFeed) {
+            [strongSelf sendLongpollResponseForChanges: @[] since: _changesSince];
```
snej: By accessing `_changesSince` the block implicitly captures `self`, making the cycle reappear … change to `strongSelf->_changesSince`.

Source/CBLRouter+Changes.m
```diff
@@ -235,21 +256,49 @@ - (void) dbChanged: (NSNotification*)n {
...
+        } else {
+            [strongSelf sendContinuousLine: $dict({@"last_seq", @(_changesSince)})];
```
snej: Same issue as line 290.

Source/CBLRouter+Changes.m
pasin: Ah, I forgot to put __weak.

pasin:
@snej I fixed the issue per your comment. Please review again. Thanks.

PR #1544, opened 2017-01-12
[2.0] Implement CBLDatabase and CBLDocument

pasin:
* Create a new xcode project that builds a universal dynamic CBL library.
* Use xcconfigs
* Implement CBLDatabase and CBLDocument

Note:
1. This seems to be a huge PR but the key classes to review are `CBLDatabase`, `CBLDocument`, and `CBLProperties`.  The rest are supported classes, several of which were taken from CBL 1.0 (but only the methods that are used).
2. Currently I made `CBLProperties` a base class of `CBLDocument` (and `CBLSubdocument` which is not yet included in the PR). `CBLProperties` provides the implementation of the property accessors that can be shared b/w `CBLDocument` and `CBLSubdocument`. Alternatively if we do not want to use subclass model, we can change to use component or delegate model.


snej:
I haven't started looking at the code, but it just occurred to me that we should use different class names since the API is changing so much. The best way seems to be to use a different prefix. How about `CB2`? What do you think?

Objective-C/CBLDatabase.mm
```diff
@@ -0,0 +1,373 @@
...
+    return [self initWithName: name
+                      options: [CBLDatabaseOptions defaultOptions]
+                        error: outError];
+}
+
+
+- (instancetype) initWithName: (NSString*)name
+                      options: (CBLDatabaseOptions*)options
+                        error: (NSError**)outError {
+    _name = name;
```
snej: Just noticed this — need to call `[super init]` and check that for nil, as usual in an `init` method.

Objective-C/CBLDatabase.mm
```diff
@@ -0,0 +1,373 @@
...
+    if (_c4db)
+        return YES;
+    
+    if (![self setupDirectory: _options.directory
+               fileProtection: _options.fileProtection
+                        error: outError])
+        return NO;
+    
+    NSString* path = databasePath(_name, _options.directory);    
+    CBLStringBytes bPath = CBLStringBytes(path);
```
snej: Better as `CBLStringBytes bPath(path)` ... the way you wrote it will construct the object and then copy it.

Objective-C/CBLDatabase.h
```diff
@@ -0,0 +1,115 @@
...
+ @param outError  On return, the error if any. */
+- (instancetype) initWithName: (NSString*)name
+                        error: (NSError**)outError;
+
+/** Initializes a CouchbaseLite database with a given name and database options.
+ @param name  The name of the database. May NOT contain capital letters!
+ @param options  The database options.
+ @param outError  On return, the error if any. */
+- (instancetype) initWithName: (NSString*)name
+                      options: (CBLDatabaseOptions*)options
```
pasin: I will change it to nullable.

Objective-C/CBLDatabase.h
```diff
@@ -0,0 +1,115 @@
...
+/** Deletes a database. */
+- (BOOL) deleteDatabase: (NSError**)outError;
+
+/** Deletes a database of the given name in the given directory. */
++ (BOOL) deleteDatabase: (NSString*)name
+            inDirectory: (nullable NSString*)directory
+                  error: (NSError**)outError;
+
+/** Check whether a database of the given name exists in the given directory or not. */
++ (BOOL) databaseExists: (NSString*)name
```
pasin: We don't have this method reviewed yet. I will make this note for the next API review.

Objective-C/CBLDocument.h
```diff
@@ -0,0 +1,53 @@
...
+/** Deletes this document by adding a deletion revision. 
+ This will be replicated to other databases. */
+- (BOOL) deleteDocument: (NSError**)outError;
+
+/** Purges this document from the database; this is more than deletion, it forgets entirely about it.
+ The purge will NOT be replicated to other databases. */
+- (BOOL) purge: (NSError**)outError;
+
+/** Reset changes made to the document. */
+- (void) reset;
```
pasin: We didn't discuss the name. Let's discuss a bit for the next API review.

Objective-C/CBLDatabase.mm
pasin: I will fix this.

pasin:
For the 'CB2' prefix, I'm not getting used to with having number in the prefix name. It's a bit difficult to type, I guessed. What do you think about only 'CB'?

PR #1548, opened 2017-01-14
Fix #1491 change file protection level not effected

### LINKED ISSUE #1491: Sometimes No database query works for iOS
I am using Couchbase lite iOS app with SQLite database. Everything works fine normally. But sometimes when app comes from suspended background session or background (not sure) No query works. Can not debug this as this never happened during development.

All pages on all type of queries returns no result. In my app this can happen when
1. CBLQuery object not generated
2. Query object returns empty results

When I kill the app and run again everything works fine.

I also use continuous replication in background mode (by marking suspended no) when silent notification arrives. May be this is not relevant but just putting as FYI.

I know this is very limited information so can you please help me when this can happen and how should I go to debug it? This happens once in a while (Lets say once in 25-30 app sessions)

---
- Version:   Couchbase lite 1.3.1
- Client OS: iOS 10
- Server:    Sync Gateway 1.3.1

pasin:
File protection was set only when first creating the manager directory. Need to ensure that the file protection level is set (if specified) even when the manager directory already exists.

#1491

### DIFF
--- a/Source/API/CBLManager.m
+++ b/Source/API/CBLManager.m
@@ -241,7 +241,10 @@ static CBLManager* sInstance;
                 NSArray* paths = [[fmgr subpathsAtPath: _dir] arrayByAddingObject: @"."];
                 for (NSString* path in paths) {
                     NSString* absPath = [_dir stringByAppendingPathComponent: path];
-                    if (![absPath hasSuffix:@"-shm"]) { // Not changing -shm file
+                    if (![absPath hasSuffix:@"-shm"]) {
+                        // Not changing -shm file as it has NSFileProtectionNone by default
+                        // regardless of its parent directory projection level. The -shm file
+                        // contains non-sensitive information.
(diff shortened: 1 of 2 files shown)


Source/API/CBLManager.m
```diff
@@ -217,22 +217,38 @@ - (instancetype) initWithDirectory: (NSString*)directory
...
+                if (![attributes[key] isEqual: prot]) {
+                    needChange = YES;
+                    break;
+                }
+            }
+            if (needChange) {
+                NSArray* paths = [[fmgr subpathsAtPath: _dir] arrayByAddingObject: @"."];
+                for (NSString* path in paths) {
+                    NSString* absPath = [_dir stringByAppendingPathComponent: path];
+                    if (![absPath hasSuffix:@"-shm"]) { // Not changing -shm file
```
snej: Could you add a comment describing why the `-shm` file is skipped?

pasin:
@snej I added the comment per your comment. Please review again. Thanks.

PR #1549, opened 2017-01-15
Initial CBLQuery implementation. (API not official)

snej:
Adapted from the experimental LCQuery.mm in LiteCore.

The public API is just a best guess / WIP since we don't have a 2.0 API spec for queries yet. Moreover, this implements the NSPredicate-based API, not the cross-platform 'fluent' one; we'll need to add that once we have a spec.


Objective-C/CBLDocument.h
```diff
@@ -29,6 +29,9 @@ NS_ASSUME_NONNULL_BEGIN
 /** Check whether the document exists? in the database or not. */
 - (BOOL) exists;
 
+/** Sequence number of the document in the database. */
+@property (readonly) uint64_t sequence;
```
pasin: In term of public API, what will users use document's sequence for?

pasin:
I'm merging the PR now. I have one question above related to the API spec.

PR #1552, opened 2017-01-17
Fixed #1551 Implement timeout in ChangeTracker for longpoll and websocket mode

### LINKED ISSUE #1551: ChangeTracker stops working due to half-close connection
This issue is reported from a CBSE ticket (3350) that a continuous pull replicator (either websocket or longpoll) stopped working without any error or updated status. The scenario that make this happens is when unplugging an ethernet cable from a router so that the SGW cannot be reached but the replicator is still able to connect to the router.

--------
* Version:   1.3.1
* Client OS: iOS 10.2
* Server:    1.3.1

pasin:
* Implement timeout in ChangeTracker only for longpoll and websocket mode.
* Timeout value is greater than the heartbeat value 10%.
* When timeout occurs, both socket and websocket ChangeTracker will be failed with NSURLErrorTimedOut and the errors will be handled by the puller replicator with the retry logic.

#1551

### DIFF
--- a/Source/ChangeTracker/CBLSocketChangeTracker.m
+++ b/Source/ChangeTracker/CBLSocketChangeTracker.m
@@ -36,7 +36,6 @@ UsingLogDomain(Sync);
 
 #define kReadLength 4096u
 
-#define _kCFStreamPropertyReadTimeout CFSTR("_kCFStreamPropertyReadTimeout")
 
 @implementation CBLSocketChangeTracker
 {
@@ -46,7 +45,6 @@ UsingLogDomain(Sync);
     bool _gotResponseHeaders;
     bool _readyToRead;
     NSString* _serverName;
-    NSTimer* timeoutTimer;
 }
 
 - (BOOL) start {
@@ -355,11 +353,6 @@ UsingLogDomain(Sync);
             LogTo(ChangeTracker, @"%@: Event %lx on %@", self, (long)eventCode, stream);
             break;
     }
-}
-
-
-- (void) handleTimeout {
-    [self failedWithErrorDomain: NSURLErrorDomain code: NSURLErrorTimedOut message: @"Timeout"];
 }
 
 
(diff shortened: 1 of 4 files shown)


Source/ChangeTracker/CBLSocketChangeTracker.m
```diff
@@ -350,4 +358,9 @@ - (void) stream: (NSStream*)stream handleEvent: (NSStreamEvent)eventCode {
 }
 
 
+- (void) handleTimeout {
+    [self failedWithErrorDomain: NSURLErrorDomain code: NSURLErrorTimedOut message: @"Timeout"];
```
snej: I suggest moving this up to the superclass since it's a reasonable default behavior and isn't specific to the CFStream implementation.

snej:
I did some research and found the `setsockopt` option `SO_RCVTIMEO`, which sets a timeout for read operations. I'm just not sure how CFStream will react if a read call ends due to a timeout; I'm not sure if it will recognize it as an error. Maybe you could try a quick experiment? If it works, this would be the best way to implement this, since it requires very little code and no special timers at our level.

pasin:
@snej I did try with SO_RCVTIMEO but I ran into the issue that I couldn't get the socket at the right event.

Basically I added the following code to the `stream:handleEvent:` delegate to get the socket when the stream is opened BUT sockWrapper is always NULL. Getting the socket right after the `[_trackingInput open]` call has the same issue.

```
       case NSStreamEventOpenCompleted: {
            CFSocketNativeHandle socket;
            CFDataRef sockWrapper = (CFDataRef)CFReadStreamCopyProperty((CFReadStreamRef)_trackingInput, kCFStreamPropertySocketNativeHandle);
            CFRange r = {0, sizeof(CFSocketNativeHandle)};
            CFDataGetBytes(sockWrapper, r, (UInt8 *)&socket);
            CFRelease(sockWrapper);
            struct timeval tv;
            tv.tv_sec = 10;
            setsockopt(socket, SOL_SOCKET, SO_RCVTIMEO, &tv, sizeof(tv));
            break;
        }
```

I also tried to set the SO_RCVTIMEO value in NSStreamEventHasBytesAvailable case which I could access the socket, but that's probably too late as I couldn't see any effect.

pasin:
From https://lists.apple.com/archives/macnetworkprog/2008/Aug/msg00071.html thread, it seems like we couldn't rely on getting the socket from the HTTP Stream level.

PR #1559, opened 2017-01-19
add [CBLQueryEnumerator currentRowIndex]

mglasgow:
This method exposes the index of the last row that was read by the enumerator. I have included a unit test to confirm that it works with iterator enumerators (such as ForestDB).

I've used this method in my own code base to write a category on CBLQueryEnumerator that allows peeking at the next row.


mglasgow:
I've been pondering the name (the hardest part of programming, IMHO). Perhaps lastRowIndex is more appropriate?

snej:
What's the need for this? Can't you just keep your own counter and increment it as you walk through the enumeration?

We're not making any more public API changes for 1.x; our focus is on the 2.0 API, which is entirely new. (See the `feature/2.0` branch for the current very-incomplete status.)

mglasgow:
I have a use case that needs to peek ahead at the next row to decide how to process the current row. At the time (before ForestDB), I decided that exposing CBLQueryEnumerator's _nextRowIndex and a category method which calls rowAtIndex would be the easiest way to implement it. I figured why keep my own count when internally CBLQueryEnumerator already does?

If you're not making any public changes, I could rework my peeking into a subclass.

Btw, I like the direction with couchbase lite core. I use mapbox-gl-native and their use of C++ for a common core implementation has allowed a more consistent implementation across multiple platforms.

snej:
This isn't something we're going to add to the public API.

PR #1563, opened 2017-01-20
Implement Subdocument feature

pasin:
* Any Dictionaries are mapped to CBLSubdocument.
* When document is saved, update all subdocuments with new roots or invalidate them if they do not valid anymore.

**Discussion Points:**
* Current the invalidated subdocuments are also removed from cache (_properties) so they are no longer getting any new updates.


Objective-C/CBLDatabase.mm
```diff
@@ -290,6 +291,13 @@ - (void) document: (CBLDocument*)doc hasUnsavedChanges: (bool)unsaved {
 }
 
 
+- (NSMapTable*) sharedStrings {
+    if (!_sharedStrings)
+        _sharedStrings = FLCreateSharedStringsTable();
```
snej: This NSMapTable is only to be used with a single Fleece document (root FLValue.) The docs and naming are misleading; I think I told you the wrong thing earlier because I thought you were talking about the per-database shared string table. Sorry!
Just take this out for now -- it's just an optimization. We can put it in later.

Objective-C/CBLProperties.mm
```diff
@@ -146,97 +145,403 @@ - (BOOL) containsObjectForKey: (NSString*)key {
...
 
 - (nullable id) objectForKeyedSubscript: (NSString*)key {
-    if (_properties)
-        return _properties[key];
-    else
-        return [self fleeceValueToObject: [self fleeceValueForKey: key]];
+    id obj = _properties ? _properties[key] : nil;
+    if (!obj) {
+        auto value = [self fleeceValueForKey: key];
+        obj = [self fleeceValueToObject: value  forKey: key asJson: NO];
```
snej: You can check for `value==nullptr` here and return nil, since it's a common case, and the other two calls are just no-ops.

Objective-C/CBLProperties.mm
```diff
@@ -146,97 +145,403 @@ - (BOOL) containsObjectForKey: (NSString*)key {
...
+
+
+- (void) cacheValue: (nullable id)value forKey: (NSString*)key changed: (BOOL)changed {
+    BOOL shouldCache = changed ||
+                       [value isKindOfClass: [CBLSubdocument class]] ||
+                       [value isKindOfClass: [NSArray class]];
+    if (shouldCache) {
+        if (!_properties)
+            _properties = [NSMutableDictionary dictionary];
+        _properties[key] = value;
```
snej: This will crash if value is nil.
An alternative is `[_properties setValue: value forKey: key]`, which allows a nil value.

Objective-C/CBLProperties.mm
```diff
@@ -146,97 +145,403 @@ - (BOOL) containsObjectForKey: (NSString*)key {
...
 - (void) setObject: (nullable id)value forKeyedSubscript: (NSString*)key {
-    [self mutateProperties];
-    
     // NSDate:
     if ([value isKindOfClass: NSDate.class])
         value = [CBLJSON JSONObjectWithDate: value];
     
     id oldValue = _properties[key];
     
     if (![value isEqual: oldValue] && value != oldValue) {
```
snej: Since `_properties` is a cache, this method will do the wrong thing if called with a nil `value`, in a document where the key has a value but it hasn't been accessed yet. `oldValue` will be nil, so the `if` block will be skipped and nothing will change.

I think you need to change `_properties[key]` to `self[key]`.

Objective-C/CBLProperties.mm
```diff
@@ -146,97 +145,403 @@ - (BOOL) containsObjectForKey: (NSString*)key {
...
+            subdoc = [self subdocumentWithRoot: nullptr withProperties: value forKey: key];
+        return subdoc;
+    } else if ([value isKindOfClass: [NSArray class]]) {
+        NSArray* oldArray = $castIf(NSArray, oldValue);
+        NSArray* nuArray = (NSArray*)value;
+        NSMutableArray* result = [NSMutableArray arrayWithCapacity: nuArray.count];
+        for (int i = 0; i < nuArray.count; i++) {
+            id oldValue = oldArray && i < oldArray.count ? oldArray[i] : nil;
+            id nuValue = nuArray[i];
+            result[i] = [self convertValue: nuValue oldValue: oldValue forKey: key];
```
snej: This assumes the old and new values at the same indexes correspond to each other. But that's not the case if items have been inserted or removed in the middle of the array. Or if the two arrays are entirely different.

Objective-C/CBLProperties.mm
```diff
@@ -146,97 +145,403 @@ - (BOOL) containsObjectForKey: (NSString*)key {
...
+                [oldSubdoc invalidate];
+            }
+        } else {
+            CBLSubdocument* oldSubdoc = $castIf(CBLSubdocument, oldValue);
+            [oldSubdoc invalidate];
+        }
+        return result;
+    } else {
+        CBLSubdocument* nuSubdoc = $castIf(CBLSubdocument, value);
+        [self attachSubdocument: nuSubdoc forKey: key];
```
snej: This will blow up if `value` isn't a `CBLSubDocument`, i.e. it's an NSString, NSNumber, NSNull. -attachSubdocument doesn't allow a nil first parameter.

Objective-C/CBLProperties.mm
```diff
@@ -146,97 +145,403 @@ - (BOOL) containsObjectForKey: (NSString*)key {
...
+        
+        if ([oldValue isKindOfClass: [CBLSubdocument class]])
+            [oldValue invalidate];
```
snej: What about if nuSubDoc == oldSubDoc?

Objective-C/CBLProperties.mm
```diff
@@ -146,97 +145,403 @@ - (BOOL) containsObjectForKey: (NSString*)key {
...
+            FLDictIterator_Begin(_root, &iter);
+            do {
+                NSString* key = [self fleeceValueToKeyString: FLDictIterator_GetKey(&iter)];
+                id value = _properties[key];
+                if (value) {
+                    if (asJson)
+                        value = [self jsonValueForValue: value];
+                    [changedNames removeObject: key];
+                } else {
+                    value = [self fleeceValueToObject: FLDictIterator_GetValue(&iter)
```
snej: This doesn't work for the case of a property that's been deleted (but not yet saved). In this case the property will have a value in `_root`, but not in `_properties`. (And the key will be in `_changedNames`.)
I think the right algorithm to use is
1. Set the result to a mutable copy of `_properties`
2. Iterate through `_root`.
   1. If the key is not in `_changedKeys`, and there's no value for the key in `_properties`, decode the Fleece value and add it to the result.

Objective-C/CBLProperties.mm
```diff
@@ -146,97 +145,403 @@ - (BOOL) containsObjectForKey: (NSString*)key {
...
+            id value = _properties[key];
+            result[key] = asJson ? [self jsonValueForValue: value] : value;
+        }
+    }
+    return result;
+}
+
+
+- (id) jsonValueForValue: (id)value {
+    if ([value conformsToProtocol: @protocol(CBLJSONEncoding)]) {
```
snej: NSDate?

Objective-C/CBLProperties.mm
```diff
@@ -146,97 +145,403 @@ - (BOOL) containsObjectForKey: (NSString*)key {
...
 
-- (void) mutateProperties {
-    if (!_properties)
-        _properties = [self.savedProperties mutableCopy];
+- (void) markChangedKeys: (NSArray*)keys {
+    if (!_changedNames)
+        _changedNames = [NSMutableSet set];
+    
+    for (NSString* key in keys) {
+        if (_properties[key] != nil && ![_changedNames containsObject: key])
```
snej: You don't need to call -containsObject:. -addObject: will do the right thing.

Objective-C/CBLProperties.mm
```diff
@@ -22,85 +22,84 @@
...
 
 
 - (BOOL) booleanForKey: (NSString*)key {
-    if (_properties || _rootProps)
-        return [numberProperty(_properties ? _properties : _rootProps, key) boolValue];
-    else
-        return FLValue_AsBool([self fleeceValueForKey: key]);
-    return NO;
+    id v = _properties ? _properties[key] : nil;
+    return v ? [numberProperty(v) boolValue] : FLValue_AsBool([self fleeceValueForKey: key]);
```
pasin: ok, I will make it consistent with the lite-core or fleece.

Objective-C/CBLDocument.mm
```diff
@@ -106,7 +106,7 @@ - (BOOL) purge: (NSError**)outError {
 
 
 - (void) revert {
-    [self resetChanges];
+    [self revertChanges];
```
pasin: I will make the change.

Objective-C/CBLProperties.mm
```diff
@@ -22,85 +22,84 @@
...
-        return slice2string((FLValue_AsString([self fleeceValueForKey: key])));
+    id v = _properties ? _properties[key] : nil;
+    return v ? $castIf(NSString, v) : slice2string((FLValue_AsString([self fleeceValueForKey: key])));
+}
+
+
+- (CBLSubdocument*) subdocumentForKey: (NSString*)key {
+    id subdoc = self[key];
+    if (!subdoc) {
+        subdoc = [self subdocumentWithRoot: nullptr withProperties: nil forKey: key];
```
pasin: By the API design, -subdocumentForKey: has the same behavior as Database's -documentWithID:, which will do get-or-create. We could discuss about next week. I will setup a meeting.

Objective-C/CBLProperties.mm
pasin: I don't have a good solution to match subdocuments in two arrays. From above, it just tries to match as possible. This is another thing that I want to discuss next week.

Objective-C/CBLProperties.mm
pasin: Yep.

Objective-C/CBLProperties.mm
pasin: Yep, I missed that case.

Objective-C/CBLProperties.mm
pasin: In most case, when setting a NSDate object, it will get converted to string. BUT there is one case that the NSDate objects will not get converted are when they are in an array.

Objective-C/CBLProperties.mm
pasin: When I tested it, it didn't crash. It just removed the value of the key.

Objective-C/CBLProperties.h
```diff
@@ -44,6 +44,9 @@ NS_ASSUME_NONNULL_BEGIN
 /** Get an NSString property by key. Returns nil if the property doesn't exists. */
 - (nullable NSString*) stringForKey: (NSString*)key;
 
+/** Get or create a new CBLSubdocument by key. */
```
pasin: Will discuss about this next week.

Objective-C/CBLProperties.mm
```diff
@@ -146,97 +145,403 @@ - (BOOL) containsObjectForKey: (NSString*)key {
...
+                return [self subdocumentWithRoot: (FLDict)value withProperties: nil forKey:key];
+            }
+            case kFLArray: {
+                FLArray array = FLValue_AsArray(value);
+                int count = FLArray_Count(array);
+                NSMutableArray* result = [[NSMutableArray alloc] initWithCapacity: count];
+                if (count > 0) {
+                    FLArrayIterator iter;
+                    FLArrayIterator_Begin(array, &iter);
+                    do {
```
pasin: OK

Objective-C/CBLProperties.mm
pasin: I thought the deleted case got taken care after iterating loop. I will revise the logic. Thanks.

Objective-C/CBLProperties.mm
pasin: Actually there is an if check if count > 0 then doing the loop, so I think I will keep the code the same as it looks cleaner than checking if FLArrayIterator_GetValue return nullprt.

pasin:
I fixed the code except the ones that are related to the API that we will discuss next week. I also rebased with the current feature/2.0 branch to get all of the new configuration.

Objective-C/CBLProperties.mm
snej: Hey, you're right — the API docs for [NSMutableDictionary](https://developer.apple.com/reference/foundation/nsmutabledictionary/1574187-setobject) say that the subscript operator allows a nil value, unlike -setObject:forKey:. I didn't know that!

pasin:
@snej I have made changes according to our conversation today (Commit 9e56c3dd004c5b9416ca6e81c0ad020fa210eeed). I also rebased with the current feature/2.0. Could you please re-review? Thanks.

Objective-C/CBLProperties.mm
```diff
@@ -22,85 +22,83 @@
...
 
 
 - (BOOL) booleanForKey: (NSString*)key {
-    if (_properties || _rootProps)
-        return [numberProperty(_properties ? _properties : _rootProps, key) boolValue];
-    else
+    id v = _properties[key];
+    if (v) {
+        id n = numberProperty(v);
+        return n ? [n boolValue] : YES;
```
snej: NSNull should also return NO

Objective-C/CBLProperties.mm
```diff
@@ -22,85 +22,83 @@
...
+    } else
         return FLValue_AsBool([self fleeceValueForKey: key]);
-    return NO;
 }
 
 
 - (nullable NSDate*) dateForKey: (NSString*)key {
     NSString* dateStr = [self stringForKey: key];
-    return [CBLJSON dateWithJSONObject: dateStr];
+    return dateStr ? [CBLJSON dateWithJSONObject: dateStr] : nil;
```
snej: You don't need to test for nil; +dateWithJSONObject already handles that.

Objective-C/CBLProperties.mm
```diff
@@ -22,85 +22,83 @@
...
 }
 
 
 - (nullable NSString*) stringForKey: (NSString*)key {
-    if (_properties || _rootProps)
-        return _properties ? _properties[key] : _rootProps[key];
-    else
-        return slice2string((FLValue_AsString([self fleeceValueForKey: key])));
+    id v = self[key];
+    return v ? $castIf(NSString, v) : nil;
```
snej: Don't need to test for nil; `$castIf` already handles that.

Objective-C/CBLProperties.mm
```diff
@@ -22,85 +22,83 @@
...
+}
+
+
+- (nullable CBLSubdocument*) subdocumentForKey: (NSString*)key {
+    id v = self[key];
+    return v ? $castIf(CBLSubdocument, v) : nil;
```
snej: Don't need to test for nil; `$castIf` already handles that.

Objective-C/CBLProperties.mm
```diff
@@ -146,101 +144,346 @@ - (BOOL) containsObjectForKey: (NSString*)key {
 
 
 - (nullable id) objectForKeyedSubscript: (NSString*)key {
-    if (_properties)
-        return _properties[key];
-    else
-        return [self fleeceValueToObject: [self fleeceValueForKey: key]];
+    id obj = _properties ? _properties[key] : nil;
```
snej: Don't need to check for nil.

Objective-C/CBLProperties.mm
```diff
@@ -146,101 +144,346 @@ - (BOOL) containsObjectForKey: (NSString*)key {
...
     }
 }
 
 
-#pragma mark - INTERNAL:
+- (void) revert {
+    for (NSString* key in _changedNames) {
+        id value = _properties[key];
+        if ([value isKindOfClass: [CBLSubdocument class]]) {
+            CBLSubdocument* subdoc = $cast(CBLSubdocument, value);
```
snej: `$cast` will do another isKindOf: test, which is redundant; just do a C cast instead. (Or do a `$castIf` and assign it to subdoc, then test if subdoc is nil; that's the pattern I usually use.)

Objective-C/CBLProperties.mm
```diff
@@ -22,85 +22,83 @@
...
+    return [self currentPropertiesAsJson: NO];
 }
 
 
 - (void) setProperties: (NSDictionary*)properties {
-    _properties = properties ? [properties mutableCopy] : [NSMutableDictionary dictionary];
-    [self markChanges];
+    _properties = [self convertProperties: properties];
+    _changedNames = nil;
+    [self markChangedKeys: [_properties allKeys]];
```
snej: I don't think this will do the right thing for keys that are in `_root` but are not in the new `properties`. They should be added to `_changedNames`, but they won't be.

Objective-C/CBLProperties.mm
```diff
@@ -146,101 +144,346 @@ - (BOOL) containsObjectForKey: (NSString*)key {
 
 
 - (nullable id) objectForKeyedSubscript: (NSString*)key {
-    if (_properties)
-        return _properties[key];
-    else
-        return [self fleeceValueToObject: [self fleeceValueForKey: key]];
+    id obj = _properties ? _properties[key] : nil;
+    if (!obj) {
```
snej: This method doesn't handle removed properties. If a property exists in `_root`, and then I set its value to `nil`, then this method _should_ return the new value of `nil`. But this code will see that there's no value in `_properties`, and fall back to reading the value from Fleece.
This line should also be checking whether `key` is in `_changedNames`, and if it is, skipping the following block. I.e. `if (!obj && ![_changedNames containsObject: key])`.
I don't know if you've looked at the property-handling code of CBLModel, but it does this.

Objective-C/CBLProperties.mm
pasin: I totally missed this. Thanks!

Objective-C/CBLProperties.mm
pasin: Again, I totally missed the remove case. Thanks!

pasin:
@snej I made changes per your comments and push. Thanks a lot for the review.

pasin:
I just fixed the conflicts and a broken integration with notification after the merge.

pasin:
As the feature/2.0 branch has been changed a lot, I'm taking the PR back and will resubmit after reimplementing.

PR #1566, opened 2017-01-21
Build configuration changes (warnings, optimizations)

snej:
* Added a bunch of warnings and static-analyzer checks, as used in LiteCore and CBL 1.x
* Added optimization settings to the Release.xcconfig
* Removed unused .xcconfig files
* Fixed some warnings triggered by the above


snej:
Oh, and I forgot to note that this switches to linking with the static library of LiteCore -- this is so we can do link-time optimization of the whole code base, which improves code generation. (It also means we won't have to deal with copying the LiteCore dylib into the framework bundle, which we hadn't done yet.)

PR #1567, opened 2017-01-23
Query: Implemented “returning:” and FTS

snej:
* Added “returning:” parameter to the full -createQueryWhere: method. This takes an array of NSExpressions denoting extra ‘columns’ of values to return for each row.
* Added a family of xxxAtIndex: methods to CBLQueryRow to get these returned values, as various types.
* The `expressions` array parameter to -createIndex: can now contain NSStrings, which will be parsed into NSExpressions.
* Added -[CBLTestCase loadJSONResource:] as a utility to read JSON from a resource file into the database.


snej:
I added two more commits enabling FTS. (They're dependent on the changes in the first commit on this branch, so it didn't make sense to create a separate PR.)

You may want to review each commit separately, since they do different things.

Objective-C/CBLDatabase.h
```diff
@@ -142,31 +143,42 @@ typedef struct {
...
 /** Creates a value index (type kCBLValueIndex) on a given document property.
     This will speed up queries that test that property, at the expense of making database writes a
     little bit slower.
-    @param expressions  Expressions to index, typically key-paths.
+    @param expressions  Expressions to index, typically key-paths. Can be NSExpression objects,
+                    or NSStrings that are expression format strings.
     @param error  If an error occurs, it will be stored here if this parameter is non-NULL.
     @return  True on success, false on failure. */
-- (bool) createIndexOn: (NSArray<NSExpression*>*)expressions
+- (bool) createIndexOn: (NSArray*)expressions
```
pasin: Should we use BOOL for consistency?

snej: Good point; is that being used in the rest of the public headers? If so I'll change this.

pasin: Yes. I think we always use BOOL in 1.0 API and Apple Objective-C Foundation mostly uses BOOL.

PR #1578, opened 2017-01-26
Add CouchbaseLite.podspec

pasin:
Create a podspec that builds the binary from source. Need to change the umbrella header file to include with flat structure.


pasin:
Branch name is incorrect. Will need to push a fix.

Issue #1590, opened 2017-01-27
2.0 - Read existing blob could return dictionary instead of Blob


pasin:
Reading an existing blob could return NSDictionary instead of CBLBlob. The scenario that makes this happen is here:

1. Open a database
2. Edit or add non-blob value to a document
3. Save
4. Access blob value -> NSDictionary not CBLBlob

```
- (void)failingTestReadExistingBlob {
    CBLDocument* doc = [self.db documentWithID: @"doc1"];
    NSData* content = [@"12345" dataUsingEncoding:NSUTF8StringEncoding];
    NSError* error;
    CBLBlob *data = [[CBLBlob alloc] initWithContentType:@"text/plain" data:content error:&error];
    Assert(data, @"Failed to create blob: %@", error);
    doc[@"data"] = data;
    doc[@"name"] = @"Jim";
    Assert([doc save: &error], @"Saving error: %@", error);
    
    doc = [[self.db copy] documentWithID: @"doc1"];
    Assert([doc[@"data"] isKindOfClass:[CBLBlob class]]);
    data = doc[@"data"];
    AssertEqualObjects(data.content, content);
    
    doc = [[self.db copy] documentWithID: @"doc1"];
    doc[@"foo"] = @"bar";
    Assert([doc save: &error], @"Saving error: %@", error);
    Assert([doc[@"data"] isKindOfClass:[CBLBlob class]]);
    data = doc[@"data"];
    AssertEqualObjects(data.content, content);
}
```

--------
* Version:   VERSION
* Client OS: NAME_AND_VERSION
* Server:    NAME_AND_VERSION

PR #1591, opened 2017-01-28
Fix converting fleece root to dictionary

pasin:
* Calling -fleeceValueToObject: to convert _root directly will end up always converting the whole _root into dictionary as checking dictionary tyle (_cbltype) will always be null at the _root level. Instead, constructing the dictionary by looping through each key/value at the root level and calling fleeceValueToObject() to convert each value.

* Removed unused variable.

#1590


snej:
FYI, another way to test loading docs from the database is to use the -reopenDB method I added to CBLTestCase yesterday, which closes and reopens `self.db`.

pasin:
Thanks. Will use that instead of copying the db.

PR #1618, opened 2017-02-10
Fixed #1616 : Add thread support features

### LINKED ISSUE #1616: 2.0: Add threading support
1. Add dispatchQueue option to CBLDatabaseOptions.
2. Ensure to post notification on the database thread or on the dispatchQueue if specified.
3. Add doAsync: and doSync: method to the public API.

pasin:
* Added dispatchQueue option to CBLDatabaseOptions.
* Added doAsync: and doAsync: to CBLDatabase. The implementation was copied from 1.x branch. I didn’t include the private runloop mode as we don’t  need it now. We can include it later when we need it.
* Ensure to post notification on the database thread or on the dispatch queue if specified in the database options.
* No assertion checking if users use CBL objects on the wrong thread. Will do this in a seperate task.

#1616


pasin:
As @borrrden has concerned about setting a dispatch queue that cannot guarantee to be a serial dispatch queue. I'm taking the PR out until the discussion is finalized.

PR #1636, opened 2017-02-23
Fixed #1624: Implement Swift Subdocument

### LINKED ISSUE #1624: 2.0: Implement Swift Subdocument
The initial implementation is in https://github.com/couchbase/couchbase-lite-ios/tree/feature/2.0-swift-subdoc branch. Need to add unit tests to complete.

--------
* Version:   2.0DB2

pasin:
* Cached both Swift Document and Subdocument inside CBLDocument and CBLSubdocument respectively.
* Implement Swift Subdocument by wrapping the CBLSubdocument.

#1624

### DIFF
--- a/Swift/Document.swift
+++ b/Swift/Document.swift
@@ -91,7 +91,26 @@ public class Document : Properties {
         database = inDatabase
         _docimpl = impl
         super.init(impl)
+        
+        _docimpl.swiftDocument = self
+    }
+    
+    deinit {
+        _docimpl.swiftDocument = nil
     }
 
     let _docimpl: CBLDocument
+}
+
+
+extension CBLDocument {
+    private struct AssociatedKeys {
+        static var SwiftDocument = "SwiftDocument"
+    }
+    
+    var swiftDocument: Document? {
+        get {return objc_getAssociatedObject(self, &AssociatedKeys.SwiftDocument) as? Document}
+        set { objc_setAssociatedObject(self, &AssociatedKeys.SwiftDocument, newValue,
+                                     objc_AssociationPolicy.OBJC_ASSOCIATION_ASSIGN)}
+    }
 }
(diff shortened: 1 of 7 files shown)


Swift/Document.swift
```diff
@@ -91,7 +91,26 @@ public class Document : Properties {
...
+
+
+extension CBLDocument {
+    private struct AssociatedKeys {
+        static var SwiftDocument = "SwiftDocument"
+    }
+    
+    var swiftDocument: Document? {
+        get {return objc_getAssociatedObject(self, &AssociatedKeys.SwiftDocument) as? Document}
+        set { objc_setAssociatedObject(self, &AssociatedKeys.SwiftDocument, newValue,
```
snej: It would be more efficient to add a private property to CBLDocument for this; we own the class so we don't need to use associatedObject (which is kind of slow.)

I'm a little worried that because the reference is weak, the Document object is likely to get dealloced very quickly unless something on the Swift side is holding onto it. But it's not a correctness problem, just (possibly) performance.

Swift/Subdocument.swift
```diff
@@ -0,0 +1,46 @@
...
+    let _subdocimpl: CBLSubdocument
+}
+
+extension CBLSubdocument {
+    private struct AssociatedKeys {
+        static var SwiftSubdocument = "SwiftSubdocument"
+    }
+    
+    var swiftSubdocument: Subdocument? {
+        get {return objc_getAssociatedObject(self, &AssociatedKeys.SwiftSubdocument) as? Subdocument}
```
snej: Same comment as for Document ... we can just add a private property to CBLSubDocument.

pasin:
> can you change the associatedObjects to properties soon?

Yes, I will do it soon.

Swift/Document.swift
pasin: > I'm a little worried that because the reference is weak, the Document object is likely to get dealloced very quickly unless something on the Swift side is holding onto it. But it's not a correctness problem, just (possibly) performance.

Maybe we could do mutable cache in Swift as well. I will create a ticket to track this for improvement.

PR #1657, opened 2017-03-09
Fixed #1649 : pusher stopped when getting a forbidden error

### LINKED ISSUE #1649: missing role appears to cause continuous push replication to stop [v1.4]
If I client attempts to edit a document for which they are not permission using roles i see the "missing role" error message in the sync gateway and iOS logs as you would expect.  What appears to be happening after is that no more changes that the client makes are pushed to the sync gateway.  It's as if the pull replication process stops completely.  Is this intended behaviour?

Here are the logs: https://gist.github.com/npomfret/359800d530ef07b539fdbf52c9acdbbb

pasin:
Not treating 403 or forbidden error as an error when uploading multipart or non-multipart revision. This is inline with uploading bulk docs.

#1649

### DIFF
--- a/Source/CBLRestPusher.m
+++ b/Source/CBLRestPusher.m
@@ -560,7 +560,10 @@ CBLStatus CBLStatusFromBulkDocsResponseItem(NSDictionary* item) {
                           _dontSendMultipart = YES;
                           [self uploadJSONRevision: rev];
                       } else {
-                          self.error = error;
+                          // 403/Forbidden means validation failed; don't treat it as an error
+                          // because I did my job in sending the revision.
+                          if (error.code != kCBLStatusForbidden)
+                              self.error = error;
                           [self revisionFailed];
                       }
                   } else {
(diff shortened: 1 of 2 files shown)


npomfret:
Is there any indication on the document that the push has failed?

pasin:
There is no API to indicate which documents has been failed from the push replication. You may check the push replicator's last error so see if there was an error or not.

Source/CBLRestPusher.m
```diff
@@ -560,7 +560,10 @@ - (BOOL) uploadMultipartRevision: (CBL_Revision*)rev {
                           _dontSendMultipart = YES;
                           [self uploadJSONRevision: rev];
                       } else {
-                          self.error = error;
+                          // 403/Forbidden means validation failed; don't treat it as an error
+                          // because I did my job in sending the revision.
+                          if (error.code != kCBLStatusForbidden)
```
snej: Should this follow the same logic as the equivalent code in -uploadBulkDocs:? That code also treats `kCBLStatusUnauthorized` the same way. And for both of them it does not call `[self revisionFailed]`, because that would result in retrying the document next time.

Source/CBLRestPusher.m
```diff
@@ -603,7 +606,10 @@ - (void) uploadJSONRevision: (CBL_Revision*)originalRev {
                       body: rev.properties
               onCompletion: ^(id response, NSError *error) {
                   if (error) {
-                      self.error = error;
+                      // 403/Forbidden means validation failed; don't treat it as an error
+                      // because I did my job in sending the revision.
+                      if (error.code != kCBLStatusForbidden)
```
snej: Same comment as above. Since this logic is in 3 places now, maybe it should be factored out into a method?

Source/CBLRestPusher.m
pasin: Both uploadMultipartRevision: and uploadJSONRevision: are check the error from the request directly which are different from uploadBulkDocs:changes: that checks the error from the result. I'm not sure if we should treat kCBLStatusUnauthorized as not an replication error.

pasin:
@snej I have push the change. Please review. Thanks.

PR #1658, opened 2017-03-09
Changing the protocol scheme from git to https for better support

thertzelle:
carthage update on any sub-module of couchbase-lite-ios for iOS Development causes a timeout when checking out all dependancies.

Many companies restrict using the git:// scheme due to it using a specific port and not using proper authorization. Https is much more secure and more commonly unblocked.

Proposing this file be standardized to use https rather than git protocol.

This fixes the carthage update timeout issue.

Alternatively users cloning this project will need to run a command similar to:

git config --global url."https://".insteadOf git://

or remove --global and do it on a project-to-project basis.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,24 +1,24 @@
 [submodule "vendor/CocoaHTTPServer"]
 	path = vendor/CocoaHTTPServer
-	url = git://github.com/couchbaselabs/CocoaHTTPServer.git
+	url = https://github.com/couchbaselabs/CocoaHTTPServer.git
 [submodule "vendor/MYUtilities"]
 	path = vendor/MYUtilities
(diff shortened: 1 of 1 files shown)


pasin:
Can you explain a bit more about git:// is causing the timeout?

thertzelle:
Most schools, businesses behind a decent firewall will block arbitrary ports. In this situation, carthage will go out and pull a submodule directly with git, using the git protocol, which maps to port 9418. Most large enterprises block this port. https (443) is a more common, standardized port for submodules since its for the most part, available for all clients.

I submitted a similar pull request to Facebook-SDK-IOS a few months ago, they had the same problems, and accepted the change without any issues.

pasin:
@thertzelle Thanks for the clarification.

PR #1673, opened 2017-03-17
Fixed #1655 : Race causing the change tracker to start twice

### LINKED ISSUE #1655: Fatal Exception: NSInternalInconsistencyException [1.4]
http://crashes.to/s/9f71ca0e1a6

pasin:
The issue couldn’t reproduce but here is what is speculated to happen:

There could be a race scenario between stopping the fetch remote checkpoint request when putting the replicator offine and the request itself is completed. If stopping the request is too late, the onComplete will be called in the next runloop, and that will cause a change tracker to start in offline mode.

To prevent above scenario, when the request is completed, check whether the replicator is still online or not. If the replicator is not online, just finish the request and return.

#1655


pasin:
We have a customer (CBSE) that is experiencing the same issue. @snej could you please review this issue so that they could try to see if the PR could fix the issue or not.?

PR #1738, opened 2017-04-28
Prevent bad optimization of SQLite queries for view indexing & changes

### LINKED ISSUE #1737: View indexing is very slow, due to mis-optimized SQLite query
(CBSE-3690) View indexing has gotten drastically slower, at least in some use cases, in iOS 10 due to SQLite's mis-optimization of the main query used to find the new revisions to index — it takes a very long time (~500ms) and generates huge numbers of reads. 

Using `EXPLAIN QUERY PLAN` I found that the optimizer has generated a horrifically bad plan, essentially an inside-out join, that results in linear scans of two large tables. (Presumably this is due to some change in the optimizer between the older and newer versions of SQLite in iOS.)

The query, with some extraneous details removed, looks like:
```sql
	SELECT docs.doc_id, revs.sequence, docs.docid, rev.revid, rev.json, […] FROM revs, docs 
	WHERE revs.sequence>? AND revs.current!=0 AND revs.doc_id = docs.doc_id 
	ORDER BY revs.doc_id, revs.deleted, revs.revid DESC
```
Basically this query is meant to find the latest revs, including their “docid” strings joined from the “docs” table. It should just be using the “revs” table’s primary index starting from the “?” parameter, then looking up the corresponding doc for each revision. Instead it does this:
```
selectid = 0
  order = 0
   from = 1
 detail = SCAN TABLE docs USING COVERING INDEX docs_docid

selectid = 0
  order = 1
   from = 0
 detail = SEARCH TABLE revs USING INDEX revs_current (doc_id=?)

selectid = 0
  order = 0
   from = 0
 detail = USE TEMP B-TREE FOR ORDER BY
```
It’s scanning through all 50,000+ rows in “docs”, and then for each document looking up all the rows in “revs” that reference it (“revs” has 175,000 rows), and comparing their sequences to the parameter.

snej:
I’ve also added explicit JOIN syntax in a few places, since it’s said
to give the optimizer more insight into the meaning of the query.

I also removed several unnecessary indexes, where SQLite already
indexes the column in question to implement a UNIQUE constraint. This
was recommended by SQLite gurus to avoid confusing the optimizer.

Fixes #1737 

NOTE: If you're not familiar with it, the unary "+" operator in SQLite tells the query optimizer to avoid using an index on the referenced column. It's otherwise a no-op.

### DIFF
--- a/Source/CBL_SQLiteStorage.m
+++ b/Source/CBL_SQLiteStorage.m
@@ -447,7 +447,6 @@ DefineLogDomain(SQL);
             CREATE TABLE docs (\
                 doc_id INTEGER PRIMARY KEY,\
                 docid TEXT UNIQUE NOT NULL);\
-            CREATE INDEX docs_docid ON docs(docid);\
             \
             CREATE TABLE revs (\
                 sequence INTEGER PRIMARY KEY AUTOINCREMENT,\
@@ -467,7 +466,6 @@ DefineLogDomain(SQL);
                 docid TEXT UNIQUE NOT NULL,\
                 revid TEXT NOT NULL COLLATE REVID,\
                 json BLOB);\
-            CREATE INDEX localdocs_by_docid ON localdocs(docid);\
             \
             CREATE TABLE views (\
                 view_id INTEGER PRIMARY KEY,\
@@ -475,7 +473,6 @@ DefineLogDomain(SQL);
                 version TEXT,\
                 lastsequence INTEGER DEFAULT 0,\
                 total_docs INTEGER DEFAULT -1);\
-            CREATE INDEX views_by_name ON views(name);\
             \
             CREATE TABLE info (\
                 key TEXT PRIMARY KEY,\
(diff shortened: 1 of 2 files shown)


snej:
SQLite mailing list thread where people gave me advice starts [here](https://www.mail-archive.com/sqlite-users@mailinglists.sqlite.org/msg103412.html).
There's another thread about it [here](https://www.mail-archive.com/sqlite-users@mailinglists.sqlite.org/msg103416.html).

Source/CBL_SQLiteStorage.m
```diff
@@ -447,7 +447,6 @@ - (BOOL) open: (NSError**)outError {
             CREATE TABLE docs (\
                 doc_id INTEGER PRIMARY KEY,\
                 docid TEXT UNIQUE NOT NULL);\
-            CREATE INDEX docs_docid ON docs(docid);\
```
pasin: Do we need to also execute a SQL statement to delete the index?

PR #1742, opened 2017-05-04
Implement new objects

pasin:
The following items haven't been done:
* CBLArray optimization (Now CBLArray loads all fleece items in memory.)
* Conflict Resolution optimization
* Dictionary and Array should implement Java `Iterable` equivalent interface. I understand that the equivalent one that Objective-C should implement is `NSFastEnumeration` protocol but I would like to do a little more research if it's correct or not.
* Still have Subdocument (From the recent API feedbacks's review, we decided to use Dictionary instead of Subdocument).
* Fragment refactoring as refactoring needs to make some changes in object hierarchy in spec. 
* No Swift Change yet.
* There are some unit tests marked as failing test (mostly related to operations after closing or deleting the database). This will get taken care later.

Pending questions from @hideki's review:
* Do FLValue_AsInt, FLValue_AsFloat, and FLValue_AsDouble always return the same result as NSNumber's -IntegerValue, floatValue, and doubleValue or not (especially float and double value).
* Why boolean value has a special logic to return YES if the object which is not NSNumber is NOT null?


Objective-C/CBLReadOnlyDictionary.mm
```diff
@@ -0,0 +1,194 @@
...
+- (BOOL) containsObjectForKey: (NSString*)key {
+    FLValueType type = FLValue_GetType([self fleeceValueForKey: key]);
+    return type != kFLUndefined;
+}
+
+
+- (NSArray*) allKeys {
+    NSMutableArray* keys = [NSMutableArray array];
+    if (_dict != nullptr) {
+        FLDictIterator iter;
```
hideki: `iter` should be released in general.

I believe FLDictIterator automatically release itself when it reaches at the end.

Objective-C/CBLReadOnlyDictionary.mm
```diff
@@ -0,0 +1,194 @@
...
+             database: (CBLDatabase*)database
+                error: (NSError**)outError
+{
+    NSArray* keys = [self allKeys];
+    FLEncoder_BeginDict(encoder, keys.count);
+    for (NSString* key in keys) {
+        CBLStringBytes bKey(key);
+        FLEncoder_WriteKey(encoder, bKey);
+        id value = [self objectForKey: key];
+        if ([self isFleeceEncodableValue: value]) {
```
hideki: In case this returns `false`, Key is already written. This cause key&value mismatching.

Objective-C/CBLReadOnlyFragment.h
```diff
@@ -0,0 +1,91 @@
...
+NS_ASSUME_NONNULL_BEGIN
+
+/** CBLReadOnlyFragment protocol provides readonly access to the data value wrapped by
+    a fragment object. */
+@protocol CBLReadOnlyFragment <NSObject>
+
+/** Gets the value as an integer.
+    Floating point values will be rounded. The value `true` is returned as 1, `false` as 0.
+    Returns 0 if the value is nil or is not a numeric value. */
+@property (nonatomic, readonly) NSInteger integerValue;
```
hideki: why not `int` instead of NSInteger?

Objective-C/Internal/CBLData.h
```diff
@@ -0,0 +1,29 @@
...
+@class CBLDatabase;
+@class CBLC4Document;
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface CBLData : NSObject
+
++ (BOOL) validateValue: (id)value;
+
++ (BOOL) booleanValueForObject: (id)object;
```
hideki: why does boolean have a special method?

Objective-C/Internal/CBLDictionaryObject.m
```diff
@@ -0,0 +1,402 @@
...
+    }
+}
+
+
+- (NSInteger) integerForKey: (NSString*)key {
+    id value = _dict[key];
+    if (!value)
+        return [super integerForKey: key];
+    else
+        return [$castIf(NSNumber, value) integerValue];
```
hideki: Some codes use NSNumber integerValue, floatValue. Some codes use FLValue.asInt(), asFloat()....
I am not sure if they are exactly same implementation. 
I recommend using one of them.

Objective-C/CBLArray.h
```diff
@@ -0,0 +1,82 @@
...
+
+#import <Foundation/Foundation.h>
+#import "CBLReadOnlyArray.h"
+#import "CBLArrayFragment.h"
+@class CBLSubdocument;
+@class CBLArray;
+
+NS_ASSUME_NONNULL_BEGIN
+
+/** CBLDictionary protocol defines a set of methods for getting and setting array data. */
```
pasin: Thanks!

Objective-C/CBLArray.h
```diff
@@ -0,0 +1,82 @@
...
+
+/** Adds an object to the end of the array. Adding a nil value is equivalent
+    to adding an NSNull object. 
+    @param object   the object. */
+- (void) addObject: (nullable id)object;
+
+/** Inserts an object at the given index. Inserting a nil value is equivalent to inserting 
+    an NSNull object. 
+    @param object   the object.
+    @param index    the index. This value must not exceed the bounds of the array. */
```
pasin: It will throw a runtime exception like IndexOutOfBoundException. I don't remember the exact name.

Objective-C/CBLArray.h
```diff
@@ -0,0 +1,82 @@
...
+
+/** Gets a CBLSubdocument at the given index. Returns nil if the value is not a dictionary. 
+    @param index    the index. This value must not exceed the bounds of the array. */
+- (nullable CBLSubdocument*) subdocumentAtIndex: (NSUInteger)index;
+
+/** Gets a CBLArray at the given index. Returns nil if the value is not an array. 
+    @param index    the index. This value must not exceed the bounds of the array. */
+- (nullable CBLArray*) arrayAtIndex: (NSUInteger)index;
+
+/** Set an array as a content. Allowed value types are NSArray, NSDate, NSDictionary, NSNumber,
```
pasin: Thanks. Will do.

Objective-C/CBLArray.m
```diff
@@ -0,0 +1,342 @@
...
+
+
+- (nullable id) objectAtIndex: (NSUInteger)index {
+    return _array[index];
+}
+
+
+- (BOOL) booleanAtIndex: (NSUInteger)index {
+    id value = _array[index];
+    return [CBLData booleanValueForObject: value];
```
pasin: Current boolean has a special logic to convert the value. If the value is not NSNumber and NSNull, it will return true. I this the same LiteCore boolean conversion behavior. I will check with Jens the reason behind this logic. I could guess that this is inline with if (obj) in iOS.

Objective-C/CBLArray.m
```diff
@@ -0,0 +1,342 @@
...
+
+- (void) objectDidChange: (id)object {
+    [self setChanged];
+}
+
+
+#pragma mark - PRIVATE
+
+
+- (void) loadBackingData {
```
pasin: It loading all data from fleece into an (mutable) array for edit. Probably I should change the method name.

Objective-C/CBLDatabase.mm
```diff
@@ -259,18 +248,13 @@ - (BOOL) inBatch: (NSError**)outError do: (void (^)())block {
 }
 
 
-- (CBLDocument*) document {
-    return [self documentWithID: [self generateDocID]];
+- (CBLDocument*) documentWithID: (NSString*)documentID {
```
pasin: It's correct to use mustExist = YES. 

How about new doc with given ID.
> It will return null. The new API spec doesn't do get or create anymore. This method is only get.

Objective-C/CBLDictionary.h
```diff
@@ -0,0 +1,60 @@
...
+    object will be converted to an ISO-8601 format string. 
+    @param value    the object value.
+    @param key      the key. */
+- (void) setObject: (nullable id)value forKey: (NSString*)key;
+
+/** Set a dictionary as a content. Allowed value types are NSArray, NSDate, NSDictionary, NSNumber, 
+    NSNull, NSString, CBLArray, CBLBlob, CBLSubdocument. The NSArrays and NSDictionaries must 
+    contain only the above types.
+    @param dictionary  the dictionary. */
+- (void) setDictionary: (NSDictionary<NSString*,id>*)dictionary;
```
pasin: This is different from setObjectForKey:. This is setting the content of the CBLDictionary with a platform dictionary (AKA JSON Dictionary).

Objective-C/CBLDictionary.m
```diff
@@ -0,0 +1,444 @@
...
+#import "CBLData.h"
+#import "CBLDocument+Internal.h"
+#import "CBLFragment.h"
+#import "CBLJSON.h"
+#import "CBLSubdocument.h"
+
+
+@implementation CBLDictionary {
+    NSMutableDictionary<NSString*, id>* _dict;
+    NSMapTable* _changeListeners;
```
pasin: This is internal implementation for tracking changes from chid to parent.

Objective-C/CBLDictionary.m
```diff
@@ -0,0 +1,444 @@
...
+}
+
+
+@synthesize changed=_changed;
+
+
+static id kRemovedValue;
++ (void) initialize {
+    if (self == [CBLDictionary class]) {
+        kRemovedValue = [[NSObject alloc] init];
```
pasin: It's a placeholder value indicating the entry in the dictionary is removed.

Objective-C/CBLDocument.mm
```diff
@@ -109,107 +225,53 @@ - (BOOL) deleteDocument: (NSError**)outError {
...
-            
-            // Reload:
-            if (![self loadDoc_mustExist: NO error: outError])
-                return NO;
-            
-            self.properties = nil;
-            [self resetChangesKeys];
             
+            // Reset:
+            [self setC4Doc: nil];
```
pasin: When calling setC4Doc: method, the raw C4Document object will be wrapped by the CBLC4Document which will free C4Document object when it gets dealloc.

Objective-C/CBLDocument.mm
```diff
@@ -109,107 +225,53 @@ - (BOOL) deleteDocument: (NSError**)outError {
...
-
-
 // (Re)loads the document from the db, updating _c4doc and other state.
 - (BOOL) loadDoc_mustExist: (BOOL)mustExist error: (NSError**)outError {
     auto doc = [self readC4Doc_mustExist: mustExist error: outError];
     if (!doc)
         return NO;
-    [self setC4Doc: doc];
-    self.hasChanges = NO;
+    [self setC4Doc: [CBLC4Document document: doc]];
```
pasin: The c4doc is wrapped by CBLC4Document which will free the c4doc when the object is dealloc.

Objective-C/CBLDocument.mm
```diff
@@ -330,52 +397,61 @@ - (BOOL) mergeWithConflictResolver: (id<CBLConflictResolver>)resolver
...
-    NSDictionary* resolved;
+    
+    FLDict curRoot = nullptr;
+    auto curBody = rawDoc->selectedRev.body;
+    if (curBody.size > 0)
+        curRoot = (FLDict) FLValue_FromTrustedData({curBody.buf, curBody.size});
+    
+    // Create the current readonly document with the current revision:
+    CBLC4Document* curC4doc = [CBLC4Document document: rawDoc];
+    CBLFLDict* curDict = [[CBLFLDict alloc] initWithDict: curRoot
```
pasin: As far as there is no circular reference,  it should get released? Do you see anything that could prevent the release to happen?

Objective-C/CBLDocument.mm
```diff
@@ -419,15 +500,27 @@ - (BOOL) saveWithConflictResolver: (id<CBLConflictResolver>)resolver
...
     
-    [self postChangedNotificationExternal:NO];
     return YES;
 }
 
 
+#pragma mark - FLEECE ENCODING
+
+
+- (FLSliceResult) encodeWith: (FLEncoder)encoder error: (NSError**)outError {
```
pasin: I don't understand this comment so I will chat with you to get more info.

Objective-C/CBLFragment.m
```diff
@@ -0,0 +1,139 @@
...
+    if ([_value conformsToProtocol: @protocol(CBLArray)])
+        return [_value objectAtIndexedSubscript: index];
+    return [[CBLFragment alloc] initWithValue: nil parent: nil parentKey: nil];
+}
+
+
+#pragma mark - SET
+
+
+- (void) setValue: (NSObject*)value {
```
pasin: It just uses the parent to set the value at the key or index. We can chat about this.

Objective-C/CBLReadOnlyArray.h
```diff
@@ -0,0 +1,112 @@
...
+    @param index    the index.
+    @result the boolean value. */
+- (BOOL) booleanAtIndex: (NSUInteger)index;
+
+/** Gets value at the given index as an integer.
+    Floating point values will be rounded. The value `true` is returned as 1, `false` as 0.
+    Returns 0 if the property doesn't exist or does not have a numeric value.
+    @param index    the index.
+    @result the integer value. */
+- (NSInteger) integerAtIndex: (NSUInteger)index;
```
pasin: Objective-C preferred NSInteger because it can take care 32 / 64 bit. If using int, it means 32 bit. 64 bit int is int_64.

Objective-C/CBLReadOnlyArray.mm
```diff
@@ -0,0 +1,155 @@
...
+        _sharedKeys = _data.database.sharedKeys;
+    }
+    return self;
+}
+
+
+#pragma mark - GETTER
+
+
+- (nullable id) objectAtIndex: (NSUInteger)index {
```
pasin: This is an optimization so that fleece primitive data doesn't need to get converted to NSObject or NSNumber first. I saw Jens used it in Lite Core so I copied it. I will check with Jens on this.

Objective-C/CBLReadOnlyDictionary.mm
pasin: do you mean I need to free iter?

Objective-C/CBLReadOnlyDictionary.mm
pasin: Thanks to catch this.

Objective-C/CBLReadOnlyFragment.h
pasin: Same as explanation above.

Objective-C/Internal/CBLData.h
pasin: Same as explanation above.

Objective-C/Internal/CBLDictionaryObject.m
pasin: Same as explanation above. It's an optimization. Will get consult with Jens on this.

Objective-C/CBLArray.h
```diff
@@ -0,0 +1,82 @@
...
+    NSNull, NSString, CBLArray, CBLBlob, CBLSubdocument. The NSArrays and NSDictionaries must
+    contain only the above types.
+    @param array  the array. */
+- (void) setArray: (NSArray*)array;
+
+/** Subscripting access to a CBLFragment object that represents the value at the given index. 
+    @param index    the index. If the index value exceeds the bounds of the array, 
+                    the CBLFragment will represent a nil value.
+    @result the CBLFragment object. */
+- (CBLFragment*) objectAtIndexedSubscript: (NSUInteger)index;
```
pasin: This is objective-c subscription method.

snej:
How do you create a mutable copy of a read-only array/dictionary?
How do you convert a CBLArray or CBLDictionary into an NSArray or NSDictionary?

pasin:
> How do you create a mutable copy of a read-only array/dictionary?
Currently there is no API for that. I will create an API ticket for this.

> How do you convert a CBLArray or CBLDictionary into an NSArray or NSDictionary?
using CBLArray's toArray and CBLDictionary's toDictionary method.

snej:
Overall I have a very strong feeling that we should not be creating our own incompatible array/dictionary APIs. Any Cocoa programmer is going to look at these and go "WTF?" 

1. These require developers to learn new class and method names, adding to the learning curve.
2. The APIs are quite limited compared to NSArray and NSDictionary. There are dozens of useful Cocoa methods not available on them like `-objectsForKeys:`, `-keysOfEntriesPassingTest:`, `-subarrayWithRange:`, `-sortedArrayUsingComparator:`, etc. etc.
3. Interoperability with existing app code (or other libraries) that use Cocoa collections is going to be awkward, involving expensive conversions.

At this low level, having common classes between platforms isn't worth it, because they directly conflict with ubiquitous platform classes that the developer already has to use.

pasin:
@snej Can we take that comment to the API discussion and get the implementation of the current API spec merged in?

Objective-C/CBLDictionary.m
```diff
@@ -0,0 +1,444 @@
...
+    id value = _dict[key];
+    if (!value)
+        return [super numberForKey: key];
+    else
+        return $castIf(NSNumber, value);
+}
+
+
+- (nullable NSDate*) dateForKey: (NSString*)key {
+    return [CBLJSON dateWithJSONObject: [self stringForKey: key]];
```
snej: Use `objectForKey`.

Objective-C/CBLDictionary.m
```diff
@@ -0,0 +1,444 @@
...
+- (CBLFragment*) objectForKeyedSubscript: (NSString*)key {
+    id value = [self objectForKey: key];
+    return [[CBLFragment alloc] initWithValue: value parent: self parentKey: key];
+}
+
+
+#pragma mark - INTERNAL
+
+
+- (BOOL) isEmpty {
```
snej: The logic here is wrong. Consider where the superclass has an FLDict containing only the key 'x', and this object's `_dict` maps 'x' to 'kRemovedValue'. This method will incorrectly return `NO`.

Objective-C/CBLReadOnlyArray.mm
```diff
@@ -0,0 +1,155 @@
...
+
+
+#pragma mark - FLEECE ENCODABLE
+
+
+- (BOOL) fleeceEncode: (FLEncoder)encoder
+             database: (CBLDatabase*)database
+                error: (NSError**)outError
+{
+    NSUInteger count = self.count;
```
snej: All this method needs to do is call `FLEncoder_WriteValue(_array)`. The code here belongs in the subclass CBLArray.

Objective-C/CBLReadOnlyDictionary.mm
```diff
@@ -0,0 +1,194 @@
...
+- (BOOL) isFleeceEncodableValue: (id)value {
+    return YES;
+}
+
+
+- (BOOL) fleeceEncode: (FLEncoder)encoder
+             database: (CBLDatabase*)database
+                error: (NSError**)outError
+{
+    NSArray* keys = [self allKeys];
```
snej: All this method needs to do is call `FLEncoder_WriteValue(_dict)`. The code here belongs in the subclass CBLDictionary.

Objective-C/CBLArray.m
```diff
@@ -0,0 +1,343 @@
...
+
+#import "CBLArray.h"
+#import "CBLBlob.h"
+#import "CBLData.h"
+#import "CBLDocument+Internal.h"
+#import "CBLJSON.h"
+#import "CBLSubdocument.h"
+
+
+@implementation CBLArray {
```
pasin: I don't thing we have ever thought about the sibling relationship. I created https://github.com/couchbaselabs/couchbase-lite-apiv2/issues/84 for this.

Objective-C/CBLReadOnlyArray.h
```diff
@@ -0,0 +1,112 @@
...
+    Returns nil if the property doesn't exist, or its value is not a string.
+    @param index    the index.
+    @result the NSString object or nil. */
+- (nullable NSString*) stringAtIndex: (NSUInteger)index;
+
+/** Gets value at the given index as a number.
+    Returns nil if the property doesn't exist, or its value is not a number.
+    @param index    the index.
+    @result the NSNumber object or nil. */
+- (nullable NSNumber*) numberAtIndex: (NSUInteger)index;
```
pasin: I added the method for the sake of cross-platform consistency. I created I created https://github.com/couchbaselabs/couchbase-lite-apiv2/issues/85 to re-evaluate.

Objective-C/CBLDictionary.m
pasin: Thanks!

Objective-C/CBLDictionary.m
```diff
@@ -0,0 +1,444 @@
...
+#pragma mark - PRIVATE
+
+
+- (nullable id) prepareValue: (nullable id)value {
+    Assert([CBLData validateValue: value], @"Unsupported value type.");
+    return [self convertValue: value];
+}
+
+
+- (id) convertValue: (id)value {
```
pasin: Thanks! I like the code better after refactoring this.

Objective-C/CBLDictionary.m
```diff
@@ -0,0 +1,444 @@
...
+- (BOOL) isFleeceEncodableValue: (id)value {
+    return value != kRemovedValue;
+}
+
+
+#pragma mark - PRIVATE
+
+
+- (nullable id) prepareValue: (nullable id)value {
+    Assert([CBLData validateValue: value], @"Unsupported value type.");
```
pasin: Thanks!

Objective-C/CBLFragment.h
```diff
@@ -0,0 +1,44 @@
...
+NS_ASSUME_NONNULL_BEGIN
+
+/** CBLFragment protocol provides read and write access to the data value wrapped by 
+    a fragment object. */
+@protocol CBLFragment <CBLReadOnlyFragment>
+
+/** Gets or sets the value as an object. The object types are CBLBlob, CBLReadOnlyArray, 
+    CBLReadOnlyDictionary, NSNumber, or NSString based on the underlying data type; or nil 
+    if the value is nil. */
+@property (nonatomic, nullable) NSObject* value;
```
pasin: We have those in CBLReadOnlyFragment. The value property is mainly for setting as setting to object is weird.

Objective-C/CBLReadOnlyDictionary.mm
pasin: Thanks!

Objective-C/Internal/CBLFLArray.h
```diff
@@ -0,0 +1,31 @@
...
+//
+
+#import <Foundation/Foundation.h>
+#import "Fleece+CoreFoundation.h"
+@class CBLC4Document;
+@class CBLDatabase;
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface CBLFLArray : NSObject
```
pasin: I need more time to work on this as I don't fully understand about ARC on C++. This shouldn't change the way the other platform will be porting the code as well.

https://github.com/couchbase/couchbase-lite-ios/issues/1748

Objective-C/Internal/CBLFLDict.h
```diff
@@ -0,0 +1,30 @@
...
+//
+
+#import <Foundation/Foundation.h>
+#import "Fleece+CoreFoundation.h"
+@class CBLC4Document;
+@class CBLDatabase;
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface CBLFLDict : NSObject
```
pasin: I need more time to work on this as I don't fully understand about ARC on C++. This shouldn't change the way the other platform will be porting the code as well.

https://github.com/couchbase/couchbase-lite-ios/issues/1748

Objective-C/CBLArray.m
```diff
@@ -0,0 +1,343 @@
+//
+//  CBLArraym
```
pasin: Thanks

Objective-C/Internal/CBLFLDict.h
borrrden: I, for one, will not care if I have to port to C# from C++ or Objective-C.  C++ might even be easier because Objective-C collections have always annoyed me because they are non-generic.  Just be sure not to make use of C++ specific things like multiple inheritance from non-interface classes though (doubt that will be an issue here).

Objective-C/CBLReadOnlyArray.h
borrrden: Does Java have an equivalent class?  C# does not.  The reason that NSNumber exists is to represent differently sized numbers as one type but that reason is irrelevant for managed languages.  When porting this, I simply skipped this method.

pasin: It's good to know that C# doesn't have Number object.

pasin:
@snej I made the changes per code review. I created #1748 for completing the fix later.

pasin:
@snej Could you please re-review? Thanks.

pasin:
@snej I committed the fix for #1752 issue that @borrrden found.

pasin:
I squashed the recent changes into one for easily reviewing.

pasin:
@snej suggested that I should just merge it and we will make changes or fix issues in the further commits.

PR #1762, opened 2017-05-17
2.0: LiveQuery

### LINKED ISSUE #1759: 2.0: CBLLiveQuery (final API)
This is probably blocking some developers from being able to try 2.0, since it's a very widely used feature.

Note that we don't have a solid API spec for it yet.

snej:
Implementation of CBLLiveQuery, with provisional API. NOTE: Creating a LiveQuery is different. Instead of using a -asLiveQuery method to convert a query, you directly instantiate a CBLLiveQuery.

The implementation is adapted from the 1.x sources, but I simplified it to run the query on the main thread, since we don't have a mechanism to run things in the background yet.

The first commit is about creating a lazy NSArray class for the query's `rows` property, so it doesn't have to instantiate every row as Cocoa objects all at once.

Fixes #1759

### DIFF
--- a/Objective-C/CBLLiveQuery.mm
+++ b/Objective-C/CBLLiveQuery.mm
@@ -127,7 +127,6 @@ static const NSTimeInterval kDefaultLiveQueryUpdateInterval = 0.2;
     if (_willUpdate)
         return;  // Already a pending update scheduled
     _willUpdate = YES;
-    CBLLog(Query, @"%@: Will update after %g sec...", self, updateDelay);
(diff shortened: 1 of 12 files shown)


Objective-C/CBLLiveQuery.mm
```diff
@@ -0,0 +1,179 @@
...
+
+
+- (BOOL) waitForRows {
+    if (!_rows && !_lastError)
+        [self update];
+    return _rows != nil;
+}
+
+ 
+- (void) queryOptionsChanged {
```
pasin: This is currently not in the public header and has not ben used anywhere. There is only one option which is `updateInterval`; maybe we don't need to make it public and call this method when the the updateInterval is changed?

PR #1767, opened 2017-05-25
Optimize array by lazy copy data from fleece

pasin:
* Copied data from fleece when needed
* Bring array code closer to dictionary code

### DIFF
--- a/Objective-C/CBLArray.m
+++ b/Objective-C/CBLArray.m
@@ -42,89 +42,104 @@
 }
 
 
-- /*internal */ (instancetype) initWithFleeceData: (CBLFLArray*)data {
-    self = [super initWithFleeceData: data];
-    if (self ) {
-        _array = [NSMutableArray array];
-        [self loadBackingFleeceData];
-    }
-    return self;
-}
-
-
 #pragma mark - GETTER
 
 
 - (nullable id) objectAtIndex: (NSUInteger)index {
+    if (!_array) {
+        id value = [super objectAtIndex: index];
+        if ([value isKindOfClass: [CBLReadOnlyDictionary class]] ||
+            [value isKindOfClass: [CBLReadOnlyArray class]]) {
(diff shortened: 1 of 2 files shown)


pasin:
@hideki Per your code review comment, I assigned this PR to you for review.

Objective-C/CBLArray.m
```diff
@@ -42,89 +42,104 @@ - (instancetype) initWithArray: (NSArray*)array {
...
-
 #pragma mark - GETTER
 
 
 - (nullable id) objectAtIndex: (NSUInteger)index {
+    if (!_array) {
+        id value = [super objectAtIndex: index];
+        if ([value isKindOfClass: [CBLReadOnlyDictionary class]] ||
+            [value isKindOfClass: [CBLReadOnlyArray class]]) {
+            [self copyFleeceData];
```
hideki: I guess CBLDictionary/CBLArray could be instantiated with NSDictionary/NSArray from ReadOnlyDictionary/ReadOnlyArray. This could be cheaper than calling copyFleeceData.
However once loaded into `NSMutableArray`, following data access will be cheaper. What do you think?

Objective-C/CBLArray.m
```diff
@@ -42,89 +42,104 @@ - (instancetype) initWithArray: (NSArray*)array {
...
+    if (!_array)
+        return super.count;
+    else
+        return _array.count;
 }
 
 
 - (NSArray*) toArray {
+    if (!_array)
+        [self copyFleeceData];
```
hideki: Same question with above.

Objective-C/CBLArray.m
pasin: We need to copyFleeceData because we need to cache the converted Dictionary and Array so that it can return the same instance if users calling to get the object second time.

hideki: I see.

PR #1775, opened 2017-06-06
Make database thread safe

pasin:
* Created CBLLock that uses either NSLock or NSRecursiveLock.
* Applied lock to CBLDatabase where appropriate

### DIFF
--- a/Objective-C/CBLDictionary.mm
+++ b/Objective-C/CBLDictionary.mm
@@ -14,6 +14,7 @@
 #import "CBLDocument+Internal.h"
 #import "CBLFragment.h"
 #import "CBLJSON.h"
+#import "CBLLock.h"
 #import "CBLStringBytes.h"
 
 
@@ -22,6 +23,8 @@
     NSMapTable* _changeListeners;
     BOOL _changed;
     NSArray* _keys; // key cache
+    CBLLock* _lock; // Recursive lock
+    CBLLock* _changedLock;
 }
 
 
@@ -33,13 +36,23 @@
 }
 
 
+- /* internal */ (instancetype) initWithFleeceData: (nullable CBLFLDict*)data {
+    self = [super initWithFleeceData: data];
+    if (self) {
+        _lock = [[CBLLock alloc] initWithName: @"Dictionary" recursive: YES];
+        _changedLock = [[CBLLock alloc] initWithName: @"Dictionary-Changed"];self.data = data;
+    }
+    return self;
+}
+
+
 - (instancetype) init {
     return [self initWithFleeceData: nil];
 }
 
 
 - (instancetype) initWithDictionary: (NSDictionary<NSString*,id>*)dictionary {
-    self = [self initWithFleeceData: nil];
+    self = [self init];
     if (self) {
         [self setDictionary: dictionary];
     }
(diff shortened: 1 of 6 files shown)


pasin:
I have also include another commit that makes Document, Dictionary, and Array thread safe. Could you please review? Thanks.

Objective-C/Internal/CBLLock.m
```diff
@@ -0,0 +1,60 @@
...
+            _lock = [[NSRecursiveLock alloc] init];
+        else
+            _lock = [[NSLock alloc] init];
+    }
+    return self;
+}
+
+
+- (NSString*) description {
+    return [NSString stringWithFormat: @"%@[%@][resursive=%d]", self.class, _name, _recursive];
```
snej: typo ("resursive")

Objective-C/CBLDictionary.mm
```diff
@@ -33,13 +36,23 @@ + (instancetype) dictionary {
 }
 
 
+- /* internal */ (instancetype) initWithFleeceData: (nullable CBLFLDict*)data {
+    self = [super initWithFleeceData: data];
+    if (self) {
+        _lock = [[CBLLock alloc] initWithName: @"Dictionary" recursive: YES];
+        _changedLock = [[CBLLock alloc] initWithName: @"Dictionary-Changed"];self.data = data;
```
snej: Missing line-break before `self.data = data`

Objective-C/CBLDictionary.mm
```diff
@@ -321,27 +390,45 @@ - (NSArray*) allKeys {
 
 
 - (void) setChanged {
+    // Has it own _changedLock instead of using the _lock to avoid deadlock when encoding
+    // the object into fleece and at the same time there are some changes in a child
+    // dictionary or array by another thread. There is one side effect of doing this is that
+    // the changed status could be NO when one thread is trying to get the changed status
+    // but the other thread is modifying the object at the same time.
+    [_changedLock lock];
     if (!_changed) {
```
snej: This is technically not thread-safe, because the other methods above use a different lock to control access to `_changed`, so it's possible for two threads to read/write it at the same time by holding different locks.

Objective-C/CBLDictionary.mm
```diff
@@ -321,27 +390,45 @@ - (NSArray*) allKeys {
...
         _changed = YES;
         [self notifyChangeListeners];
```
snej: It's generally dangerous to call into code you don't control, while holding a lock.
In this case I think deadlocks are possible:
* From -setChanged, a change listener calls back into this object, which will then try to acquire `_lock`.
* Meanwhile, another thread is in a mutating method of this object that already has `_lock` and then calls into -setChanged to acquire `_changedLock`.

snej:
In general it would be better to use `@synchronized{...}`. It frees you from having to work with NSLock objects, it automatically unlocks at scope exit, and it's easier to read.
If you need a second lock, as in CBLDictionary, you can use `@synchronized(someOtherObject)` to effectively get a second lock. But be very careful about lock ordering.

snej:
Have we explicitly decided that we need this degree of thread-safety? Standard Cocoa collection classes are not this safe: in general read-only access is thread-safe, but mutations are not, so while one thread is mutating the collection it's illegal for another thread to access it. Most of the complexity in this PR seems to come from trying to provide that safety for Dictionary and Array classes.

pasin:
Thanks for review. I will make change to use @synchronized instead. I will dicuss with you next week about the degree of thread safety. I will talk to Dan about holding this feature for discussion on Monday.

PR #1790, opened 2017-06-19
Detect whether doc has any changes by comparing encoded binaries

pasin:
* Compare encoded binaries to detect whether the document has any changes or not. If no changes, save operation is no ops.
* Check self.changed (direct changes to the document) to bypass the comparison.


pasin:
Close. Due to some code changes, the PR is not valid anymore.

PR #1791, opened 2017-06-19
Make database thread safe

pasin:
* Applied @synchronized to make the database thread safe (serialize database operation).
* Note that in -postDatabaseChanged:, the same database lock is applied, this is not an ideal as calling to the application code via notification is not safe when the database lock is active (e.g. could cause deadlock or slowness). This needs to be revisited.

### DIFF
--- a/Objective-C/CBLDatabase.mm
+++ b/Objective-C/CBLDatabase.mm
@@ -96,6 +96,7 @@ static NSString* defaultDirectory() {
     CBLPredicateQuery* _allDocsQuery;
     NSMutableDictionary* _docChangeListeners;
     NSMutableDictionary* _docObs;
+    NSObject* _lock;
 }
 
 
@@ -149,6 +150,7 @@ static void docObserverCallback(C4DocumentObserver* obs, C4Slice docID, C4Sequen
                         error: (NSError**)outError {
     self = [super init];
     if (self) {
+        _lock = [[NSObject alloc] init];
         _name = name;
         _config = config != nil? [config copy] : [CBLDatabaseConfiguration new];
         if (![self open: outError])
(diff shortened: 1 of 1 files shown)


Objective-C/CBLDatabase.mm
```diff
@@ -178,12 +178,16 @@ - (NSString*) description {
 
 
 - (NSString*) path {
-    return _c4db != nullptr ? sliceResult2FilesystemPath(c4db_getPath(_c4db)) : nil;
+    @synchronized (self) {
```
hideki: Question: Could `@synchronized (self)` be the best approach instead of `@synchronized (non-self variable)`.  Same questions for all `@synchronized (self)`

pasin: I understand that it behaves the same. What is the benefit of using a mutex object instead of self?

Objective-C/CBLDatabase.mm
```diff
@@ -627,18 +658,18 @@ - (nullable CBLDocument*) documentWithID: (NSString*)documentID
                                mustExist: (bool)mustExist
                                    error: (NSError**)outError
 {
-    [self mustBeOpen];
-    
-    return [[CBLDocument alloc] initWithDatabase: self
-                                      documentID: documentID
-                                       mustExist: mustExist
-                                           error: outError];
+    @synchronized (self) {
```
pasin: Yes, it could make concurrent read slow. What is your suggestion on this?

Side note: I'm trying not to make complicated locks until we really have the real use cases that have performance issue. But I'm open for discussion.

Objective-C/CBLDatabase.mm
```diff
@@ -425,94 +446,104 @@ - (BOOL) createIndexOn: (NSArray*)expressions
...
 
 
 - (void) postDocumentChanged: (NSString*)documentID {
-    if (!_docObs[documentID] || !_c4db ||  c4db_isInTransaction(_c4db))
-        return;
+    NSSet* listeners;
+    @synchronized (self) {
+        if (!_docObs[documentID] || !_c4db)
+            return;
+        listeners = [_docChangeListeners objectForKey: documentID];
```
pasin: Thanks, I forgot to call -copy.

Objective-C/CBLDatabase.mm
pasin: I can think of one use case that developers also do synchronized (database). I think it might be better not to do with self. I will make changes.

hideki: Other classes can lock (@synchronized) on CBLDatabase instance.  This could cause the dead-lock or other threading problems.

In most of the cases, you would like to protect c4Database instance instead of CBLDatabase. This is main reason that I asked this question.

pasin:
I made two changes per code review. What is your suggestion on multiple read to get the documents?

Objective-C/CBLDatabase.mm
hideki: I don't have a good solution now. But current thread-safe logic does not allow concurrent reads too. So it is better to address this as soon as possible.

pasin:
Will submit a new PR.

PR #1792, opened 2017-06-19
Make document thread-safe

pasin:
* Assuming that CBLDatabase and CBLDictionary are thread safe.
* Apply @synchronized to make CBLDocument / CBLReadOnlyDocument thread safe.

### DIFF
--- a/Objective-C/CBLDocument.mm
+++ b/Objective-C/CBLDocument.mm
@@ -27,11 +27,14 @@
 #import "CBLStatus.h"
 
 
+@interface CBLDocument ()
+@property (readonly, nonatomic) CBLDictionary* dict;
+@end
+
 @implementation CBLDocument {
     C4Database* _c4db;      // nullable
     CBLDictionary* _dict;
 }
-
 
 @synthesize database=_database;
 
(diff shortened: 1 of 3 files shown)


Objective-C/CBLDocument.mm
```diff
@@ -198,80 +201,102 @@ - (NSUInteger)countByEnumeratingWithState: (NSFastEnumerationState *)state
...
-            [self setC4Doc: nil];
-            return YES;
+    @synchronized (self.lock) {
+        assert(_database && _c4db);
+        
+        if (!self.exists) {
+            return createError(kCBLStatusNotFound, outError);
+        }
+        
+        C4Transaction transaction(_c4db);
```
hideki: Does not require to lock `c4db`?

Objective-C/CBLDocument.mm
```diff
@@ -198,80 +201,102 @@ - (NSUInteger)countByEnumeratingWithState: (NSFastEnumerationState *)state
...
+                                        error: outError];
+    }
 }
 
 
 - (BOOL) deleteDocument: (NSError**)outError {
-    return [self saveWithConflictResolver: self.effectiveConflictResolver
-                                 deletion: YES
-                                    error: outError];
+    @synchronized (self.lock) {
```
hideki: Lock with `saveWithConflictResolver ` method is enough?? the same question applies to all method call `saveWithConflictResolver `.

Objective-C/CBLDocument.mm
pasin: c4db is locked from the CBLDatabase when calling CBLDatabase.purge() method.

Objective-C/CBLDocument.mm
pasin: There are only two and locking there seems to be easier to understand.

pasin:
Will submit a new PR.

PR #1793, opened 2017-06-20
Make CBLDictionary and CBLArray thread safe

pasin:


pasin:
Will submit a new PR.

PR #1794, opened 2017-06-21
Make Database, Document, Dictionary, Array thread safe

pasin:
* Created CBL_LOCK macro wrapping @syncronized(m) block.
* Applied CBL_LOCK to Database, Document, Dictionary, and Array to make them thread safe.
* NOTE: During the optimization we could consider not making Document and its related object not thread safe if locking process cause too much performance hits.

### DIFF
--- a/Objective-C/CBLArray.m
+++ b/Objective-C/CBLArray.m
@@ -18,6 +18,7 @@
     NSMutableArray* _array;
     NSMapTable* _changeListeners;
     BOOL _changed;
+    NSObject* _lock;
 }
 
 @synthesize swiftObject=_swiftObject;
@@ -25,6 +26,15 @@
 
 + (instancetype) array {
     return [[self alloc] init];
+}
+
+
+- /* internal */ (instancetype) initWithFleeceData: (nullable CBLFLArray*)data {
+    self = [super initWithFleeceData: data];
+    if (self) {
+        _lock = [[NSObject alloc] init];
+    }
+    return self;
 }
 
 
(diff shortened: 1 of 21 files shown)


Objective-C/CBLDocument.mm
```diff
@@ -198,80 +201,102 @@ - (NSUInteger)countByEnumeratingWithState: (NSFastEnumerationState *)state
...
-    return convertError(err, outError);
 }
 
 
 #pragma mark - PRIVATE
 
 
+- (CBLDictionary*) dict {
+    CBL_LOCK(self.lock) {
+        return _dict;
```
snej: This isn't safe, for the same reason as the -database method. Let the compiler synthesize this.

Objective-C/Internal/CBLDocument+Internal.h
```diff
@@ -60,6 +60,8 @@ NS_ASSUME_NONNULL_BEGIN
 
 @property (nonatomic, nullable) CBLC4Document* c4Doc;
```
snej: _All_ of the nonatomic properties need to be made atomic. In all the classes.

snej:
CBLBlob needs to be made thread-safe too.

Objective-C/CBLDocument.mm
```diff
@@ -198,80 +201,102 @@ - (NSUInteger)countByEnumeratingWithState: (NSFastEnumerationState *)state
...
-    return _dict.isEmpty;
+    return self.dict.isEmpty;
 }
 
 
 - (BOOL) save: (NSError**)outError {
-    return [self saveWithConflictResolver: self.effectiveConflictResolver
-                                 deletion: NO
-                                    error: outError];
+    CBL_LOCK(self.lock) {
```
pasin: I think the above sample does't cause the deadlock because only one thread can access the database at a time. The second thread doesn't hold the document's lock yet when waiting to access the databse to save the document.

I agree that it's dangerous to hold a lick while calling an unknown code. This is a difficult one to fix as
1. The database lock still needs to be hold until the save process is done.
2. The document's lock could be released during the conflict resolution but this could meanthat the document needs to be copied to send to the conflict handler. And any changes happen to the document from another thread while fixing conflicts will be ignored.

Could we discuss this during offline tomorrow?

Objective-C/CBLArray.m
```diff
@@ -18,6 +18,7 @@ @implementation CBLArray {
     NSMutableArray* _array;
     NSMapTable* _changeListeners;
     BOOL _changed;
+    NSObject* _lock;
```
pasin: Just to avoid the case that the app lock the object externally.

Objective-C/CBLDatabase.mm
```diff
@@ -425,94 +448,110 @@ - (BOOL) createIndexOn: (NSArray*)expressions
...
+        do {
+            // Read changes in batches of kMaxChanges:
+            bool newExternal;
+            nChanges = c4dbobs_getChanges(_obs, changes, kMaxChanges, &newExternal);
+            if (nChanges == 0 || external != newExternal || docIDs.count > 1000) {
+                if(docIDs.count > 0) {
+                    CBLDatabaseChange* change = [[CBLDatabaseChange alloc] initWithDocumentIDs: docIDs
+                                                                                    isExternal: external];
+                    NSDictionary *userInfo = @{ kCBLDatabaseChangesUserInfoKey: change };
+                    [[NSNotificationCenter defaultCenter] postNotificationName: kCBLDatabaseChangeNotification
```
pasin: I agree. Need to think how to avoid this when the notification happens in batch. Probably send the notification async?

Objective-C/CBLDocument.mm
```diff
@@ -198,80 +201,102 @@ - (NSUInteger)countByEnumeratingWithState: (NSFastEnumerationState *)state
...
+            return createError(kCBLStatusNotFound, outError);
+        }
+        
+        C4Transaction transaction(_c4db);
+        if (!transaction.begin())
+            return convertError(transaction.error(),  outError);
+        
+        C4Error err;
+        if (c4doc_purgeRevision(self.c4Doc.rawDoc, C4Slice(), &err) >= 0) {
+            if (c4doc_save(self.c4Doc.rawDoc, 0, &err)) {
```
snej: The `c4doc_save` call needs to be synchronized with the other C4 calls in CBLDatabase.m. That means the CBL_LOCK call needs to use the database's lock. Same is true with any other `c4_` call here that could access the database.

pasin: Currently -save:, deleteDocument:, and purge: method are only called from the database (for save, delete, and pure document API) which already locks the database. I should add comment about that here at least.

pasin:
@snej Could you please re-review the PR again? I also rebased the branch with the current feature/2.0 and did forced commit to make the PR clean. There are a couple of things to note:

1. I didn't make the macro for marking atomic property yet. To make that we need to expose thread-safe flag in the public header. I'm not sure if I should make the flag public or not.

2. I didn't do Thread-safe on Query and Replicator yet. That will be next after this PR gets merged.

Thanks.

pasin:
Rebase to fix conflicts.

pasin:
I rebased the code to fix the conflicts.

PR #1802, opened 2017-06-27
Resolve replication conflicts

### LINKED ISSUE #1799: 2.0: Immediately resolve pulled conflicting revisions
When the replicator pulls a revision that conflicts with a local one, the conflict resolver callback needs to run immediately, before that revision appears in the database, and the merged revision should be added instead.

This depends on new LiteCore replicator functionality, which is being worked on as https://github.com/couchbase/couchbase-lite-core/issues/164.

(This is arguably a bug, since the current behavior is that the conflicting revision semi-randomly either replaces the local one, or never shows up at all; in neither case does the app get a chance to deal with the conflict.)

snej:
CBL automatically detects when a replication creates a conflict (thanks to a new LiteCore callback), invokes the app's conflict resolver, and adds revision(s) to resolve the conflict.

It may be easiest to review the three commits individually.

Fixes #1799


Objective-C/CBLReadOnlyDocument.mm
```diff
@@ -109,4 +135,80 @@ - (BOOL) exists {
...
+                                                            error: outError];
+    if (!otherDoc || ![otherDoc selectConflictingRevision])
+        return false;
+
+    // Read the common ancestor revision (if it's available):
+    auto baseDoc = [[CBLReadOnlyDocument alloc] initWithDatabase: self.database
+                                                      documentID: self.documentID
+                                                       mustExist: YES
+                                                           error: outError];
+    if (![baseDoc selectCommonAncestorOfDoc: self andDoc: otherDoc] || !baseDoc.data)
```
pasin: This is a question. Why do we need to check baseDoc.data? Doesn't -selectCommonAncestorOfDoc return false if there is no base doc?

Objective-C/CBLDocument.mm
```diff
@@ -407,48 +357,34 @@ - (BOOL) mergeWithConflictResolver: (id<CBLConflictResolver>)resolver
...
     CBLReadOnlyDocument* resolved;
     if (deletion) {
         // Deletion always loses a conflict:
         resolved = current;
-    } else if (resolver) {
-        // Call the custom conflict resolver:
-        CBLReadOnlyDocument* base = [[CBLReadOnlyDocument alloc] initWithDocumentID: self.documentID
-                                                                              c4Doc: super.c4Doc
-                                                                         fleeceData: super.data];
+    } else {
```
pasin: Should we also add another if else to check whether current is deleted (same logic with replicator's conflict resolution)?
```
if (current.isDeleted) {
   resolved = self;
}
```

pasin: Actually this is not fully related to the PR so I could add this logic after merging the PR.

PR #1803, opened 2017-06-27
Swift example project

jamiltz:


jamiltz:
@pasin similarly to https://github.com/couchbase/couchbase-lite-ios/pull/1814 but for swift. Ready for review.

docs/examples/swift-api-walkthrough/api-walkthrough/ViewController.swift
```diff
@@ -0,0 +1,148 @@
...
+            .from(DataSource.database(database))
+            .where(
+                Expression.property("type").equalTo("user")
+                    .and(Expression.property("admin").equalTo(false))
+        )
+        
+        do {
+            let rows = try query.run()
+            for row in rows {
+                print("doc ID :: \(row.documentID)")
```
pasin: In DB13, query.run() returns ResultSet which is an iterator of Result. The Result doesn't have documentID property anymore. I'm wondering if the code is compiled?

To get the document ID, the projection is needed. See [here](https://github.com/couchbase/couchbase-lite-ios/blob/feature/2.0/Swift/Tests/QueryTest.swift#L485-L514) as an example to get the document ID.

docs/examples/swift-api-walkthrough/api-walkthrough/ViewController.swift
```diff
@@ -0,0 +1,148 @@
...
+            print(error.localizedDescription)
+        }
+        
+        let whereClause = Expression.property("name").match("'buy'")
+        let ftsQuery = Query.select().from(DataSource.database(database)).where(whereClause)
+        
+        do {
+            let ftsQueryResult = try ftsQuery.run()
+            for row in ftsQueryResult {
+                print("document properties \(row.document.toDictionary())")
```
pasin: We don't have .document property in DB13. The document ID projection is needed.

Note: We are still discussing about how to get the document from the query result.

jamiltz:
@pasin the project wasn't linking to the correct framework. Updated that and the code so it compiles against DB13. It does compile now but it needs your review (not sure [this](https://github.com/couchbase/couchbase-lite-ios/pull/1803/commits/ca536b2b2e4fd6c47810d1c68c9a9b36ab05afe8#diff-26b403fac0a74da51fb3ad7c3f6bd64d) is the right way to get the doc ID now).

jamiltz:
@pasin added a commit with the config I used with SG 1.5 477. When ran in Xcode the app is crashing but I don't know why (seems related to replication, when SG isn't running the app doesn't crash). Here's a screenshot but probably not very helpful https://cl.ly/3M0J3S1A1I3k/Screen%20Shot%202017-07-26%20at%2016.36.07.png

It would be great if you could try it locally on your machine.

PR #1805, opened 2017-06-28
NotifyChanges crash 

trr-amsiq:
Any way this can be tested?? #1739

### DIFF
--- a/Source/CBLDatabase+Internal.m
+++ b/Source/CBLDatabase+Internal.m
@@ -538,6 +538,7 @@ static BOOL sAutoCompact = YES;
 
 
 - (void) dbChanged: (NSNotification*)n {
+    [self doAsync: ^{
     CBLDatabase* senderDB = n.object;
     // Was this posted by a _different_ CBLDatabase instance on the same database as me?
     if (senderDB != self && [senderDB.dir isEqualToString: _dir]) {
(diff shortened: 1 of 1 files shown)


pasin:
Can you explain what this PR is for? I couldn't figure out the changes from looking at the PR because the PR contains a lot of indent changes (please revert those indent changes).

trr-amsiq:
It is for issue #1739.
The changes are within  `- (void) dbChanged: (NSNotification*)n` 
The `doAsync` call is wrapping the whole function so that the `echoedChanges` array is owned by the `doAsync` block. This may prevent it from getting garbage collected.

pasin:
Can you update the PR so I can take a look?

trr-amsiq:
Just did. 
I've reverted all the indent changes but those are still showing up.

pasin:
Thanks. I can see your change now. I'm not sure if the `echoedChanges` is the cause of the issue. I will try to work on this ASAP.

tristanrichard:
We created the issue on May 1th and had over 3000 crashes of that kind since. 
Any way we can get fix for any time soon? It is a very critical issue for us!

trr-amsiq:
We created the issue on May 1th and had over 3000 crashes of that kind since.
Any way we can get fix any time soon? It is a very critical issue for us!

pasin:
Do you have a symbolicated crash log with all threads and the exception name? I can you either update it here or at the original issue?

pasin:
Also what version of CBL iOS are you using?

trr-amsiq:
It was first seen on 1.3.1 and we are now running 1.4

I'll update this comment as soon as I can reproduce it in debug

pasin:
I understand that the issue is caused by `_changesToNotify` was released by another thread not that the `echoedChanges` was release. I'm going to close this PR. So please update the crash log reproduced with 1.4.0 binary on the original issue. I will try take a look right away once you updated the issue.

One thing that you should make sure in your application code is that you shouldn't share the database object across thread. Thanks.

PR #1814, opened 2017-07-10
objc-api-walkthrough project

jamiltz:
This project contains ObjC code snippets used in the docs. They are located in **docs/examples/objc-api-walkthrough/objc-api-walkthrough/ViewController.m**. Other files are required for the project to run in Xcode.


docs/examples/objc-api-walkthrough/objc-api-walkthrough/ViewController.m
```diff
@@ -0,0 +1,134 @@
...
+    
+    // mutate document
+    [newTask setObject:@"Apples" forKey:@"name"];
+    [database saveDocument:newTask error:&error];
+    
+    // typed accessors
+    [newTask setObject:[NSDate date] forKey:@"createdAt"];
+    NSDate* date = [newTask dateForKey:@"createdAt"];
+    
+    // database transaction
```
pasin: We should say that it's a batch operation not transaction even though we also guarantee the transaction in CBL's batch operation.

docs/examples/objc-api-walkthrough/objc-api-walkthrough/ViewController.m
```diff
@@ -0,0 +1,134 @@
...
+    [database saveDocument: newTask error:&error];
+    if (error) {
+        NSLog(@"Cannot save document %@", error);
+    }
+    
+    CBLBlob* taskBlob = [newTask blobForKey:@"avatar"];
+    UIImage* taskImage = [UIImage imageWithData:taskBlob.content];
+    
+    // query
+    CBLQuery* query = [CBLQuery select:[CBLQuerySelect all]
```
pasin: We have changed this API in DB12 as we support projection, and We don't support SELECT * yet as of DB12. The same behavior here will be the following:

```
[CBLQuery select:@[]
```
This means that the statement projects nothing.

jamiltz:
@pasin Thanks for the review! Updated per your comments.

PR #1838, opened 2017-07-24
Sped up array/dictionary conversions and Fleece encoding

snej:
In the TuneMark benchmark's import phase, the time was dominated by adding an NSDictionary to an empty document, and encoding the document to Fleece. I've sped this up, and some other methods that showed up as hot-spots.

The biggest problem was the large number of -conformsToProtocol: calls. This method is very slow. I've replaced those with category methods on NSObject, which the CBL classes override. These are declared in CBLData.h.

There are some more changes in CBLDictionary to reduce the amount of work done in -setDictionary: and -toDictionary.

### DIFF
--- a/Objective-C/CBLArray.m
+++ b/Objective-C/CBLArray.m
@@ -140,16 +140,16 @@
     if (!_array)
         [self copyFleeceData];
     
-    NSMutableArray* array = [NSMutableArray array];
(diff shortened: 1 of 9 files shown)


pasin:
@hideki  and @borrrden 

Could you please review this PR? Most of the codes in this PR is Objective-C language specific.

PR #1857, opened 2017-08-11
Updated query generation to compensate for changed LiteCore "IN" op

snej:
See couchbase/couchbase-lite-core#204

The LC change (not yet pushed to Github) is that the "IN" operation now takes exactly 2 args: first the item, second the array expression. Also I added an array-literal syntax so you can do the SQL-style "IN" where you give a list of values. That's a "[]" expression where the args are the array elements.


rajagp:
@pasin  Does this mean that we can use any array literals with IN operator? I don't think I've ever been able to do that (i.e. the only option exposed via the API is to specify the property name corresponding to the array)

pasin:
Yes, we can use array literals in IN operator. Here is the unit test:
https://github.com/couchbase/couchbase-lite-ios/blob/ababc38d8a476085726bd6980099a37f62da557b/Swift/Tests/QueryTest.swift#L242-L255.

PR #1901, opened 2017-09-19
Implement replicator backgrounding option

pasin:
* Add runInBackground option to the ReplicatorConfiguration. The default value is NO.

* When runInBackground is not enabled, the replicator will be suspended or stopped if the replicator is not continuous after it becomes inactive.

* When runInBackground is enabled, the replicator will not continue to run in the background and it’s application’s responsibility to manage the background task.

* Created a new serial queue for replicator instead of using main queue; the notification is still posted on the main queue.

* Made start / stop running inside the replicator dispatch queue.

* MYBackgroundMonitor is not compiled on XCode9; wait for https://github.com/snej/MYUtilities/pull/16 to be merged.

### DIFF
--- a/Objective-C/CBLReplicator.mm
+++ b/Objective-C/CBLReplicator.mm
@@ -6,6 +6,7 @@
 //  Copyright © 2017 Couchbase. All rights reserved.
 //
 
+#import "CBLReplicator+Backgrounding.h"
 #import "CBLReplicator+Internal.h"
 #import "CBLReplicatorChange+Internal.h"
 #import "CBLReplicatorConfiguration.h"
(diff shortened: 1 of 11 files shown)


pasin:
I added the ability to resume the replicator when the app is already in background, and updated the documentation. Please review.

pasin:
Found a bug that the replicator doesn't get suspended when starting the replicator in the background and when there is an offline or transient error occurred. So I fixed the bug and push one more commit.

Objective-C/CBLReplicator.mm
```diff
@@ -118,20 +118,24 @@ - (void) clearRepl {
...
+        CBLLog(Sync, @"%@: Starting", self);
+        _retryCount = 0;
+        _suspended = NO;
+        [self _start];
+    });
 }
 
 
 - (void) retry {
     if (_repl || _rawStatus.level != kC4Offline)
```
snej: This accesses `_repl` on the calling queue, which isn't thread-safe since it's also used on the `_dispatchQueue` (e.g. in -start.) There are probably other places where this happens.

PR #1909, opened 2017-09-23
Fixed #1758 Updated designed doc’s map function not get re-compiled

### LINKED ISSUE #1758: Updating a design doc view map function does not take effect
The issue was originally reported against android:
https://github.com/couchbase/couchbase-lite-android/issues/1139

Test case:

1. Add 10 docs to the client with content
2. Add 5 docs to the client without content
3. Create design doc version 1 to fetch doc._id, doc._rev for docs with content
4. Run a query and check for 10 expected docs with design doc version 1
5. Update design doc to version 2 to fetch doc._id, doc._rev for docs with no content
6. Run a query and check for 5 expected docs with design doc version 2
-> Even though design doc is updated to version 2, the second query will use design doc 1 and return 10 docs instead of 5

Steps to reproduce:

pytest -k test_design_doc_update --liteserv-platform=macosx --liteserv-version=1.4.0-3 --liteserv-host=localhost --liteserv-port=59840 --liteserv-storage-engine=SQLite --sync-gateway-version=1.4.2-368 --sync-gateway-mode=cc --server-version=5.0.0-2873 testsuites/listener/shared/client_sg/

--------
* Version:   1.4.0-3
* Client OS: mac
* Server:  SG 1.4.2-368

pasin:
If view’s properties is not nil, always compare the version (SHA1 hash of the new view’s properties) whether it is the same as the current version or not. If version is not the same, recompile the map/reduce fn and re-register the map/reduce fn.

#1758

### DIFF
--- a/Source/CBLView+REST.m
+++ b/Source/CBLView+REST.m
@@ -16,16 +16,18 @@
 
 
 - (CBLStatus) compileFromDesignDoc {
-    if (self.registeredMapBlock != nil)
-        return kCBLStatusOK;
-
     // see if there's a design doc with a CouchDB-style view definition we can compile:
     NSString* language;
     NSDictionary* viewProps = $castIf(NSDictionary, [self.database getDesignDocFunction: self.name
                                                                                     key: @"views"
                                                                                language: &language]);
-    if (!viewProps)
-        return kCBLStatusNotFound;
+    if (!viewProps) {
+        if (self.registeredMapBlock != nil)
+            return kCBLStatusOK;
+        else
+            return kCBLStatusNotFound;
+    }
+    
     LogTo(View, @"%@: Attempting to compile %@ from design doc", self.name, language);
     if (![CBLView compiler])
         return kCBLStatusNotImplemented;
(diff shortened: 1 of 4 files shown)


Source/CBLView+REST.m
```diff
@@ -16,16 +16,18 @@ @implementation CBLView (REST)
...
     // see if there's a design doc with a CouchDB-style view definition we can compile:
     NSString* language;
     NSDictionary* viewProps = $castIf(NSDictionary, [self.database getDesignDocFunction: self.name
                                                                                     key: @"views"
                                                                                language: &language]);
-    if (!viewProps)
-        return kCBLStatusNotFound;
+    if (!viewProps) {
+        if (self.registeredMapBlock != nil)
+            return kCBLStatusOK;
```
snej: This change means that after the design doc is deleted, or the map function is removed from the design doc, the view continues to work. Isn't that wrong?

Source/CBLView+REST.m
```diff
@@ -16,16 +16,18 @@ @implementation CBLView (REST)
 
 
 - (CBLStatus) compileFromDesignDoc {
-    if (self.registeredMapBlock != nil)
```
snej: This same check you deleted here is still in `View.mapBlock`, which calls this method. That means that querying a view from the native API still has the bug. (We have at least one developer doing this, posting questions on the forum.)
I don't know what the best way to fix that is. Always checking for a design doc when querying any view will affect performance. Maybe add a flag to CBLView that remembers that the map block comes from a design doc, and only re-check the design doc if this flag is set?

pasin:
I fixed the code per your comment. The confusion part is that there could be a native design doc like view defined and a REST design doc query against it as well. Could you please review?

PR #1915, opened 2017-09-28
Fixed #1750 REST returns code when getting deleted doc with revid

### LINKED ISSUE #1750: Wrong return code for listener causes problems with puller and deleted items
From https://github.com/couchbase/couchbase-lite-net/issues/850

Sync Gateway will return 200 with a small body when a deleted revision is requested by ID, but CBL will return 404 with "reason: deleted". This will cause a puller to skip this revision and never delete a document locally if it happens to get into a non-bulk pull request. The simple test case from a forum post seem to be start up instances, one with a listener running and the other with continuous push and pull, and then add a document to the listener side and subsequently delete it with enough delay in between to ensure that both end up requesting the document directly. This logic appears to affect Objective-C as well, but I have not checked Java.

pasin:
When getting a deleted doc with revid, need to return status OK with doc info and _deleted = true.

#1750

### DIFF
--- a/Source/CBL_Router+Handlers.m
+++ b/Source/CBL_Router+Handlers.m
@@ -618,9 +618,15 @@ static NSArray* parseJSONRevArrayQuery(NSString* queryStr) {
             if (rev)
                 rev = [self applyOptions: options toRevision: rev status: &status];
             if (!rev) {
-                if (status == kCBLStatusDeleted)
-                    _response.statusReason = @"deleted";
-                else
+                if (status == kCBLStatusDeleted) {
+                    if (revID) {
+                        status = kCBLStatusOK;
+                        _response.bodyObject = $dict({@"_id", docID},
+                                                     {@"_rev", revID.asString},
+                                                     {@"_deleted", $true});
+                    } else
+                        _response.statusReason = @"deleted";
+                } else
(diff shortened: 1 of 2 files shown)


Source/CBL_Router+Handlers.m
```diff
@@ -618,9 +618,15 @@ - (CBLStatus) do_GET: (CBLDatabase*)db docID: (NSString*)docID {
             if (rev)
                 rev = [self applyOptions: options toRevision: rev status: &status];
             if (!rev) {
-                if (status == kCBLStatusDeleted)
-                    _response.statusReason = @"deleted";
-                else
+                if (status == kCBLStatusDeleted) {
+                    if (revID) {
+                        status = kCBLStatusOK;
```
snej: I don't think this line is reachable. If -getDocumentWithID:revisionID:withBody:status: is called with a non-nil revID, and it finds a deleted revision, it will return that revision. It will only return nil if there is no such revision, and in that case status will be set to kCBLStatusNotFound.

pasin: You are right. It turns out iOS doesn't have this issue. I will close the PR and commit only unit test. @borrrden and @hideki, can you review the issue / fix again?

PR #1928, opened 2017-10-10
Fixed #1918: Skip the removal rev when getting ancestor rev ids

### LINKED ISSUE #1918: Channel removal potentially disrupts attachments
In the following scenario, attachments will not be able to be found:

1. Client goes offline
2. Attachments are added to the document
3. Channel removal happens and the client loses access to the document
4. Client comes online and receives the removal (since there are no attachments on a removal the client assumes there are none at all)
5. Client regains access and discovers that there are attachments, and searches for them. However it won't find them because it will only check back to the last known revision (the removed one)

https://github.com/couchbase/couchbase-lite-net/issues/910

--------
* Version:   1.4.0
* Client OS: All
* Server:    All

pasin:
This logic is ported from .NET (https://github.com/couchbase/couchbase-lite-net/commit/07c1374d373b4d0ce8a4bedb640294aeec232d71) to fix channel removal potentially disrupts attachments issue (#1918)

The following code are slightly different from .NET:

https://github.com/couchbase/couchbase-lite-ios/blob/2c86216a1f8e921b1e9cc1a1bacf652d11903dd9/Source/CBLBulkDownloader.m#L58 (.NET always true)

https://github.com/couchbase/couchbase-lite-ios/blob/2c86216a1f8e921b1e9cc1a1bacf652d11903dd9/Source/CBLRestPuller.m#L429 (.NET always true)

https://github.com/couchbase/couchbase-lite-ios/blob/2c86216a1f8e921b1e9cc1a1bacf652d11903dd9/Source/CBL_ForestDBStorage.mm#L459-L465 (.NET doesn't load the revision body).

### DIFF
--- a/Source/CBLBulkDownloader.m
+++ b/Source/CBLBulkDownloader.m
@@ -54,7 +54,8 @@
         NSArray<CBL_RevID*>* possibleAncestors;
         possibleAncestors = [database.storage getPossibleAncestorRevisionIDs: rev
                                                        limit: kMaxNumberOfAttsSince
-                                                  haveBodies: (attachments ? &haveBodies : NULL)];
+                                                  haveBodies: (attachments ? &haveBodies : NULL)
+                                              withBodiesOnly: attachments];
         NSMutableDictionary* key = $mdict({@"id",  rev.docID},
                                           {@"rev", rev.revIDString});
         if (possibleAncestors) {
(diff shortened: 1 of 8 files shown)


pasin:
@snej could you please review this for 1.4.1?

PR #1943, opened 2017-11-01
Fixed #1927 session cookie overwritten by the cookie in shared storage

### LINKED ISSUE #1927: 2.0: Swift: SesssionAuthenticator doesn't work
After logging in and get all the sync_session information, I'm trying to authenticate with SessionAuthenticator: 

        replicatorConfig.authenticator = SessionAuthenticator(sessionID: syncSessionInfo.sessionID, expires: syncSessionInfo.expiration, cookieName: syncSessionInfo.cookieName)

But it always fails with 401 unauthorized. 

The same user login and server works with Android login though.

Tried also BasicAuthenticator with Swift, and it does work.

--------
* Version:   2.0DB016
* Client OS: iOS 11
* Server:    Sync Server 1.5

pasin:
Set request.HTTPShouldHandleCookies to NO so that the CBLHTTPLogic will not add the cookies from the shared cookie storage. Also we don’t actually use the shared cookie storage.

### DIFF
--- a/Objective-C/Internal/Replicator/CBLWebSocket.mm
+++ b/Objective-C/Internal/Replicator/CBLWebSocket.mm
@@ -116,6 +116,7 @@ static void doCompletedReceive(C4Socket* s, size_t byteCount) {
         _options = AllocedDict(options);
 
         NSMutableURLRequest *request = [NSMutableURLRequest requestWithURL: url];
+        request.HTTPShouldHandleCookies = NO;
         _logic = [[CBLHTTPLogic alloc] initWithURLRequest: request];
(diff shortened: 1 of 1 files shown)


pasin:
Another way to fix this is to remove this [lines](https://github.com/couchbase/couchbase-lite-ios/blob/feature/2.0/Objective-C/Internal/Replicator/CBLHTTPLogic.m#L142-L150). @snej should the logic to set the cookies from the shared storage be removed as we don't use the shared cookie storage?

pasin:
I think the fix is right. I'm merging this in myself.

PR #1966, opened 2017-11-23
Fixed #1948 Make database operations thread safe

### LINKED ISSUE #1948: 2.0: Make Database thread safe
We will make the Database thread safe.

pasin:
* Created CBL_LOCK(m) macro for locking.
* Applied the lock to non-thread safe database operations by following the guideline from https://github.com/couchbase/couchbase-lite-core/wiki/Thread-Safety.

#1948

### DIFF
--- a/Objective-C/CBLDatabase.mm
+++ b/Objective-C/CBLDatabase.mm
@@ -37,6 +37,7 @@ using namespace fleece;
     CBLDatabaseConfiguration* _config;
     CBLPredicateQuery* _allDocsQuery;
     
+    NSObject* _listenersMutex;
     C4DatabaseObserver* _dbObs;
     NSMutableSet* _dbListenerTokens;
     
@@ -46,6 +47,7 @@ using namespace fleece;
 
 
 @synthesize name=_name;
+@synthesize dispatchQueue=_dispatchQueue;
 @synthesize c4db=_c4db, sharedKeys=_sharedKeys;
 @synthesize replications=_replications, activeReplications=_activeReplications;
 
@@ -59,7 +61,7 @@ static const C4DatabaseConfig kDBConfig = {
 
 static void dbObserverCallback(C4DatabaseObserver* obs, void* context) {
     CBLDatabase *db = (__bridge CBLDatabase *)context;
-    dispatch_async(dispatch_get_main_queue(), ^{        //TODO: Support other queues
+    dispatch_async(db.dispatchQueue, ^{
         [db postDatabaseChanged];
     });
 }
@@ -69,7 +71,7 @@ static void docObserverCallback(C4DocumentObserver* obs, C4Slice docID, C4Sequen
                                 void *context)
 {
     CBLDatabase *db = (__bridge CBLDatabase *)context;
-    dispatch_async(dispatch_get_main_queue(), ^{        //TODO: Support other queues
+    dispatch_async(db.dispatchQueue, ^{
         [db postDocumentChanged: slice2string(docID)];
     });
 }
@@ -99,6 +101,9 @@ static void docObserverCallback(C4DocumentObserver* obs, C4Slice docID, C4Sequen
         _config = config != nil? [config copy] : [CBLDatabaseConfiguration new];
         if (![self open: outError])
             return nil;
+        
+        _dispatchQueue = dispatch_queue_create("CBLDatabase", DISPATCH_QUEUE_SERIAL);
+        _listenersMutex = [NSObject new];
         _replications = [NSMapTable strongToWeakObjectsMapTable];
         _activeReplications = [NSMutableSet new];
     }
@@ -123,12 +128,16 @@ static void docObserverCallback(C4DocumentObserver* obs, C4Slice docID, C4Sequen
 
 
 - (NSString*) path {
-    return _c4db != nullptr ? sliceResult2FilesystemPath(c4db_getPath(_c4db)) : nil;
+    CBL_LOCK(self) {
+        return _c4db != nullptr ? sliceResult2FilesystemPath(c4db_getPath(_c4db)) : nil;
+    }
 }
 
 
 - (uint64_t) count {
-    return _c4db != nullptr ? c4db_getDocumentCount(_c4db) : 0;
+    CBL_LOCK(self) {
+        return _c4db != nullptr ? c4db_getDocumentCount(_c4db) : 0;
+    }
 }
 
 
@@ -141,7 +150,9 @@ static void docObserverCallback(C4DocumentObserver* obs, C4Slice docID, C4Sequen
 
 
 - (CBLDocument*) documentWithID: (NSString*)documentID {
-    return [self documentWithID: documentID mustExist: YES error: nil];
+    CBL_LOCK(self) {
+        return [self documentWithID: documentID error: nil];
+    }
 }
 
 
@@ -149,8 +160,9 @@ static void docObserverCallback(C4DocumentObserver* obs, C4Slice docID, C4Sequen
 
 
 - (BOOL) containsDocumentWithID: (NSString*)docID {
-    id doc = [self documentWithID: docID mustExist: YES error: nil];
-    return doc != nil;
+    CBL_LOCK(self) {
+        return [self documentWithID: docID error: nil] != nil;
+    }
 }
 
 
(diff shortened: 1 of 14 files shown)


pasin:
@snej Along with the PR, I opened #1967 that I need some guideline to prevent race when one thread is getting the document from the database and the other thread is trying to create a mutable copy of the document with the same ID, **BUT** the two same documents are different instances.

pasin:
@snej do you have a chance to look at this?

Objective-C/CBLReplicator.mm
```diff
@@ -180,8 +180,12 @@ - (void) _start {
         .callbackContext = (__bridge void*)self,
         // TODO: Add .validationFunc (public API TBD)
     };
-    C4Error err;
-    _repl = c4repl_new(_config.database.c4db, addr, dbName, otherDB.c4db, params, &err);
+
```
snej: This PR isn't supposed to make CBLReplicator thread-safe; that's coming later, right? Because there are still race conditions involving `_repl`.

Objective-C/Internal/CBLLock.h
```diff
@@ -0,0 +1,21 @@
+//
+//  CBLLock.h
+//  CouchbaseLite
+//
+//  Created by Pasin Suriyentrakorn on 11/22/17.
+//  Copyright © 2017 Couchbase. All rights reserved.
+//
+
+
+#ifndef CBLLock_h
```
snej: Just use `#pragma once`.

Objective-C/Internal/CBLLock.h
```diff
@@ -0,0 +1,21 @@
...
+#define CBLLock_h
+
+#if CBL_THREADSAFE
```
snej: If `CBL_THREADSAFE` is undefined or 0, the code won't compile because `CBL_LOCK` is undefined. So is there any point in having `CBL_THREADSAFE` at all, since it needs to be `1`?

Objective-C/CBLDatabase.mm
```diff
@@ -141,16 +150,19 @@ - (CBLDatabaseConfiguration*) config {
 
 
 - (CBLDocument*) documentWithID: (NSString*)documentID {
-    return [self documentWithID: documentID mustExist: YES error: nil];
+    CBL_LOCK(self) {
```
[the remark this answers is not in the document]
pasin: I'm trying to put the lock at the API level to make sure that I don't miss anything. I will make change on this.

Objective-C/CBLDatabase.mm
```diff
@@ -99,6 +101,9 @@ - (instancetype) initWithName: (NSString*)name
         _config = config != nil? [config copy] : [CBLDatabaseConfiguration new];
         if (![self open: outError])
             return nil;
+        
+        _dispatchQueue = dispatch_queue_create("CBLDatabase", DISPATCH_QUEUE_SERIAL);
```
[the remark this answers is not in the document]
pasin: That's a good idea.

Objective-C/CBLDatabase.mm
```diff
@@ -622,123 +676,149 @@ - (BOOL) prepareDocument: (CBLDocument*)document error: (NSError**)error {
...
-            if(docIDs.count > 0) {
-                CBLDatabaseChange* change = [[CBLDatabaseChange alloc] initWithDocumentIDs: docIDs
-                                                                                isExternal: external];
-                for (CBLChangeListenerToken* token in _dbListenerTokens) {
-                    void (^listener)(CBLDatabaseChange*) = token.listener;
-                    dispatch_async(token.queue, ^{
-                        listener(change);
-                    });
+    CBL_LOCK(self) {
+        if (!_dbObs || !_c4db || c4db_isInTransaction(_c4db))
```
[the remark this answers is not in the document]
pasin: I think I'm over thinking about having two locks. I will change to just one lock.

Objective-C/Internal/CBLLock.h
pasin: Just in case we want to turn it off to measure the performance.

Objective-C/Internal/CBLLock.h
pasin: Oh, I don't know this option.

Objective-C/CBLReplicator.mm
pasin: Yes. I haven't made the replicator thread-safe yet.

Objective-C/CBLQuery.mm
```diff
@@ -494,23 +496,36 @@ + (instancetype) selectDistinct: (NSArray<CBLQuerySelectResult*>*)select
...
-    if (!_c4Query && ![self check: outError])
-        return nil;
-    return sliceResult2string(c4query_explain(_c4Query));
+    CBL_LOCK(self) {
+        if (!_c4Query && ![self check: outError])
+            return nil;
+    }
+    
+    __block NSString* result;
+    [self.database withLock: ^{
```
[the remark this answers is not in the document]
pasin: OK, I will change accordingly.

Objective-C/CBLQuery.mm
```diff
@@ -494,23 +496,36 @@ + (instancetype) selectDistinct: (NSArray<CBLQuerySelectResult*>*)select
 
 
 - (NSString*) explain: (NSError**)outError {
-    if (!_c4Query && ![self check: outError])
-        return nil;
-    return sliceResult2string(c4query_explain(_c4Query));
+    CBL_LOCK(self) {
```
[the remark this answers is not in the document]
pasin: OK

pasin:
@snej I have made changes accordingly. Could you please re-review? Thanks.

Objective-C/CBLDatabase.mm
```diff
@@ -864,139 +870,201 @@ - (CBLDocument*) saveDocument: (CBLDocument*)document
...
-    return [[CBLDocument alloc] initWithDatabase: self
-                                      documentID: document.id
-                                           c4Doc: c4Doc];
 }
 
 
 #pragma mark - RESOLVING REPLICATED CONFLICTS:
 
 
+// Called inside the replicator's database transaction.
```
snej: I don't think this is true; the calling method doesn't open a transaction, and it's called from dispatch_async.

Objective-C/CBLDatabase.mm
```diff
@@ -864,139 +870,201 @@ - (CBLDocument*) saveDocument: (CBLDocument*)document
...
-    if (!otherDoc || ![otherDoc selectConflictingRevision])
-        return false;
-
-    // Read the common ancestor revision (if it's available):
-    auto baseDoc = [[CBLDocument alloc] initWithDatabase: self
+                             error: (NSError**)outError
+{
+    CBLDocument* doc, *otherDoc, *baseDoc;
+    CBL_LOCK(self) {
+        // Open transaction to be able to see current uncommitted changes
```
snej: What does this mean? Opening a transaction doesn't let you see changes made by other transactions.

Objective-C/CBLQuery.mm
```diff
@@ -494,14 +497,19 @@ + (instancetype) selectDistinct: (NSArray<CBLQuerySelectResult*>*)select
...
 
 - (NSString*) explain: (NSError**)outError {
-    if (!_c4Query && ![self check: outError])
+    if (![self check: outError])
         return nil;
-    return sliceResult2string(c4query_explain(_c4Query));
+    
+    NSString* result;
+    CBL_LOCK(self.database) {
+        result = sliceResult2string(c4query_explain(_c4Query));
```
snej: Can simplify this to just `return sliceResult2string...` without the variable.

Objective-C/CBLQueryResult.mm
```diff
@@ -46,7 +46,9 @@ - (instancetype) initWithResultSet: (CBLQueryResultSet*)rs
 
 
 - (NSUInteger) count {
-    return c4query_columnCount(_rs.c4Query);
+    CBL_LOCK(self) {
```
snej: Other uses of C4Query (in CBLQuery.mm) are synchronized with the CBLDatabase, so this should be too, right?

Objective-C/CBLQueryResultSet.mm
```diff
@@ -87,17 +92,22 @@ - (CBLDatabase*) database {
...
-    id row = nil;
-    if (c4queryenum_next(_c4enum, &_error)) {
-        row = self.currentObject;
-    } else if (_error.code)
-        CBLWarnError(Query, @"%@[%p] error: %d/%d", [self class], self, _error.domain, _error.code);
-    else
-        CBLLog(Query, @"End of query enumeration (%p)", _c4enum);
-    return row;
+    CBLDatabase* strongDB = self.database;
+    Assert(strongDB, @"Database has been released.");
```
snej: The message seems wrong -- it's really that the query's been dealloced, not the db.
And is it required that the app hold onto the CBLQuery object for as long as it's using the result set, or else it'll crash? That seems kind of fragile.

Objective-C/CBLQueryResultSet.mm
```diff
@@ -110,17 +120,29 @@ - (id) currentObject {
 
 // Called by CBLQueryResultsArray
 - (id) objectAtIndex: (NSUInteger)index {
-    if (!c4queryenum_seek(_c4enum, index, &_error)) {
-        NSString* message = sliceResult2string(c4error_getMessage(_error));
-        [NSException raise: NSInternalInconsistencyException
-                    format: @"CBLQueryEnumerator couldn't get a value: %@", message];
+    CBLDatabase* strongDB = self.database;
+    Assert(strongDB, @"Database has been released.");
```
snej: If this check is going to happen every time self.database is called, maybe move it into the `database` method.

Objective-C/CBLQueryResultSet.mm
```diff
@@ -143,8 +165,17 @@ - (CBLQueryResultSet*) refresh: (NSError**)outError {
     if (outError)
         *outError = nil;
     
+    CBLDatabase* strongDB = self.database;
+    if (!strongDB) {
+        CBLWarn(Query, @"Database has already been released.");
```
snej: Why is this just a warning here but an exception in the other methods?

pasin:
I push the fix per review and discussion. Could you re-review again? Thanks!

pasin:
@hideki and @borrrden , could you align the implementation on android and .net with this PR?

Issue #1958, opened 2017-11-10
2.0: When database is closed, the replicator and live query should be stopped.


pasin:
When a database is closed, the replicator and live query should be stopped.

borrrden:
Just for clarification:  If a DB is closed, its change event will no longer be fired and nothing will happen for that live query after that point.  Does there still need to be something explicit happening here?  Could I also free the native objects at this point for .NET?

pasin:
I think you can free the object because to reopen the database, a new database object should be created.

jayzhanghs:
Is this the cause of failure to delete database which I see pretty often in my case?  We sometimes need to delete the database for a fresh re-sync.

snej:
Probably.

pasin:
This feature has been implemented and merged into the feature/2.0 branch. We could close the issue now.

PR #1990, opened 2017-12-19
Fix for 1958 : When database is closed, the replicator and live query…

rajagp:
This is fix for https://github.com/couchbase/couchbase-lite-ios/issues/1958

### DIFF
--- a/Objective-C/CBLDatabase.mm
+++ b/Objective-C/CBLDatabase.mm
@@ -28,6 +28,7 @@
 #import "CBLStringBytes.h"
 #import "CBLStatus.h"
 
+
 using namespace fleece;
 
 #define kDBExtension @"cblite2"
@@ -47,7 +48,7 @@ using namespace fleece;
 
 @synthesize name=_name;
 @synthesize c4db=_c4db, sharedKeys=_sharedKeys;
-@synthesize replications=_replications, activeReplications=_activeReplications;
+@synthesize replications=_replications, activeReplications=_activeReplications, liveQueries= _liveQueries;
 
 
 static const C4DatabaseConfig kDBConfig = {
@@ -101,6 +102,7 @@ static void docObserverCallback(C4DocumentObserver* obs, C4Slice docID, C4Sequen
             return nil;
         _replications = [NSMapTable strongToWeakObjectsMapTable];
         _activeReplications = [NSMutableSet new];
+        _liveQueries = [NSMutableSet new];
     }
     return self;
 }
@@ -236,6 +238,10 @@ static void docObserverCallback(C4DocumentObserver* obs, C4Slice docID, C4Sequen
     
     CBLLog(Database, @"Closing %@ at path %@", self, self.path);
     
+    [self stopActiveReplications];
+    
+    [self stopLiveQueries];
+    
     _allDocsQuery = nil;
     
     C4Error err;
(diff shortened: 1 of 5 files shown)


Objective-C/Internal/CBLDatabase+Internal.h
```diff
@@ -17,6 +17,8 @@
 #import "CBLMutableDocument.h"
 #import "CBLDocumentChange.h"
 #import "CBLJSONCoding.h"
+#import "CBLReplicator.h"
```
snej: You don't need to import these headers; just use `@class` to declare the class name.

Objective-C/CBLDatabase.mm
```diff
@@ -47,7 +48,7 @@ @implementation CBLDatabase {
 
 @synthesize name=_name;
 @synthesize c4db=_c4db, sharedKeys=_sharedKeys;
-@synthesize replications=_replications, activeReplications=_activeReplications;
+@synthesize replications=_replications, activeReplications=_activeReplications, liveQueries= _liveQueries;
```
snej: Limit lines to 100 characters, please.

Objective-C/CBLDatabase.mm
```diff
@@ -496,6 +506,18 @@ - (BOOL) open: (NSError**)outError {
     return YES;
 }
 
+- (void) stopActiveReplications {
+    // Stop all active replications
+    for (CBLReplicator* repl in _activeReplications) {
```
snej: This will cause trouble when the replicators being stopped remove themselves from `_activeReplications` -- NSSet will throw an exception because the set is being mutated during enumeration. To work around this you'll need to make a copy of the set and enumerate that.

Objective-C/CBLDatabase.mm
```diff
@@ -496,6 +506,18 @@ - (BOOL) open: (NSError**)outError {
...
+        [repl stop];
+    }
+}
+
+-(void) stopLiveQueries {
+    for (CBLLiveQuery* query in _liveQueries) {
```
snej: Same issue as above (line 511).

snej:
Could you add a unit test for this, too?

Objective-C/CBLDatabase.mm
rajagp: Will do.

Objective-C/CBLDatabase.mm
rajagp: Good catch. Will do

Objective-C/CBLDatabase.mm
rajagp: Good catch. Will do

Objective-C/Internal/CBLDatabase+Internal.h
rajagp: I couldn't make this a forward declaration because I need to call "stop" on the replicator objects

rajagp:
Yes. Will add unit tests.  Also noticed that feature/2.0 has diverged from the version from which this branch was created causing conflicts. I will update to latest feature/2.0 when I submit the revised version of the PR.

rajagp:
@snej  : I have updated this branch with the suggested changes with the exception of the forward declaration change (See comments earlier). Can you review please.

Objective-C/Internal/CBLDatabase+Internal.h
snej: Not in this header, though! The header just needs to know that `CBLReplicator` is a class name. In the .m file(s) you would import the headers.

snej:
Oops -- Just realized that CBLDatabase is thread-safe now (correct, @pasin?), so the accesses to `_liveQueries` and `_activeReplications` need to be protected by a mutex.

pasin:
Yes, those two should be protected.

PR #2011, opened 2018-01-10
Fix(?) for threading issues with CBLUITableSource

codehearted:
Hopefully this will fix the errors such as "This application is modifying the autolayout engine from a background thread after the engine was accessed from the main thread. This can lead to engine corruption and weird crashes."

Addresses: [ATOM-2128]


codehearted:
Sorry, I accidentally created this in the wrong repository.

PR #2046, opened 2018-01-26
Fixed #1981 Make replicator thread-safe and fix stop after offline

### LINKED ISSUE #1981: 2.0 : Make Replicator thread-safe
We should also make the replicator thread safe. Also internally the replicator handles the change events which comes from a different thread. We need to make sure that start() and stop() work thread-safely with the event handlers.

pasin:
* Made the replicator thread-safe by using lock.

* Fix the stop status not getting reported when calling replicator.stop() after the replicator is in the offline state.

#1981

### DIFF
--- a/Objective-C/CBLReplicator.mm
+++ b/Objective-C/CBLReplicator.mm
@@ -120,22 +120,26 @@ static NSTimeInterval retryDelay(unsigned retryCount) {
 
 
 - (void) start {
-    if (_repl) {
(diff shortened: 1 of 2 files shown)


pasin:
Just rebased with the current feature/2.0-dev branch.

hideki:
PR looks fine. From code diff, I am not 100% sure if all scenarios are covered. I will merge the PR, but we need to keep observing this.

Issue #2047, opened 2018-01-26
2.0 : Include unit tests to cover cross and left join


rajagp:
Parity with  .Net tests . Port tests over

PR #2049, opened 2018-01-27
Ported cross join and left join tests from .Net

rajagp:
Fix for #2047 . Ported the tests from .Net to ObjC and Swift.


rajagp:
@borrrden  After reviewing your tests , I determined that your tests covered the cases that I was planning to cover with mine . I was using a different data set but I determined that didn't affect the outcome of the tests so just went with yours.

PR #2059, opened 2018-02-03
Fixed #2058 : Implement replicator backgrounding support

### LINKED ISSUE #2058: 2.0: Support Replicator Backgrounding
Added a flag to enable / disable feature in CBLReplicatorConfiguration. The current proposed naming of the flag is `allowReplicatingInBackground`. 
* If the allowReplicatingInBackground = NO (default), the replicator will be suspended when the app goes into background and the replicator is IDLE. The replicator will be resume when the app is brought foreground.
* If the allowReplicatingInBackground = YES, the replicator will keep running in background BUT it's the app's responsibility to extend the background task to keep the replicator running properly.

pasin:
* Added ReplicatorConfiguration.allowReplicatingInBackground flag for disabling or enabling the feature.

* ReplicatorConfiguration.allowReplicatingInBackground=NO, the replicator will be suspended when the app goes into background and the replicator becomes idle. The replicator will be resumed when the app is brought foreground.

* ReplicatorConfiguration.allowReplicatingInBackground=YES, the replicator will keep running the background BUT it is the applicataion’s responsibility to extend the background task to keep the replicator running properly.

* If the replicator is started while the app is in background, the replicator will keep running until IDLE before getting suspended.

#2058

### DIFF
--- a/Objective-C/CBLReplicator.mm
+++ b/Objective-C/CBLReplicator.mm
@@ -17,6 +17,7 @@
 //  limitations under the License.
 //
 
+#import "CBLReplicator+Backgrounding.h"
 #import "CBLReplicator+Internal.h"
 #import "CBLReplicatorChange+Internal.h"
 #import "CBLReplicatorConfiguration.h"
@@ -77,11 +78,14 @@ static NSTimeInterval retryDelay(unsigned retryCount) {
     unsigned _retryCount;
     CBLReachability* _reachability;
     NSMutableSet* _listenerTokens;
+    BOOL _isStopping;
+    BOOL _isSuspending;
 }
 
 @synthesize config=_config;
 @synthesize status=_status;
-
+@synthesize bgMonitor=_bgMonitor;
+@synthesize suspended=_suspended;
 
 + (void) initialize {
     if (self == [CBLReplicator class]) {
@@ -139,6 +143,8 @@ static NSTimeInterval retryDelay(unsigned retryCount) {
         
         CBLLog(Sync, @"%@: Starting", self);
         _retryCount = 0;
+        _suspended = NO;
+        _isSuspending = NO;
         [self _start];
     }
 }
@@ -146,8 +152,11 @@ static NSTimeInterval retryDelay(unsigned retryCount) {
 
 - (void) retry {
     CBL_LOCK(self) {
-        if (_repl || _rawStatus.level != kC4Offline)
+        if (_repl || _rawStatus.level != kC4Offline) {
+            CBLLog(Sync, @"%@: Ignore retrying (status = %d)", self, _rawStatus.level);
             return;
+        }
+        
         CBLLog(Sync, @"%@: Retrying...", self);
         [self _start];
     }
@@ -210,6 +219,11 @@ static NSTimeInterval retryDelay(unsigned retryCount) {
         CBL_LOCK(_config.database) {
             [_config.database.activeReplications addObject: self];     // keeps me from being dealloced
         }
+        
+#if TARGET_OS_IPHONE
+        if (!_config.allowReplicatingInBackground)
+            [self setupBackgrounding];
+#endif
     } else {
         status = {kC4Stopped, {}, err};
     }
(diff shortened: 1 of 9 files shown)


Objective-C/CBLReplicatorConfiguration.h
```diff
@@ -92,6 +92,19 @@ typedef enum {
  */
 @property (nonatomic, nullable) NSArray<NSString*>* documentIDs;
 
+
+#if TARGET_OS_IPHONE
+/**
+ Allows the replicator to continue replicating in the background. The default
+ value is NO, which means that the replicator will suspend itself when the
+ replicator detects that the application is running in the background.
```
snej: Better to say "the application is being backgrounded".

Objective-C/CBLReplicatorConfiguration.h
```diff
@@ -92,6 +92,19 @@ typedef enum {
...
+ 
+ If setting the value to YES, please ensure that the application requests
+ for extending the background task properly.
```
snej: How about "your application delegate requests background time from the OS until the replication finishes".

Actually, is this true? The code seems to take care of requesting a background task on its own, just as 1.x did.

Objective-C/CBLReplicatorConfiguration.m
```diff
@@ -132,13 +136,24 @@ - (void) setChannels: (NSArray<NSString *> *)channels {
 }
 
 
+#if TARGET_OS_IPHONE
+- (void) setAllowReplicatingInBackground: (BOOL)allowReplicatingInBackground {
+    [self checkReadonly];
+    
+    if (_allowReplicatingInBackground != allowReplicatingInBackground) {
```
snej: The `if` doesn't accomplish anything; just set the variable. Actually, the same is true of all the other setter methods.

Objective-C/CBLReplicator.mm
```diff
@@ -77,11 +78,14 @@ @implementation CBLReplicator
     unsigned _retryCount;
     CBLReachability* _reachability;
     NSMutableSet* _listenerTokens;
+    BOOL _isStopping;
```
snej: Would it be clearer to use a state machine (i.e. a `_state` ivar whose type is an enum) rather than multiple booleans? (I don't know the answer, just wondering if you've considered it.)

Objective-C/CBLReplicator.mm
```diff
@@ -210,6 +219,11 @@ - (void) _start {
         CBL_LOCK(_config.database) {
             [_config.database.activeReplications addObject: self];     // keeps me from being dealloced
         }
+        
+#if TARGET_OS_IPHONE
+        if (!_config.allowReplicatingInBackground)
```
snej: These 2 lines are indented one stop too far.

Objective-C/CBLReplicator.mm
pasin: I think the code could look clearer. I will put the TODO: for refactoring the code later after 2.0.

Objective-C/CBLReplicatorConfiguration.h
pasin: If set to YES, the code didn't start the background process. It will be up to the app developer to do that. This is for the app like location based app that may want to have the replicator keeps running in the background without suspending.

Objective-C/CBLReplicatorConfiguration.m
pasin: I will remove all that.

rajagp:
So is this background support a native iOS only capability. What about other platforms? (For instance Xamarin iOS or Android's asynctask "doInBackground" type ..not sure other options).  (@borrrden , @hideki ). 

@pasin : Regarding this option

> "ReplicatorConfiguration.allowReplicatingInBackground=YES, the replicator will keep running the background BUT it is the applicataion’s responsibility to extend the background task to keep the replicator running properly."

Not entirely clear how this works - Assuming you are referring to [background task API](https://developer.apple.com/documentation/uikit/uiapplication/1623051-beginbackgroundtask). So can you clarify how we "extend" this ? Is there a unit test for this - That will help clarify

pasin:
> So is this background support a native iOS only capability. What about other platforms? 

Android OS allows the replicator to keep running in background so we don't need to suspend the replicator when the app is backgrounded. We could reconsider this once the Android OS starts to limit what could be run the background.

For .NET Xamarin iOS, @borrrden told me that to do that .NET would need to do that as an dependency injection API. I understand that it's not trivial as the replicator's code is in the shared API. We could wait and see the CBL Xamarin iOS adoption and implement the feature later if needed.

Not entirely clear how this works - Assuming you are referring to background task API. So can you clarify how we "extend" this ? Is there a unit test for this 

> Yes, that is the API. Basically the app need to detect then it is enter background via either app delegate or notification. Then the app will need to start a background task. The app could end the background task when the app stops the replicator. This mode is normally for the app like location based app that could run stuff in the background. Regarding the unit test, we don't have the unit test for this. Unit test cannot test app backgrounded.

PR #2065, opened 2018-02-15
MB-27954 Added script to build cbl ios testapp

kimtrang:
Script to build ios TestApp from Jenkins


kimtrang:
Per Pasin comments, this should live in mobile-testkit repo, closing pull request

PR #2068, opened 2018-02-20
 Update save document, delete document, and conflict resolver API

pasin:
* Removed custom Conflict Resolver API.
* Updated Database.saveDocument() and Database.deleteDocument() API.
* Removed Document.isDeleted.
* Updated unit tests.


pasin:
I have separated the API and implementation changes and unit test changes into two commits. The unit test review is optional.

Objective-C/CBLDatabase.mm
```diff
@@ -790,13 +814,107 @@ - (void) freeC4DB {
...
+
+
+- (BOOL) saveDocument: (CBLDocument*)document
+   concurrencyControl: (CBLConcurrencyControl)concurrencyControl
+           asDeletion: (BOOL)deletion
+                error: (NSError**)outError
+{
+    if (deletion && !document.revID) {
+        [NSException raise: NSInvalidArgumentException
+                    format: @"Do not allow to delete a newly created document "
```
snej: Message isn't grammatical. How about "Document doesn't exist in the database"?

Objective-C/CBLDatabase.mm
```diff
@@ -790,13 +814,107 @@ - (void) freeC4DB {
...
+            return NO;
+        
+        // Begin a db transaction:
+        C4Transaction transaction(_c4db);
+        if (!transaction.begin())
+            return convertError(transaction.error(), outError);
+        
+        C4Document* curDoc = nil;
+        if (deletion) {
+            // Check existing, NO-OPS if the document doesn't exist:
```
snej: I don't think you need to make this extra check; won't the `-saveDocument:into:...` call fail with a conflict, and you can detect that?

Objective-C/CBLDatabase.mm
```diff
@@ -790,13 +814,107 @@ - (void) freeC4DB {
...
+            if (curDoc)
+                c4doc_free(curDoc);
+            return NO;
+        }
+        
+        if (newDoc) {
+            if (curDoc)
+                c4doc_free(curDoc);
+            
+            // Save succeeded; now commit the transaction:
```
snej: These lines (867-873) are pretty much identical to 904-909. Could you refactor to avoid the duplication? Like, take these lines out and instead wrap everything from 876-902 in `if (!newDoc) {...}`.

Objective-C/CBLDatabase.mm
```diff
@@ -790,13 +814,107 @@ - (void) freeC4DB {
...
+    CBL_LOCK(self) {
+        if (![self prepareDocument: document error: outError])
+            return NO;
+        
+        // Begin a db transaction:
+        C4Transaction transaction(_c4db);
+        if (!transaction.begin())
+            return convertError(transaction.error(), outError);
+        
+        C4Document* curDoc = nil;
```
snej: I recommend using the utility class `c4::ref<T>` in `c4.hh`. It's a smart pointer that works with most C4 types, freeing them when it goes out of scope. Then you can yank out all the lines here that free `curDoc` and `newDoc`.

Objective-C/CBLDatabase.mm
pasin: Currently when saving a deleted doc to non-existing doc, LC returns error code 15 (invalid parameter). I'm not sure that is the right approach to check the error code 15 and no-ops.

Objective-C/CBLDatabase.mm
pasin: I tried to use c4::ref. Giving that I'm not good at C++, I don't know how to make it retained when passing to keep the object in another class. Basically I would like to keep the C4Document (curDoc / newDoc) in the CBLC4Document class. Can you suggest me on this?

pasin: Alternatively I could do @try @finally to free curDoc and newDoc in the finally block. What do you think?

Objective-C/CBLDatabase.mm
pasin: Would it make sense to have LC return NOT_FOUND error instead so we can check the error code for this case? The current INVALID_ARGUMENT (15) error code returned is not clear.

Objective-C/CBLDatabase.mm
snej: `@finally` is probably the simplest approach -- I hadn't thought of that.

Objective-C/CBLDatabase.mm
snej: Hmm, I just wrote a unit test of calling `c4_put` to save a deletion revision to a nonexistent doc, and it returns kC4ErrorNotFound.

snej: And `c4_update` of a purged document returns `kC4ErrorConflict`. What are the details of the call you make that returns `kc4InvalidArgument`?

pasin:
@snej The PR is ready for re-review. Could you please review the update? Thank you very much.

pasin:
@snej I renamed the ConcurrencyControl enum values per latest changes in spec. Could you re-review the PR? Thanks!

Objective-C/CBLDatabase.mm
```diff
@@ -832,72 +832,70 @@ - (BOOL) saveDocument: (CBLDocument*)document
...
         }
-        
-        if (curDoc)
-            c4doc_free(curDoc);
-        
-        if (!transaction.commit()) {
-            c4doc_free(newDoc);
-            return convertError(transaction.error(), outError);
+        @finally {
+            if (curDoc)
```
snej: FYI, these `if` tests aren't necessary; all the `c4*free` and `c4*close` functions allow null arguments.

Objective-C/CBLDatabase.mm
```diff
@@ -235,6 +235,12 @@ - (BOOL) purgeDocument: (CBLDocument*)document error: (NSError**)error {
                 if (!transaction.commit())
                     return convertError(transaction.error(), error);
                 
+                // Reload c4doc, still preserve the document data:
```
snej: How can the doc be reloaded after it's been purged? It should be completely gone from the database. If not, something's wrong...

Objective-C/Tests/ConcurrentTest.m
```diff
@@ -274,9 +274,9 @@ - (void) testConcurrentPurgeDocs {
     
     [self concurrentRuns: kNConcurrents waitUntilDone: YES withBlock: ^(NSUInteger rIndex) {
         for (CBLDocument* doc in docs) {
-            NSError* error;
-            if (![self.db purgeDocument: doc error: &error])
-                AssertEqual(error.code, 404);
+            [self mayHaveException: NSInvalidArgumentException in: ^{
```
snej: It seems wrong that this can throw an exception — that indicates a fatal programmer mistake, not a runtime error. Where does that exception come from? Can we change it to an NSError?

Objective-C/Tests/DatabaseTest.m
```diff
@@ -682,10 +682,9 @@ - (void) testDeleteNonExistingDoc {
     AssertNil([self.db documentWithID: doc1a.id]);
     
     // Delete doc1a, no-ops:
-    Assert([self.db deleteDocument: doc1a error: &error]);
-    AssertNil(error);
-    AssertEqual(self.db.count, 0u);
-    AssertNil([self.db documentWithID: doc1a.id]);
+    [self expectException: @"NSInvalidArgumentException" in: ^{
+        [self.db deleteDocument: doc1a error: nil];
```
snej: Again, throwing an exception is the wrong behavior here, as far as I can tell.

Objective-C/Tests/DatabaseTest.m
```diff
@@ -918,8 +917,9 @@ - (void) testPurgeSameDocTwice {
     AssertEqual(0, (long)self.db.count);
     
     // Purge Doc second time
-    [self purgeDocAndVerify: doc];
-    AssertEqual(0, (long)self.db.count);
+    [self expectException: @"NSInvalidArgumentException" in: ^{
```
snej: Ditto.

Objective-C/Tests/ConcurrentTest.m
pasin: Agreed. I have been back and forth on this. I will change to return NOT-FOUND error. @borrrden probably will review this again for the error review task.

Objective-C/CBLDatabase.mm
pasin: Currently the mostExist is set to false. It will get a new doc.
```
auto newDoc = c4doc_get(_c4db, bDocID, false, &err);
```

pasin:
@snej I have pushed update per your comment. Please also see my comment about reloading the doc after purge above. Thanks.

Objective-C/CBLDatabase.mm
pasin: Or probably I could just set the c4doc to nil? This is probably the same thing.

pasin:
Instead of re-getting the new doc after purge, I reset it to nil as it has the same effect. Please re-review. Thanks!

pasin:
@snej do you have a chance to re-review? Can we merge this?

Objective-C/CBLDatabase.mm
```diff
@@ -193,12 +217,9 @@ - (BOOL) purgeDocument: (CBLDocument*)document error: (NSError**)error {
...
             return NO;
         
-        if (!document.revID) {
-            [NSException raise: NSInvalidArgumentException
-                        format: @"Do not allow to purge a newly created document "
-                                 "that has not been saved into the database."];
-            return NO;
-        }
+        if (!document.revID)
+            return createError(CBLErrorInvalidParameter,
```
snej: Shouldn't this be CBLErrorNotFound?

Objective-C/CBLDatabase.mm
```diff
@@ -790,13 +812,83 @@ - (void) freeC4DB {
...
+
+
+- (BOOL) saveDocument: (CBLDocument*)document
+   concurrencyControl: (CBLConcurrencyControl)concurrencyControl
+           asDeletion: (BOOL)deletion
+                error: (NSError**)outError
+{
+    if (deletion && !document.revID)
+        return createError(CBLErrorInvalidParameter,
+                           @"Document doesn't exist in the database.", outError);
```
snej: Shouldn't this be CBLErrorNotFound?

Objective-C/CBLDatabase.mm
```diff
@@ -790,13 +812,83 @@ - (void) freeC4DB {
...
+                        return convertError(err, outError);
+                    } else if ((curDoc->flags & kDocDeleted) != 0) {
+                        [document replaceC4Doc: [CBLC4Document document: curDoc]];
+                        curDoc = nil;
+                        return YES;
+                    }
+                }
+                
+                // Save changes on the current branch:
+                assert(curDoc);
```
snej: This doesn't take into account an error getting curDoc (unless `deletion` is true.) For example, saving a doc when another thread has purged it. Or some I/O error on the database, perhaps.

Objective-C/CBLDatabase.mm
pasin: It's 50/50 between CBLErrorInvalidParameter or CBLErrorNotFound. For example, the document is actually in the database but we don't allow this because the document should be gotten from the database first.

Objective-C/CBLDatabase.mm
pasin: Same reason as above.

Objective-C/CBLDatabase.mm
pasin: That's true.

Objective-C/CBLDatabase.mm
pasin: I created https://github.com/couchbase/couchbase-lite-ios/issues/2075 to track this and get opinion from @borrrden .

PR #2069, opened 2018-02-20
Fixed: Build failed due to the new Xcode 9.3 compiler error

pitiphong-p:
Casting an NSInteger argument for String Format to long as suggested in https://developer.apple.com/library/content/documentation/Cocoa/Conceptual/Strings/Articles/formatSpecifiers.html


pasin:
Thanks for the PR.

PR #2077, opened 2018-02-27
MB-28145 Update EE license file for CBL iOS

kimtrang:
@pasin This is to update the latest EE license file for all of mobile products.  I've added the check if building on jenkins, then get the official license file from the build repo: https://github.com/couchbase/build/tree/master/license/couchbase-lite.  I didn't want to break the existing dev build incase dev doesn't have build repo checkout.

### DIFF
--- a/Scripts/Support/License/LICENSE_community.txt
+++ b/Scripts/Support/License/LICENSE_community.txt
@@ -6,7 +6,7 @@ IMPORTANT-READ CAREFULLY: BY CLICKING THE “I ACCEPT” BOX OR INSTALLING, DOWN
 
 Couchbase ®
 http://www.couchbase.com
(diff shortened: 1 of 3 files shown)


pasin:
@kimtrang are the LICENSE files in couchbase-lite-ios repo and the build repo are the same? Is there a reason to keep the license in the build repo? Can we have them only in couchbase-lite-ios repo so we don't need to keep them in-sync?

kimtrang:
I would prefer to have the License files in the build repo rather than the product source repo for maintenance purpose.  If we need to update it, only need to update in one place, rather than to update multiple places.  If you don't need to run this build script on your end, then can definitely remove the if/else check.  I wasn't sure if it's going to break things in dev side.

Issue #2078, opened 2018-02-28
CBLReplicator can leak websocket connections when "Operation timed out" errors occur


tleyden:
## Overview

In 
https://github.com/couchbase/couchbase-lite-ios/issues/1822, I noticed `NSPOSIXError[60, "Operation timed out"` messages in the logs and decided to look into it.

From what I can tell, when these websocket errors occur, it can cause CBL to [leak websocket connections](https://github.com/couchbase/couchbase-lite-ios/issues/1822#issuecomment-369050119).

## Details

Based on these logs:

```
2018-02-27 11:31:46.228996-0800 Todo[61139:9301843] CouchbaseLite Sync Info: CBLReplicator[<*> URL[ws://localhost:4984/todo]]: Transient error (Operation timed out); will retry in 2 sec...
2018-02-27 11:31:46.229287-0800 Todo[61139:9301843] CouchbaseLite Sync Info: CBLReplicator[<*> URL[ws://localhost:4984/todo]] is offline, progress 484400/484402, error: Error Domain=NSPOSIXErrorDomain Code=60 "Operation timed out" UserInfo={NSLocalizedDescription=Operation timed out}
2018-02-27 11:31:46.229520-0800 Todo[61139:9301843] [Todo] Replicator: 484400/484402, error: Error Domain=NSPOSIXErrorDomain Code=60 "Operation timed out" UserInfo={NSLocalizedDescription=Operation timed out}
2018-02-27 11:31:48.229305-0800 Todo[61139:9301843] CouchbaseLite Sync Info: CBLReplicator[<*> URL[ws://localhost:4984/todo]]: Retrying...
2018-02-27 11:31:48.229845-0800 Todo[61139:9301843] CouchbaseLite DB Info: Opening DataFile /Users/tleyden/Library/Developer/CoreSimulator/Devices/FD2A94F1-D45E-404E-812F-5C855E741313/data/Containers/Data/Application/AF4F1E29-84CC-4144-8BD6-C1E487ACDB0F/Library/Application Support/CouchbaseLite/user1.cblite2/db.sqlite3
2018-02-27 11:31:48.230640-0800 Todo[61139:9301843] CouchbaseLite BLIP Info: {Connection#3}==> litecore::blip::Connection ->ws:localhost:4984/todo/_blipsync
2018-02-27 11:31:48.230793-0800 Todo[61139:9301843] CouchbaseLite BLIP Info: {Connection#3} Opening connection...

```

[...]

PR #2080, opened 2018-03-01
Fix socket leaks and increase idle timeout (#2078)

pasin:
- Ensured to close read/write socket when an error occurs; the stream task might no close the socket when an error such as operation timed out (60) occurs.

- Increased the default read/write timeout to the value (320 seconds) which is greater than the default heartbeat of 300 seconds.

#2078

### DIFF
--- a/Objective-C/Internal/Replicator/CBLWebSocket.mm
+++ b/Objective-C/Internal/Replicator/CBLWebSocket.mm
@@ -62,7 +62,6 @@ static constexpr NSTimeInterval kIdleTimeout = 320.0;
     size_t _receivedBytesPending, _sentBytesPending;
     CFAbsoluteTime _lastReadTime;
     BOOL _requestedClose;
-    BOOL _closeOnError;
 }
 
 
@@ -357,9 +356,7 @@ static void doCompletedReceive(C4Socket* s, size_t byteCount) {
     {
         self->_receiving = false;
         self->_lastReadTime = CFAbsoluteTimeGetCurrent();
-        if (error)
-            [self didCloseWithError: error];
-        else {
+        if (![self checkError: error]) {
             self->_receivedBytesPending += data.length;
             CBLLogVerbose(WebSocket, @"<<< received %zu bytes%s [now %zu pending]",
                           (size_t)data.length, (atEOF ? " (EOF)" : ""), self->_receivedBytesPending);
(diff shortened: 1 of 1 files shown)


Objective-C/Internal/Replicator/CBLWebSocket.mm
```diff
@@ -41,8 +41,10 @@
 
 static constexpr size_t kMaxReceivedBytesPending = 100 * 1024;
 static constexpr NSTimeInterval kConnectTimeout = 15.0;
-static constexpr NSTimeInterval kIdleTimeout = 300.0;
 
+// The value should be greater than the heartbeat to avoid read/write timeout;
+// the current default heartbeat is 300 sec:
+static constexpr NSTimeInterval kIdleTimeout = 320.0;
```
tleyden: how hard would it be to reference the heartbeat from here and then multiply it by a ratio?  I'm guessing hard, otherwise you would have done it .. but just checking.

Objective-C/Internal/Replicator/CBLWebSocket.mm
```diff
@@ -467,10 +470,11 @@ - (void)URLSession:(NSURLSession *)session writeClosedForStreamTask:(NSURLSessio
 }
 
 - (void) streamClosed: (BOOL)isWrite {
+    BOOL expectedClose = _requestedClose || _closeOnError;
     CBLLog(WebSocket, @"CBLWebSocket %s stream closed%s",
            (isWrite ? "write" : "read"),
-           (_requestedClose ? "" : " unexpectedly"));
-    if (!_requestedClose) {
+           (expectedClose ? "" : " unexpectedly"));
```
tleyden: By saying "unexpectedly" at the end, it feels easy to overlook.  What about somehow putting "WARNING" in the message in this case to make it more apparent?

Maybe:

```
 (expectedClose ? "" : " WARNING: connection closed unexpectedly"));
```

although .. if it happens often, then it's extra noise.

Objective-C/Internal/Replicator/CBLWebSocket.mm
```diff
@@ -60,6 +62,7 @@ @implementation CBLWebSocket
     size_t _receivedBytesPending, _sentBytesPending;
     CFAbsoluteTime _lastReadTime;
     BOOL _requestedClose;
+    BOOL _closeOnError;
```
snej: `_closingOnError` would be a clearer name.

Objective-C/Internal/Replicator/CBLWebSocket.mm
snej: The heartbeat happens down inside LiteCore and isn't known at this level. CBLWebSocket just gets told to send PING messages periodically.

Objective-C/Internal/Replicator/CBLWebSocket.mm
snej: It'll get reported later in a log message starting `CBLWebSocket CLOSED WITH ERROR`.

pasin:
@snej and @tleyden I re-fixed based on @snej's suggestion by canceling the task in the `-checkError:`.

pasin:
Just re-push as I accidentally removed the logging in streamClosed: method.

pasin:
Sorry, just re-push again as I forgot to modify the idle timeout back after changing it for testing.

PR #2090, opened 2018-03-06
Fixed merging of conflicts involving a deletion

snej:
Fixes resolution of conflicts when a doc has been deleted locally but updated on the server.

* Updated LiteCore to get its half of the fix.
* When creating a merge revision to resolve a conflict, make it a
  deletion when necessary (using the new param of c4doc_resolveConflict.)
* Fixed a bug where CBLC4Document's properties referred to the current
  revision, not the selected revision (which is different when a
  conflict is being resolved.)
* Re-enabled testPullConflictDeleteWins_SG, now that it passes.

For couchbase/couchbase-lite-core#433


pasin:
CC: @borrrden @hideki

Issue #2089, opened 2018-03-05
testStopReplicatorAfterOffline_SG fails on iOS 10.2 simulator


pasin:
The test passes on 11.x simulators.

BT

```
thread #1, queue = 'com.apple.main-thread'
    frame #0: 0x0000000114f517c2 libsystem_kernel.dylib`mach_msg_trap + 10
    frame #1: 0x0000000114f50cdc libsystem_kernel.dylib`mach_msg + 60
    frame #2: 0x000000010ddc8ba4 CoreFoundation`__CFRunLoopServiceMachPort + 212
    frame #3: 0x000000010ddc8021 CoreFoundation`__CFRunLoopRun + 1345
    frame #4: 0x000000010ddc7884 CoreFoundation`CFRunLoopRunSpecific + 420
    frame #5: 0x000000010d7b64ad XCTest`-[XCTWaiter waitForExpectations:timeout:enforceOrder:] + 545
    frame #6: 0x000000010d7674d5 XCTest`-[XCTestCase(AsynchronousTesting) _waitForExpectations:timeout:enforceOrder:handler:] + 132
    frame #7: 0x000000012214b6d1 CouchbaseLiteTests`-[ReplicatorTest testStopReplicatorAfterOffline_SG](self=0x000060700002a9c0, _cmd="testStopReplicatorAfterOffline_SG") at ReplicatorTest.m:848
    frame #8: 0x000000010ddc544c CoreFoundation`__invoking___ + 140
    frame #9: 0x000000010ddc52d1 CoreFoundation`-[NSInvocation invoke] + 289
    frame #10: 0x000000010d773e30 XCTest`__24-[XCTestCase invokeTest]_block_invoke + 591
    frame #11: 0x000000010d7bd17e XCTest`-[XCUITestContext performInScope:] + 183
    frame #12: 0x000000010d773bd6 XCTest`-[XCTestCase invokeTest] + 141
    frame #13: 0x000000010d774b97 XCTest`__26-[XCTestCase performTest:]_block_invoke.369 + 42
    frame #14: 0x000000010d7c1f25 XCTest`+[XCTContext runInContextForTestCase:block:] + 163
    frame #15: 0x000000010d774533 XCTest`-[XCTestCase performTest:] + 608
    frame #16: 0x000000010d770539 XCTest`__27-[XCTestSuite performTest:]_block_invoke + 363
[...]

PR #2093, opened 2018-03-06
Fix #3089 Cancel NSURLSessionStreamTask could cause a crash in iOS 10

pasin:
- Instead of cancel, close read/write socket instead for the case of an error occurred to avoid the crash in iOS 10.

- In readClosedForStreamTask: and writeClosedForStreamTask:, also validate the task if the task is still the current task. This is required because we canceled the current task is in - receivedHTTPResponse due to a redirect or auth challenge. In iOS 10 (not iOS 11), after canceling the task, the -readClosedForStreamTask: and -writeClosedForStreamTask: delegate will be called. Without checking the -streamClosed: method will be called, and there could be a race condition which causes the new task to be canceled in the -streamClosed: method. In general we don't want -streamClosed: to be called for this case.

- We should either close the sockets or cancel the task in didCloseWithCode:reason: as well otherwise the sockets will leak. To be safe, I close the sockets instead of canceling the task.

- Enabled testStopReplicatorAfterOffline_SG and ensure that the replicator will be stopped after running the replicator even though the test failed.

#2089


Objective-C/Internal/Replicator/CBLWebSocket.mm
```diff
@@ -457,14 +456,18 @@ - (void) URLSession: (NSURLSession *)session
 
 - (void)URLSession:(NSURLSession *)session readClosedForStreamTask:(NSURLSessionStreamTask *)task
 {
-    [self streamClosed: NO];
+    if (task == _task)
```
snej: It'd be cleaner to move this test (and the one in the next method) into -streamClosed:, since we're trying to share the logic for both the read and write streams.

Objective-C/Internal/Replicator/CBLWebSocket.mm
```diff
@@ -538,6 +543,12 @@ - (void) didCloseWithError: (NSError*)error {
 }
 
 
+- (void) closeTask {
+    [_task closeWrite];
+    [_task closeRead];
+}
```
snej: Should you set `_task = nil` here, since it's now closed? Then you don't have to set it to nil after the various calls to -closeTask.

pasin: I tried to do this, but the changing scope is getting bigger. There is some complication especially with -didCloseWithError: that shouldn't be called when _task is nil. I think we should do this in post 2.0 after we have enough time to test the change to cover all edge cases. I put todo in the code.

PR #2132, opened 2018-04-03
Refactored change notification into CBLChangeNotifier

snej:
I did this because I need listeners for the upcoming CBLReplicationListener, and I don't want to copy/paste the listener-management code again :)

* CBLChangeNotifier manages a set of CBLChangeListenerTokens, and broadcasts change objects to them.
* Made CBLChangeListenerToken more self-contained by having it call the block itself.
* Replaced the listener-handling code in CBLDatabase, CBLReplicator and CBLLiveQuery with CBLChangeNotifier.
* Created CBLDocumentChangeNotifier to manage per-document change notification, so it can keep track of the C4DocumentObserver instead of CBLDatabase.
* Removed some old incorrect logic about not posting document-changed notifications if the database is in a transaction.


pasin:
I updated license headers and merged.

PR #2143, opened 2018-04-17
[docs] .and() clause

jamiltz:
I couldn't find anywhere else to document that. It's quite a low level detail to go into the general docs. May need confirmation from Jens on type coercion if you are not sure about what I wrote.

Taken from https://issues.couchbase.com/browse/DOC-3218


snej:
Shouldn't the same note be added to the OR and NOT operators?

jamiltz:
Sure, added the same to OR and NOT.

PR #2169, opened 2018-06-14
Fix CBLWebSocket proxy bug using Apple workaround

snej:
CFNetwork in iOS 11 has a bug where multiple pending writes to an NSURLSessionStreamTask
can be written in the wrong order (but only if going through an HTTP proxy...!)
To work around this, we're ensuring that we only have one outstanding write at a time,
as suggested by Apple’s Quinn.


snej:
FYI, the bug was filed with Apple as 41140274.

PR #2170, opened 2018-06-18
Proxy fixes: TLS, redirect, test setup

snej:
* When configuring NSStream TLS settings after making a proxy connnection, set the server name property to the actual server hostname, otherwise cert verification will expect the proxy’s hostname and fail.
* Handle redirect/auth status returned from proxy and retry.
* ReplicatorTest should set up proxy overrides at init time, not in -remoteEndpointWithName:, because not all the tests call that method.

### DIFF
--- a/CouchbaseLite.xcodeproj/xcshareddata/xcschemes/CBL ObjC.xcscheme
+++ b/CouchbaseLite.xcodeproj/xcshareddata/xcschemes/CBL ObjC.xcscheme
@@ -28,8 +28,7 @@
       selectedLauncherIdentifier = "Xcode.DebuggerFoundation.Launcher.LLDB"
       enableAddressSanitizer = "YES"
       enableASanStackUseAfterReturn = "YES"
-      language = ""
-      shouldUseLaunchSchemeArgsEnv = "YES">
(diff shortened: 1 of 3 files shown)


CouchbaseLite.xcodeproj/xcshareddata/xcschemes/CBL ObjC.xcscheme
```diff
@@ -51,6 +50,23 @@
...
+            key = "CBL_TEST_PROXY_HOST"
+            value = "127.0.0.1"
+            isEnabled = "NO">
+         </EnvironmentVariable>
+         <EnvironmentVariable
+            key = "CBL_TEST_PROXY_PORT"
+            value = "8888"
+            isEnabled = "NO">
+         </EnvironmentVariable>
+      </EnvironmentVariables>
```
pasin: @snej should these variable get checked-in?

snej: To make it easy to re-enable them, without having to look up the names in the source code (as I've had to do many times in the past week!) They're not enabled, so they do nothing unless you check the box.

Issue #2173, opened 2018-06-19
Replication stops when sg restarts


sridevi-15:
### Environment
* Version:   2.0.4-5
* Client OS: iOS

### Steps to reproduce : 
1. Create docs in cbl (around 1000 is better to test )
2. Start push_pull replication with continous True 
3. While replication happening, restart sync-gateway
4. Verify all docs got replication to sync-gateway

### Result : 
1. Only few docs , out of 1000 docs, around 200 docs gets replicated and then stops

### Expected : 
All docs should get replicated.

### Test : 
https://github.com/couchbaselabs/mobile-testkit/blob/master/testsuites/CBLTester/CBL_Functional_tests/TestSetup_FunctionalTests/test_replication.py#L2105
Test name is : test_CBL_push_pull_with_sg_down

### Logs : 
cbl logs : 
https://gist.github.com/sridevi-15/26057b981d4f2fc77891bda877c552eb

pasin:
Could you test with http://latestbuilds.service.couchbase.com/builds/latestbuilds/couchbase-lite-ios/2.0.4/8/?

pasin:
@sridevi-15 reported another error which also caused the replicator to stop. Here is the log:
https://gist.githubusercontent.com/sridevi-15/ed542ed79380e1289bb736b2a012e679/raw/e8a92315c2ae43fc31e4be4a4d67772194fa3fcd/cbl%2520logs2

It seems like this scenario of this case is:
1. <CBLWebSocket: 0x604000157970>: EndEncountered on read stream
2. <CBLWebSocket: 0x604000157970>: Disconnect
3. CBLWebSocket CLOSED
4. WS WARNING: {C4SocketImpl#1} Unexpected or unclean socket disconnect! (reason=WebSocket status, code=1000)
5. BLIP Info: {Connection#1} Closed with WebSocket status 1006: (null)
6. Sync Info: {Repl#1} Connection closed with WebSocket status 1006: "(null)" (state=2)
7. Sync ERROR: {Repl#1} Got LiteCore error: connection closed abnormally (6/1006)
8. CBLReplicator[<*> URL[ws://192.168.33.11:4984/db]] is busy, progress 70968/1613076, error: Error Domain=CouchbaseLite Code=11006 "connection closed abnormally" UserInfo={NSLocalizedDescription=connection closed abnormally} --> This is strange that the status is busy.
[...]

PR #2174, opened 2018-06-20
Close with ECONNRESET instead of null when the connection is reset by peer

pasin:
* The NSStreamEventEndEncountered event could be a result from either socket close requested by C4Socket or connection reset by peer.  For the latter case, we should call c4socket_closed() with ECONNRESET instead of null.

* When the stream is ended, check whether the C4Socket requested to close the connection or not. If the close wasn’t requested, use ECONNRESET to acknowledge the close when calling c4socket_closed().

#2173

### DIFF
--- a/Objective-C/Internal/Replicator/CBLWebSocket.mm
+++ b/Objective-C/Internal/Replicator/CBLWebSocket.mm
@@ -90,6 +90,7 @@ struct PendingWrite {
     bool _gotResponseHeaders;
     BOOL _connectingToProxy;
     BOOL _connectedThruProxy;
+    BOOL _closedRequest;
 }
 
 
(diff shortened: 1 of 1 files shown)


snej:
I don't think this is correct, because NSStream would have reported a real connection-reset as an error. What happened here seems to be that the remote host (SG) closed the socket (cleanly), but without sending a WebSocket close message.

CBLWebSocket shouldn't need to change what it's doing; LiteCore will recognize that the socket was closed without a close message, and treat that as an abnormal close (status 1006).

pasin:
Does this mean that we should treat 1006 as a transient error to allow the replicator to retry?

pasin:
In 2.1 CBLWebSocket, we use ECONNRESET for the case that the read/write steam is closed without close request from C4Socket. I have understood that it is the same case.

pasin:
I'm closing this PR as we fixed this by adding the websocket abnormal close (1006) in the transient error list in LiteCore.

borrrden:
Wait what?  I treat all unknown errors as 1006 in the .NET codebase, this will make all of them transient!

snej:
What sort of unknown errors? There aren't any more specific codes for them? If it's a networking-level error, it's probably better to use one of the POSIX error codes.

borrrden:
"Unknown errors" means exceptions that haven't been specifically mapped to an error code, usually because there is not really a master list of stuff that happens and exceptions that get thrown and I'm going off of what I have found so far happens.  If something else happens, it needs to get translated into a domain and number for LiteCore for a "default" bad value.

Issue #2179, opened 2018-06-26
HTTP redirects are not handled


raghusarangapani:
### Version
2.0.4-10 on iOS 11.4 iPhone

### Steps
- Set up 1 CBS and 2 SGs.
- Set up a squid proxy to redirect incoming requests from SG1 to SG2.
- Verify that the redirect works
- Install the iOS test server and start a replication with SG1.

### Actual results
Replication test fails with the below error
```
>                   raise Exception("Error while replicating", err)
E                   Exception: ('Error while replicating', 'Optional(Error Domain=CouchbaseLite Code=11008 "OK" UserInfo={NSLocalizedDescription=OK})')
```
### Test setup
CBS - 10.17.6.122
SG1 - 10.17.6.121
SG2 - 10.17.6.141

### Squid redirects
```
» http --proxy=http:http://10.17.6.118:8080 10.17.6.121:4985
HTTP/1.1 302 Found
Connection: keep-alive
Content-Length: 0
Date: Tue, 26 Jun 2018 22:31:05 GMT
Location: http://10.17.6.141:4985/
Server: squid/3.5.20
Via: 1.1 localhost.localdomain (squid/3.5.20)
X-Cache: MISS from localhost.localdomain
X-Cache-Lookup: MISS from localhost.localdomain:8080

~ » http --proxy=http:http://10.17.6.118:8080 10.17.6.121:4984
HTTP/1.1 302 Found
Connection: keep-alive
Content-Length: 0
Date: Tue, 26 Jun 2018 22:31:16 GMT
Location: http://10.17.6.141:4984/
Server: squid/3.5.20
Via: 1.1 localhost.localdomain (squid/3.5.20)
X-Cache: MISS from localhost.localdomain
X-Cache-Lookup: MISS from localhost.localdomain:8080
```
### CBL Logs
https://gist.github.com/raghusarangapani/e3f1faa8ddc8639cdeb4a1c5887b0396

### Squid config
https://gist.github.com/raghusarangapani/c7f14f612ebafe27b0635647b7d823fb

### php redirect program
```
#!/usr/bin/php
<?php

$temp = array();

// Extend stream timeout to 24 hours
stream_set_timeout(STDIN, 86400);

while ( $input = fgets(STDIN) ) {
  // Split the output (space delimited) from squid into an array.
  $temp = split(' ', $input);

  // Set the URL from squid to a temporary holder.
  $output = $temp[0] . "\n";

[...]

PR #2190, opened 2018-08-02
Fix HTTP redirect not working when using client proxy

pasin:
* When HTTP redirect happens, if using proxy, need to send a CONNECT request to the redirected host as well.

* The _logic.useProxyCONNECT is reset to NO after the CONNECT response is received. When sending the second CONNECT request for the redirected HOST, need to ensure that _logic.useProxyCONNECT is set to YES, otherwise the CONNECT request will not be sent.

* Moved `_logic.useProxyCONNECT = YES` from -initWithURL:c4socket:options:  to -_connect method where the CONNECT request will be sent.

#2179


snej:
Wow, that was a small fix :)

PR #2219, opened 2018-10-19
Allow creating Query from JSON, and getting JSON [API CHANGE]

snej:
Added to CBLQuery:
```
  - initWithDatabase:JSONRepresentation:
  @property (nonatomic, readonly) NSData* JSONRepresentation;
```
(API is provisional. In particular, should there be a factory method
on CBLDatabase, instead of an initializer?)

Restructured the implementation to avoid dependencies on the
query-builder objects (CBLQuerySelectResult, etc.) Instead, the
existing initializer translates to JSON immediately and only stores
JSON. The column names are obtained from a new LiteCore function
(c4query_columnTitle) instead of the builder objects.

Removed some unused code: the properties for the builder objects,
and the -encodeExpressions: method.

couchbaselabs/couchbase-lite-apiv2#212

### DIFF
--- a/Objective-C/CBLQuery.h
+++ b/Objective-C/CBLQuery.h
@@ -18,6 +18,7 @@
 //
 
 #import <Foundation/Foundation.h>
+@class CBLDatabase;
 @class CBLQueryParameters;
 @class CBLQueryResultSet;
 @class CBLQueryChange;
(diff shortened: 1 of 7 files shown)


pasin:
I like the idea of allowing to create the query from JSON. But I would like to get pass this with @rajagp as we also need some documentation about JSON query format when we release.

rajagp:
Agree that this is a very useful API to support. 
We should leave this as an undocumented API that will be only be available on-request. In general, we want to encourage users to use the type-safe QueryBuilder API so the usage of the API should be more of an exception - mainly to facilitate support of platform bindings that we do not support natively.

We should also make it very clear in the API definition in code (Need to ensure its not picked up during doc generation ) that the JSON parse tree can change at any point.  This will give us the flexibility to change the JSON parse tree even if it is at the expense of breaking backwards compatibility.

snej:
I see a few main use cases for this:

1. To support bindings on top of CBL, e.g. JavaScript. (See the [original email thread](https://github.com/couchbaselabs/couchbase-lite-apiv2/issues/212).)
2. To allow queries to be persisted: an app can create a query once, save it as a blob, then reconstitute it in the future. This would allow for GUI tools that create queries that can be saved as app resources, similar to how developers create fetch requests in Xcode for CoreData.
3. As a workaround to allow use of query features that we don't yet have API bindings for -- this is actually why I implemented this feature now, so that there's some way to start trying out new post-2.1 query features.

snej:
I made further changes to remove CBLPredicateQuery and associated classes; that class can be made an external utility if CBLQuery supports creation from JSON. I won't push it to this branch yet, though, to avoid mixing it up with the present commit.

geraldapeoples:
Are there any plans to enable couchbase lite queries using the n1ql query language? The use of the querybuilder seems very, very complicated compared to "SELECT model from mydb t where t.type = 'vehicle'"

pasin:
I just noticed that the column names are also changed. Not sure if this is a breaking API or not. I think I need to revert the merge for now.

PR #2272, opened 2018-12-09
Fixed #2270 : Convert from sec to msec and vice versa

### LINKED ISSUE #2270: Document Expiry doesn't seem to set the expiration timestamp in milliseconds
From the code here https://github.com/couchbase/couchbase-lite-ios/blob/master/Objective-C/CBLDatabase.mm#L540-L547, we need to convert from second to milliseconds before calling c4doc_setExpiration().

pasin:
* Convert from sec to msec when setting expiration date
* Convert from msec to sec when getting expiration date

### DIFF
--- a/Objective-C/CBLDatabase.mm
+++ b/Objective-C/CBLDatabase.mm
@@ -43,6 +43,8 @@ using namespace fleece;
 
 #define kDBExtension @"cblite2"
 
+#define msec 1000.0
+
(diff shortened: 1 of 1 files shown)


pasin:
@jayahariv I have missed this when reviewing the PR. I'm merging this now as the fix is very simple and I need to merge the fix to fix Jenkins build issue. Please review this when you have time. Thanks!

Issue #2242, opened 2018-11-30
Implement date/time function


pasin:
PRD:

https://docs.google.com/document/d/1OFw2eE9EDpApMPD9H_AR86oVZ-DesCg0CHoz6BKQn5s

--------
* Version:   Iridium

pasin:
Please file an issue on LiteCore.

snej:
Did you find this in a unit test? Those kind of tests should be in LiteCore itself.

snej:
Fixed in LiteCore.

PR #2276, opened 2018-12-11
Feature/date time function

jayahariv:
* add ObjC and Swift implementations to create query functions for str_to_millis, str_to_utc, millis_to_str, and millis_to_utc.
* made consistent notes with dotnet.
* Add StringToMillis & StringToUTC Unit Tests for different timezone, different date formats for ObjC.
* helper methods kept in the bottom.
* kept the basic date formatter as macros.

fix: #2242

### DIFF
--- a/Objective-C/Tests/DateTimeQueryFunctionTest.m
+++ b/Objective-C/Tests/DateTimeQueryFunctionTest.m
@@ -28,126 +28,105 @@
 
 @implementation DateTimeQueryFunctionTest
 
+
 #pragma mark - Tests
 
-- (void) testWrongDateFormatStringToMillis {
-    NSError* error;
-    NSDate* now = [NSDate date];
-    NSString* key = @"dateString";
-    CBLMutableDocument* doc = [self createDocument];
-    [doc setValue: [self getLongStyleDateString: now] forKey: key];
-    [self saveDocument: doc];
-    AssertNil(error);
+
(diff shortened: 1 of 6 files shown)


jayahariv:
@pasin please hold the review. I will re-write the test with dateForKey instead of directly checking for stringForKey. Soon I can add those test changes.

jayahariv:
@pasin please check this when you are free. I have updated with the refactored and date comparing test cases. Now it is more modularized. 

Also this revision contains only test cases for ObjC, for StringToMillis, StringToUTC. If this modularization is fine, i can finish this for MillisToString and MillisToUTC.

Objective-C/Tests/DateTimeQueryFunctionTest.m
```diff
@@ -0,0 +1,225 @@
...
+    
+    [self validateMillisToUTC: [NSNumber numberWithDouble: 1540319581999]
+                       result: @"2018-10-23T18:33:01.999Z"];
+}
+
+
+#pragma mark - Helper Methods
+
+
+- (void) validateStringToMillis: (NSString*)input result: (nullable NSNumber*)millis {
```
jayahariv: I can make these validate to single function, but that will make things complicated. And pass in an enum or to differentiate between the methods, like string_to_millis, etc. Now a few lines are duplicated.

Issue #2267, opened 2018-12-08
Verify TTL Test Scenarios are covered


borrrden:
As laid out in [this document](https://docs.google.com/spreadsheets/d/1RK87u2pmkhiF5IvuGW9LO8viibw2aR9NJ6slFTaTvLs/) the following cases must be in the developer test suite:

- [x] Getting / Setting expiration via API
- [x] Document purged after expiration date
- [x] Expired documents do not show up in query (even a deleted docs one)
- [x] Removing expiration causes a document to not expire
- [x] Expired documents are not replicated (testing lack of replication, which is why this one is here)
- [x] Expiring a document immediately via a call to `setDocumentExpiration(<now>)`

jayahariv:
i guess this scenario will be handled in lite-core, but should we need to check?
- set expiry and then purge document, what happens when the timer is fired

borrrden:
The timer should make a call to LiteCore to begin the purge.

jayahariv:
Pass criteria should be, 'not to raising any warnings, or errors'. But seems like more of a lite core test case.

borrrden:
It could also be folded into the first test case.  Make two documents, expire both, and purge 1.  Verify no documents are in the DB after the expiration date.

PR #2284, opened 2018-12-15
Feature/doc expiry tests

jayahariv:
Pending Test Cases
* add test case for check purge immediately in Swift too. 
* add test case for lack of replication for expired document. 

fix: #2267

### DIFF
--- a/Objective-C/Tests/DocumentExpirationTest.m
+++ b/Objective-C/Tests/DocumentExpirationTest.m
@@ -98,8 +98,8 @@
 - (void) testDocumentPurgedAfterExpiration {
     XCTestExpectation* expectation = [self expectationWithDescription: @"Document expiry test"];
     
-    NSError* err;
     CBLDocument* doc = [self generateDocumentWithID: nil];
+    AssertNil([self.db getDocumentExpirationWithID: doc.id]);
     AssertEqual(self.db.count, 1u);
     
     // Setup document change notification
@@ -113,12 +113,12 @@
     }];
 
     // Set expiry
-    AssertNil(err);
     NSDate* expiryDate = [NSDate dateWithTimeIntervalSinceNow: 1.0];
+    NSError* err;
     Assert([self.db setDocumentExpirationWithID: doc.id expiration: expiryDate error: &err]);
     AssertNil(err);
     
-    // Wait for result, it shouldn't crash due to already purged doc
+    // Wait for result
     [self waitForExpectationsWithTimeout: 5.0 handler: nil];
     
     // Remove listener
--- a/Objective-C/Tests/DocumentExpirationTest.m
+++ b/Objective-C/Tests/DocumentExpirationTest.m
@@ -105,16 +105,16 @@
     // Setup document change notification
     id token = [self.db addDocumentChangeListenerWithID: doc.id
                                                listener: ^(CBLDocumentChange *change)
-                {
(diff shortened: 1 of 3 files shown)


Objective-C/Tests/DocumentExpirationTest.m
```diff
@@ -72,28 +72,38 @@ - (void) testSetExpirationToNonExistingDocument {
...
+
     // Set expiry
-    NSDate* expiryDate = [NSDate dateWithTimeIntervalSinceNow: 1.0];
-    NSError* err;
+    NSDate* expiryDateToPurge = [NSDate dateWithTimeIntervalSinceNow: 2.0];
+    Assert([self.db setDocumentExpirationWithID: docToPurgeBeforeExpire.id
+                                     expiration: expiryDateToPurge
+                                          error: &err]);
+    AssertNil(err);
+    NSDate* expiryDate = [NSDate dateWithTimeIntervalSinceNow: 3.0];
```
pasin: What is the purpose of re-setting the expiration from 2 seconds to 3 seconds?

Objective-C/Tests/DocumentExpirationTest.m
```diff
@@ -72,28 +72,38 @@ - (void) testSetExpirationToNonExistingDocument {
...
     Assert([self.db setDocumentExpirationWithID: doc.id expiration: expiryDate error: &err]);
     AssertNil(err);
     
-    // Wait for result
+    // purge one document
+    Assert([self.db purgeDocument: docToPurgeBeforeExpire error: &err]);
```
pasin: From the name of the test case which is `testDocumentPurgedAfterExpiration`, I understand that it tested whether the document is purged after it got expired? What do you intend to test here?

Objective-C/Tests/DocumentExpirationTest.m
pasin: oh I see, they are different docs.

Objective-C/Tests/DocumentExpirationTest.m
pasin: I understand it now. I think it's better to create a new test case to test purge before expired.

borrrden: I am the one that suggested they be the same @pasin because otherwise the only "assertions" in the test will be "no crash"

Objective-C/Tests/ReplicatorTest.m
```diff
@@ -1829,4 +1829,63 @@ - (void) testPushAndPullBigBodyDocument_SG {
     [self run: config errorCode: 0 errorDomain: nil];
 }
 
+
+- (void) testPushAndPullExpiredDocument_SG {
```
borrrden: This should be a functional test as it will not be run in an automated fashion.

jayahariv: this will be run by QA. And not run during unit tests. So @pasin suggested to keep it, as long as it's not failing, QA can test this.

Objective-C/Tests/DocumentExpirationTest.m
pasin: I think it's testing different thing. The second test is testing whether purging doc before expired is ok or not. I feel like it's better to have them in two different test cases. At the end, it's about preferences.

PR #2285, opened 2018-12-17
Check database opened before purge expired docs in background

pasin:
* Need to check whether database (_c4db) is still opened before purging the expired doc in the background.
* Add error log when purging is failed.

### DIFF
--- a/Objective-C/CBLDatabase.mm
+++ b/Objective-C/CBLDatabase.mm
@@ -1041,11 +1041,22 @@ static C4DatabaseConfig c4DatabaseConfig (CBLDatabaseConfiguration* config) {
 
 - (void) purgeExpiredDocuments {
     dispatch_async(dispatch_get_global_queue( DISPATCH_QUEUE_PRIORITY_DEFAULT, 0), ^{
-        CBLLog(Database, @"Purging expired documents...");
-        C4Error err;
-        UInt64 nPurged = c4db_purgeExpiredDocs(_c4db, &err);
-        CBLLog(Database, @"Purged %lld expired documents", nPurged);
-        [self scheduleDocumentExpiration: 1.0];
+        CBL_LOCK(self) {
+            if (!_c4db)
+                return;
+            
+            CBLLog(Database, @"Purging expired documents...");
+            C4Error c4err;
+            UInt64 nPurged = c4db_purgeExpiredDocs(_c4db, &c4err);
+            if (nPurged >= 0)
+                CBLLog(Database, @"Purged %lld expired documents", nPurged);
+            else {
+                NSError* error;
+                convertError(c4err, &error);
+                CBLWarnError(Database, @"Cannot purge expired documents: %@", error);
+            }
+            [self scheduleDocumentExpiration: 1.0];
+        }
     });
 }
 


Objective-C/CBLDatabase.mm
```diff
@@ -1041,11 +1041,22 @@ - (void) scheduleDocumentExpiration: (NSTimeInterval)minimumDelay {
 
 - (void) purgeExpiredDocuments {
     dispatch_async(dispatch_get_global_queue( DISPATCH_QUEUE_PRIORITY_DEFAULT, 0), ^{
-        CBLLog(Database, @"Purging expired documents...");
-        C4Error err;
-        UInt64 nPurged = c4db_purgeExpiredDocs(_c4db, &err);
-        CBLLog(Database, @"Purged %lld expired documents", nPurged);
-        [self scheduleDocumentExpiration: 1.0];
+        CBL_LOCK(self) {
```
jayahariv: we are always performing the purgeExpiration on Main thread, so is there a chance it is going to be thread unsafe?

Objective-C/CBLDatabase.mm
```diff
@@ -1041,11 +1041,22 @@ - (void) scheduleDocumentExpiration: (NSTimeInterval)minimumDelay {
...
+            if (!_c4db)
```
jayahariv: technically the purgeExpiry shouldn't get called if the db doesn't exists(as we are cancelling all performSelectors while closing the DB).

Objective-C/CBLDatabase.mm
```diff
@@ -1041,11 +1041,22 @@ - (void) scheduleDocumentExpiration: (NSTimeInterval)minimumDelay {
...
+            
+            CBLLog(Database, @"Purging expired documents...");
+            C4Error c4err;
+            UInt64 nPurged = c4db_purgeExpiredDocs(_c4db, &c4err);
+            if (nPurged >= 0)
+                CBLLog(Database, @"Purged %lld expired documents", nPurged);
+            else {
+                NSError* error;
+                convertError(c4err, &error);
+                CBLWarnError(Database, @"Cannot purge expired documents: %@", error);
```
jayahariv: if we warn about purge without any docs, then we might need to re-write the schedule with a 1 second delay. Or else the console will print warning when the next expiry is less than a second(and the delay is 0.0, which is basically they set an expiry Date().now).

https://github.com/couchbase/couchbase-lite-ios/blob/master/Objective-C/CBLDatabase.mm#L1029

Objective-C/CBLDatabase.mm
pasin: We call -purgeExpiredDocuments on the main queue but the actual purge is done inside the global queue (see line 1043) in the background not the main queue.

Objective-C/CBLDatabase.mm
pasin: Same as my comment above. the actual purge is done in the background and could be schedule before cancelling the performSelectors.

Objective-C/CBLDatabase.mm
pasin: We warn only when the nPurge < 0 which means there is an error.

jayahariv: yea thats right.

Objective-C/CBLDatabase.mm
jayahariv: Scheduling is done on the main thread-run loop, and only the execution in background thread. So if the cancelling is done, before execution, it should remove it. anyways, we can discuss this in person.

pasin: The cancelling is not relevant here as this is the case when the block to purge has already been scheduled on the background dispatch queue at the same time as the database is closed. Yes, we can discuss this later today.

PR #2300, opened 2019-01-03
Hold strong reference to query instance to prevent deallocation from causing crash.

eric-robinson:
If a `CBLLiveQuery`'s `_query` var in dealloced, this can cause a crash in the `dispatch_after` call because the queue pass in is nil. This can be triggered in a race condition where the live query in dealloced around the same time as a database update that would cause the live query to fire.

Issue ref: #2243.


jayahariv:
Hi @eric-robinson, 

Thanks for the PR, we have added the fix as well as another deadlock condition in this PR #2301. 

Hence closing this PR.

eric-robinson:
Great, thanks!

PR #2319, opened 2019-01-11
Exclude default log directory from backup for iOS

pasin:
* Changed the default log directory for iOS to inside the App Support directory.
* Set the default log directory to exclude from backup.
* Fixed file URL in LogTest.getLogsInDirectory().

### DIFF
--- a/Objective-C/CBLFileLogger.mm
+++ b/Objective-C/CBLFileLogger.mm
@@ -27,7 +27,9 @@
 
 #define kCBLFileLoggerDefaultMaxSize 500*1024
 
-@implementation CBLFileLogger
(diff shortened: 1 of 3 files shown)


rajagp:
@jayahariv  Regarding your comment on custom logging - there are no log files in custom logging so unclear on the concern.

jayahariv:
@rajagp I mean custom directory URL logging.

pasin:
I'm not sure if we will have the default log location anymore. This PR is probably not valid anymore anyway as we pretty much will change the API and implementation. So I'm closing the PR.

pasin:
@jayahariv Thanks for reviewing anyway.

PR #2325, opened 2019-01-14
doc: update with custom instance set info

jayahariv:
* when changing a log level of the custom instance, it should be done before setting the custom instance. 
- as the synchronize call will happen only on setCustom method.

### DIFF
--- a/Objective-C/CBLLog.h
+++ b/Objective-C/CBLLog.h
@@ -35,7 +35,7 @@ NS_ASSUME_NONNULL_BEGIN
 /** File logger writing log messages to files. */
 @property (readonly, nonatomic) CBLFileLogger* file;
 
-/** For setting a custom logger. */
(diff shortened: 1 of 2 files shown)


Objective-C/CBLLog.h
```diff
@@ -35,7 +35,7 @@ NS_ASSUME_NONNULL_BEGIN
 /** File logger writing log messages to files. */
 @property (readonly, nonatomic) CBLFileLogger* file;
 
-/** For setting a custom logger. */
+/** For setting a custom logger. Set log level for custom instance, before setting this property. */
 @property (nonatomic, nullable) id<CBLLogger> custom;
```
pasin: How about something like this?

```
Changing the log level of the assigned custom logger will require the custom logger to be reassigned so that the change can be affected.
```

Issue #2321, opened 2019-01-12
Remove log rotation change batching


borrrden:
The batching of things like the file logging max size, max rotate count, etc is prone to races and generally over complicated.  This will be achieved by moving these settings into a class that will be set so that all properties are changed at once.

PR #2326, opened 2019-01-15
Implement a new file logger API

pasin:
- Implemented a new file logger API a refactor log file related settings into a single configuration file, LogFileConfiguration.

- Updated unit tests.

#2321

### DIFF
--- a/Objective-C/CBLFileLogger.h
+++ b/Objective-C/CBLFileLogger.h
@@ -26,6 +26,11 @@ NS_ASSUME_NONNULL_BEGIN
 /**
  File logger used for writing log messages to files. To enable the file logger,
  setup the log file configuration and specifiy the log level as desired.
+ 
+ It is important to configure your CBLLogFileConfiguration object appropriately before setting
+ to a logger. Logger make a copy of the configuration settings you provide and use those settings
+ to configure the logger. Once configured, the logger object ignores any changes you make to the
+ CBLLogFileConfiguration object.
  */
 @interface CBLFileLogger : NSObject <CBLLogger>
 
--- a/Objective-C/CBLFileLogger.h
+++ b/Objective-C/CBLFileLogger.h
@@ -19,41 +19,28 @@
 
 #import <Foundation/Foundation.h>
 #import "CBLLogger.h"
+@class CBLLogFileConfiguration;
 
 NS_ASSUME_NONNULL_BEGIN
 
 /**
- File logger used for writing log messages to files. The binary file format will be used by default.
- To change the file format to plain text, set the usePlainText property to true.
+ File logger used for writing log messages to files. To enable the file logger,
+ setup the log file configuration and specifiy the log level as desired.
  */
 @interface CBLFileLogger : NSObject <CBLLogger>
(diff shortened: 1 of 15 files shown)


Objective-C/CBLFileLogger.h
```diff
@@ -19,41 +19,28 @@
...
-/**
- The directory to store the log files.
- */
-@property (nonatomic) NSString* directory;
-
-/**
- To use plain text file format instead of the default binary format.
+ The log file configuration for configuring the log directory, file format, and rotation policy.
+ The config property is nil by default. Setting the config property to nil will disable the file
+ logging.
```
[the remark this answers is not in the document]
pasin: OK, PR is updated.

Objective-C/CBLLogFileConfiguration.m
```diff
@@ -0,0 +1,105 @@
...
+
+- (void) checkReadonly {
+    if (_readonly) {
+        [NSException raise: NSInternalInconsistencyException
+                    format: @"This configuration object is readonly."];
+    }
+}
+
+
+- (NSString*) defaultDirectory {
```
jayahariv: not seeing anywhere this is used, as well as an option to set directory for `CBLLogFileConfiguration`, intentional?

pasin: I left it there before talking to @rajagp yesterday to confirm that we will not have default directory anymore. So the conclusion is that we will not have default directory anymore. We could remove that.

pasin: PR updated.

Swift/LogFileConfiguration.swift
```diff
@@ -0,0 +1,84 @@
...
+        self.readonly = readonly
+        self.directory = config.directory
+        self.usePlainText = config.usePlainText
+        self.maxSize = config.maxSize
+        self.maxRotateCount = config.maxRotateCount
+    }
+    
+    func checkReadOnly() {
+        if self.readonly {
+            fatalError("This configuration object is readonly.")
```
jayahariv: did you mean 

```
if !readOnly {
  // show error
}
```

pasin: No, only show if it's readonly. I think it's already correct.

Objective-C/CBLLogFileConfiguration.m
```diff
@@ -0,0 +1,89 @@
...
+
+
+#pragma mark - Internal
+
+
+- (void) checkReadonly {
+    if (_readonly) {
+        [NSException raise: NSInternalInconsistencyException
+                    format: @"This configuration object is readonly."];
+    }
```
jayahariv: understood what this mean by, but the method confused. may be adding a single line of comment saying, "will raise an exception when called on _readOnly". Changing the method name might crush the consistency else we should have renamed the method more relevant.

pasin: I think method name is good enough. Another option might be `validateReadOnly`. I will add a comment to the method. Thanks!

PR #2341, opened 2019-01-21
Fix Blob metadata not preserved in predictive query parameters

pasin:
* Encoded blob in parameters as blob properties + content instead of content only.

* Updated TextModel in unit tests to check the content type of blob.

#2320

### DIFF
--- a/Objective-C/CBLBlob.mm
+++ b/Objective-C/CBLBlob.mm
@@ -42,7 +42,9 @@ static const size_t kMaxCachedContentLength = 8*1024;
 static const size_t kReadBufferSize = 8*1024;
 
 static NSString* const kTypeMetaProperty = @kC4ObjectTypeProperty;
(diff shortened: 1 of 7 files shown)


Objective-C/CBLBlob.mm
```diff
@@ -42,9 +42,11 @@
 static const size_t kReadBufferSize = 8*1024;
 
 static NSString* const kTypeMetaProperty = @kC4ObjectTypeProperty;
+static NSString* const kDataMetaProperty = @kC4BlobDataProperty;
 static NSString* const kBlobType = @kC4ObjectType_Blob;
 
 
+
```
jayahariv: is this intentional extra line?

pasin: I will remove the extra line. Thanks.

Issue #2344, opened 2019-01-24
Add warning when file logging is not configured


pasin:
We should add some warning when file logging is not configured.

PR #2345, opened 2019-01-24
Add warning message when file logging is not configured

pasin:
* Added warning when file logging is not configured. The warning will display only once when a database instance is created. The warning message will not show more than once even there are multiple databases created.

* I have tested it manually for both Swift and Objective-C. I was not able to write a unit test as we always have a default testing database opened in the base test class’s setup.

#2344

### DIFF
--- a/Objective-C/Internal/CBLDatabase+Swift.h
+++ b/Objective-C/Internal/CBLDatabase+Swift.h
@@ -1,0 +1,15 @@
+//
+//  CBLDatabase+Swift.h
+//  CouchbaseLite
+//
+//  Created by Pasin Suriyentrakorn on 1/23/19.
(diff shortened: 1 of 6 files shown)


Objective-C/Internal/CBLDatabase+Swift.h
```diff
@@ -0,0 +1,15 @@
+//
+//  CBLDatabase+Swift.h
+//  CouchbaseLite
+//
+//  Created by Pasin Suriyentrakorn on 1/23/19.
+//  Copyright © 2019 Couchbase. All rights reserved.
+//
```
jayahariv: i think header needs to be updated...

pasin: Thanks. I will update it.

PR #2349, opened 2019-01-25
Change to use OptionSet type for ConsoleLogger.domains

pasin:
Instead of using Set\<LogDomain\>, changed to use OptionSet.

### DIFF
--- a/Swift/ConsoleLogger.swift
+++ b/Swift/ConsoleLogger.swift
@@ -20,6 +20,33 @@
 import Foundation
 
 /// Console logger for writing log messages to the system console.
+
+/// Log domain options that can be enabled in the console logger.
+public struct LogDomains: OptionSet {
+    /// Raw value.
+    public let rawValue: Int
+    
+    /// Constructor with the raw value.
+    public init(rawValue: Int) {
+        self.rawValue = rawValue
+    }
+    
+    /// All domains.
+    public static let all = LogDomains(rawValue: Int(LogDomain.all.rawValue))
+    
+    /// Database domain.
+    public static let database = LogDomains(rawValue: Int(LogDomain.database.rawValue))
+    
+    /// Database domain.
+    public static let query = LogDomains(rawValue: Int(LogDomain.query.rawValue))
+    
+    /// Replicator domain.
+    public static let replicator = LogDomains(rawValue: Int(LogDomain.replicator.rawValue))
+    
+    /// Network domain.
+    public static let network = LogDomains(rawValue: Int(LogDomain.network.rawValue))
+}
+
 public class ConsoleLogger {
     
     /// The minimum log level of the log messages to be logged. The default log level for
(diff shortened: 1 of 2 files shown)


Swift/ConsoleLogger.swift
```diff
@@ -20,6 +20,33 @@
...
+    /// Constructor with the raw value.
+    public init(rawValue: Int) {
+        self.rawValue = rawValue
+    }
+    
+    /// All domains.
+    public static let all = LogDomains(rawValue: Int(LogDomain.all.rawValue))
+    
+    /// Database domain.
+    public static let database = LogDomains(rawValue: Int(LogDomain.database.rawValue))
```
jayahariv: are we doing this? 
```suggestion
    public static let database = LogDomains(rawValue: 1 << 0)
```

Swift/ConsoleLogger.swift
```diff
@@ -20,6 +20,33 @@
...
+    /// Raw value.
+    public let rawValue: Int
+    
+    /// Constructor with the raw value.
+    public init(rawValue: Int) {
+        self.rawValue = rawValue
+    }
+    
+    /// All domains.
+    public static let all = LogDomains(rawValue: Int(LogDomain.all.rawValue))
```
jayahariv:
```suggestion
    public static let all =[.database, .query, .replicator, .network]
```

Swift/ConsoleLogger.swift
```diff
@@ -20,6 +20,33 @@
...
+    
+    /// Database domain.
+    public static let database = LogDomains(rawValue: Int(LogDomain.database.rawValue))
+    
+    /// Database domain.
+    public static let query = LogDomains(rawValue: Int(LogDomain.query.rawValue))
```
jayahariv:
```suggestion
    public static let query = LogDomains(rawValue: 1 << 1)
```

Swift/ConsoleLogger.swift
```diff
@@ -20,6 +20,33 @@
...
+    
+    /// Replicator domain.
+    public static let replicator = LogDomains(rawValue: Int(LogDomain.replicator.rawValue))
+    
+    /// Network domain.
+    public static let network = LogDomains(rawValue: Int(LogDomain.network.rawValue))
```
jayahariv:
```suggestion
    public static let network = LogDomains(rawValue: 1 << 3)
```

Swift/ConsoleLogger.swift
pasin: I prefer to keep it that way so that we have the raw value defined in one place. The value needs to be matched with Objc’s value as well.

pasin:
It’s the same.

Swift/ConsoleLogger.swift
pasin: LogDomain.database has done that.

Swift/ConsoleLogger.swift
pasin: I want to keep the value inline with LogDomain.all and objc value. So it’s better to keep it that way.

jayahariv:
You are right. Keeping in single place will be better.

PR #2350, opened 2019-01-26
ci: coveralls

jayahariv:
* add coveralls
* enable slather and ignore all swift and vendor files.
* simulator corrected to iPhone XS

$COVERALL_REPO_TOKEN  is added as travis environment variable. 

screen shot 2019-01-26 at 12 26 52 pm

### DIFF
--- a/.coveralls.yml
+++ b/.coveralls.yml
@@ -1,0 +1,1 @@
+repo_token: $COVERALL_REPO_TOKEN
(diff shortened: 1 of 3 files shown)


jayahariv:
YaY works fine. 

screen shot 2019-01-26 at 12 40 03 pm

pasin:
Wondering if it's possible to do both ObjC and Swift?

Issue #2353, opened 2019-01-29
Validate and add NonNull to all public APIs


jayahariv:
Validate and add NonNull to all public APIs. 

DATABASE
- [x] Database.swift
- [x] DatabaseChangeListener.swift
- [x] DatabaseConfiguration.swift
- [x] DocumentChangeListener.swift
- [x] EncryptionKey.swift
- [x] ListenerToken.swift
- [x] Logging.swift
- [x] Prediction.swift

DOCUMENT
- [x] ArrayObject.swift
- [x] Blob.swift
- [x] DictionaryObject.swift
- [x] Document.swift
- [x] DocumentFragment.swift
- [x] Fragment.swift
- [x] MutableArrayObject.swift
- [x] MutableDictionaryObject.swift
- [x] MutableDocument.swift
- [x] MutableFragment.swift

REPLICATOR
- [x] Authenticator.swift
- [x] DocumentFlags.swift
- [x] Endpoint.swift
- [x] Replicator.swift
- [x] ReplicatorConfiguration.swift
- [x] Message.swift	
- [x] MessageEndpoint.swift
- [x] MessageEndpointConnection.swift
- [x] MessageEndpointListener.swift

QUERY
- [x] Iterable.swift
- [x] Parameters.swift
- [x] Query.swift
- [x] QueryBuilder.swift
- [x] QueryChangeListener.swift
- [x] Result.swift
- [x] ResultSet.swift

CLAUSE
- [x] Collation.swift
- [x] DataSource.swift
- [x] From.swift
- [x] GroupBy.swift
- [x] Having.swift
- [x] Join.swift
- [x] Joins.swift
- [x] Limit.swift
- [x] OrderBy.swift
- [x] Ordering.swift
- [x] Select.swift
- [x] SelectResult.swift
- [x] Where.swift

EXPRESSION
- [x] ArrayExpression.swift
- [x] ArrayExpressionIn.swift
- [x] ArrayExpressionSatisfies.swift
- [x] Expression.swift
- [x] FullTextExpression.swift
- [x] Meta.swift
- [x] PropertyExpression.swift
- [x] VariableExpression.swift

FUNCTION
- [x] ArrayFunction.swift
- [x] FullTextFunction.swift
- [x] Function.swift

INDEX
- [x] Index.swift

ROUTER
- [x] FromRouter.swift
- [x] GroupByRouter.swift
- [x] HavingRouter.swift
- [x] JoinRouter.swift
[...]

PR #2354, opened 2019-01-30
feat: add non-null check for parameters

jayahariv:
* added the null check for parameters.

link: #2353

### DIFF
--- a/Objective-C/CBLDatabase.mm
+++ b/Objective-C/CBLDatabase.mm
@@ -184,6 +184,8 @@ static void dbObserverCallback(C4DatabaseObserver* obs, void* context) {
 
 
 - (CBLDocument*) documentWithID: (NSString*)documentID {
+    CBLAssertNotNil(documentID);
+    
     return [self documentWithID: documentID error: nil];
 }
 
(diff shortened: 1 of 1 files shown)


Objective-C/CBLDatabase.mm
```diff
@@ -192,6 +194,8 @@ - (CBLDocument*) documentWithID: (NSString*)documentID {
 
 
 - (CBLDocumentFragment*) objectForKeyedSubscript: (NSString*)documentID {
+    CBLAssertNotNil(documentID);
```
snej: This check isn't necessary, since the `documentID` is just passed to `[self documentWithID:]`, which checks.

Objective-C/CBLDatabase.mm
```diff
@@ -184,6 +184,8 @@ - (CBLDatabaseConfiguration*) config {
 
 
 - (CBLDocument*) documentWithID: (NSString*)documentID {
+    CBLAssertNotNil(documentID);
```
snej: This isn't necessary either, because the documentID just gets passed, unused, all the way to the CBLDocument initializer, which checks it (CBLDocument.mm:46.)

jayahariv:
f97465af37e07666230cfe0ebda381697f7f142c

PR #2358, opened 2019-01-31
fix: query parameter type

jayahariv:
* corrected the array, dictionary and blob type as optional.

link: #2353

### DIFF
--- a/Swift/Parameters.swift
+++ b/Swift/Parameters.swift
@@ -50,7 +50,7 @@ public class Parameters {
     ///   - value: The ArrayObject value.
     ///   - name: The parameter name.
     /// - Returns: The self object.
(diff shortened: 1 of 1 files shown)


jayahariv:
Is there a specific reason why this function is now not covered? 

hope, this should be called, when network status changes during some replication through some dispatch queue. 

```
static void ClientCallback(SCNetworkReachabilityRef target,
                           SCNetworkReachabilityFlags flags,
                           void *info)
{
    [(__bridge CBLReachability*)info flagsChanged: flags];
}
```

pasin:
I don't think unit tests have tested ClientCallback() as it requires network status changes. It doesn't seem to be related to this PR.

Issue #2282, opened 2018-12-14
Implement CoreML Predictive Model


djpongh:
create coreML impl

PR #2365, opened 2019-02-21
Implement Core ML Predictive Model

pasin:
* Automatically converted b/w CBL data and Core ML data when calling into the MLModel object to return prediction results. All Core ML data types including Int64, Double, String, Dictionary, MultiArray, Image, and Sequence are supported.

* When the MLObject has a single input and the input type is Image, CoreMLPredictiveModel will use Vision framework via the VNCoreMLModel to process the input image and call into the MLObject object to return the prediction result. If the condition doesn’t meet, the regular MLModel will be used.

* When using VNCoreMLModel, all VNObservation types including VNClassificationObservation, VNCoreMLFeatureValueObservation, and VNPixelBufferObservation are supported. However there is a limitation for supporting VNCoreMLFeatureValueObservation and VNPixelBufferObservation that the MLModel must return a single output as currently there is no way to map the result from VNCoreMLModel back to MLModel if the output are multiple.

* When converting blob data to VNPixelBuffer for an input image, only ARGB pixel format is currently supported. However this limitation is applied only when the VNCoreMLModel cannot be used.

#2282


pasin:
Close the PR for now as I want to make some changes.

PR #2366, opened 2019-02-21
Implement Core ML Predictive Model

pasin:
* Automatically converted b/w CBL data and Core ML data when calling into the MLModel object to return prediction results. All Core ML data types including Int64, Double, String, Dictionary, MultiArray, Image, and Sequence are supported.

* When the MLObject has a single input and the input type is Image, CoreMLPredictiveModel will use Vision framework via the VNCoreMLModel to process the input image and call into the MLObject object to return the prediction result. If the condition doesn’t meet, the regular MLModel will be used.

* When using VNCoreMLModel, all VNObservation types including VNClassificationObservation, VNCoreMLFeatureValueObservation, and VNPixelBufferObservation are supported. However there is a limitation for supporting VNCoreMLFeatureValueObservation and VNPixelBufferObservation that the MLModel must return a single output as currently there is no way to map the result from VNCoreMLModel back to MLModel if the output are multiple.

* When converting blob data to VNPixelBuffer for an input image, only ARGB pixel format is currently supported. However this limitation is applied only when the VNCoreMLModel cannot be used.

#2282


Objective-C/CBLCoreMLPredictiveModel.m
```diff
@@ -0,0 +1,538 @@
...
+#define kCBLMaxVisionPredictedProbabilities 10
+
+@implementation CBLCoreMLPredictiveModel {
+    MLModel* _model;
+    VNCoreMLModel* _vnModel;
+    NSString* _vnFeatureName;
+}
+
+- (instancetype) initWithMLModel: (MLModel*)model {
+    self = [super init];
```
jayahariv:
```suggestion
    CBLAssertNotNil(model);
    self = [super init];
```

pasin: Right. We should validate non-null as well. I will fix that.

Objective-C/CBLCoreMLPredictiveModel.m
```diff
@@ -0,0 +1,538 @@
...
+    id<MLFeatureProvider> output = [_model predictionFromFeatures: features error: &error];
+    if (!output)
+        return nil;
+    
+    return [self.class dictionaryFromFeatures: output];
+}
+
+
+- (nullable CBLDictionary*) predictUsingVNModel: (CBLDictionary*)input {
+    assert(_vnModel != nil && _vnFeatureName != nil);
```
jayahariv: for the consistency, should we use ??
`CBLAssertNotNil(_vnModel);` 
`CBLAssertNotNil(_vnFeatureName);`

pasin: No. It serves different purpose. Here is code assertion (not API non-null validation).

Objective-C/CBLCoreMLPredictiveModel.m
```diff
@@ -0,0 +1,538 @@
...
+
+
+- (CBLDictionary*) predictionResultFromObservations: (NSArray*)observations {
+    if (observations.count == 0)
+        return nil;
+    
+    // See https://developer.apple.com/documentation/vision/vncoremlrequest for
+    // the information about the VNObservation results:
+    NSString* predictedFeatureName = _model.modelDescription.predictedFeatureName;
+    if (predictedFeatureName) {
```
jayahariv: Since the @"" will not be nil, but this will cover both.

```suggestion
    if ([predictedFeatureName length] > 0) {
```

Objective-C/CBLCoreMLPredictiveModel.m
```diff
@@ -0,0 +1,538 @@
...
+        CBLMutableDictionary* result = [[CBLMutableDictionary alloc] init];
+        
+        // Predicted Feature:
+        VNClassificationObservation* top = observations[0];
+        [result setValue: top.identifier forKey: predictedFeatureName];
+        
+        // Predicted Probabilities:
+        NSString* predictedProbsName = _model.modelDescription.predictedProbabilitiesName;
+        if (predictedProbsName) {
```
jayahariv: same
```suggestion
        if ([predictedProbsName length] > 0) {
```

Objective-C/CBLCoreMLPredictiveModel.m
```diff
@@ -0,0 +1,538 @@
...
+    MLDictionaryFeatureProvider* provider =
+        [[MLDictionaryFeatureProvider alloc] initWithDictionary: dict error: &error];
+    if (!provider) {
+        CBLWarn(Query, @"Error when creating a dictionary feature provider: %@", error);
+    }
+    return provider;
+}
+
+
++ (MLFeatureValue*) featureValueFromValue: (id)value type: (MLFeatureType)type {
```
jayahariv: return type seems like nullable, wonder why it is not shown on Xcode compiler.

```suggestion
+ (nullable MLFeatureValue*) featureValueFromValue: (id)value type: (MLFeatureType)type {
```

Objective-C/CBLCoreMLPredictiveModel.m
```diff
@@ -0,0 +1,538 @@
...
+            NSString* str = $castIf(NSString, value);
+            return str ? [MLFeatureValue featureValueWithString: str] : nil;
+        }
+        case MLFeatureTypeDictionary: {
+            CBLDictionary* dict = $castIf(CBLDictionary, value);
+            if (!dict)
+                return nil;
+            NSError* error;
+            MLFeatureValue* dictValue = [MLFeatureValue featureValueWithDictionary:
+                                         [dict toDictionary] error: &error];
```
jayahariv: styling something like this right?
screen shot 2019-02-21 at 1 13 47 pm

Objective-C/CBLCoreMLPredictiveModel.m
```diff
@@ -0,0 +1,538 @@
...
+                return nil;
+            return [MLFeatureValue featureValueWithPixelBuffer: pixel];
+        }
+        default:
+            return nil;
+    }
+}
+
+
++ (MLFeatureValue*) sequenceFeatureValueFromValue: (id)value type: (MLFeatureType)type
```
jayahariv: nullable??

```suggestion
+ (nullable MLFeatureValue*) sequenceFeatureValueFromValue: (id)value type: (MLFeatureType)type
```

Objective-C/Tests/PredictiveQueryTest+CoreML.m
```diff
@@ -0,0 +1,361 @@
...
+#import "CBLTestCase.h"
+#import "CBLCoreMLPredictiveModel+Internal.h"
+
+API_AVAILABLE(macos(10.13), ios(11.0))
+@interface PredictiveQueryWithCoreMLTest : CBLTestCase
+
+@end
+
+@implementation PredictiveQueryWithCoreMLTest
+
```
jayahariv: extra line? 
```suggestion


```

Objective-C/Tests/PredictiveQueryTest+CoreML.m
```diff
@@ -0,0 +1,361 @@
...
+//
+
+#import "CBLTestCase.h"
+#import "CBLCoreMLPredictiveModel+Internal.h"
+
+API_AVAILABLE(macos(10.13), ios(11.0))
+@interface PredictiveQueryWithCoreMLTest : CBLTestCase
+
+@end
+
```
jayahariv: extra line?
```suggestion


```

Objective-C/CBLCoreMLPredictiveModel.m
pasin: We have it defined as nullable in .h.

Objective-C/CBLCoreMLPredictiveModel.m
pasin: Not sure if it's necessary based on https://developer.apple.com/documentation/vision/vncoremlrequest?language=objc documentation. If that happens, it seems to an invalid model.

Objective-C/CBLCoreMLPredictiveModel.m
pasin: Not sure if it's necessary based on https://developer.apple.com/documentation/vision/vncoremlrequest?language=objc documentation. If that happens, it seems to an invalid model.

Objective-C/CBLCoreMLPredictiveModel.m
pasin: OK, I will fix that.

Objective-C/CBLCoreMLPredictiveModel.m
pasin: We have it defined nullable in .h.

Objective-C/Tests/PredictiveQueryTest+CoreML.m
pasin: OK

Objective-C/Tests/PredictiveQueryTest+CoreML.m
pasin: OK

Swift/Tests/CBLTestCase.swift
```diff
@@ -199,6 +205,18 @@ class CBLTestCase: XCTestCase {
         }
     }
     
+    
+    func ignoreExpcetion(block: @escaping () throws -> Void) {
```
jayahariv:
```suggestion
    func ignoreException(block: @escaping () throws -> Void) {
```

Swift/Tests/PredictiveQueryTest+CoreML.swift
```diff
@@ -0,0 +1,221 @@
...
+    
+    func coreMLModel(name: String, mustExist: Bool = true) throws -> MLModel? {
+        let resource = "mlmodels/\(name)"
+        guard let modelURL = urlForResource(name: resource, ofType: "mlmodel") else {
+            XCTAssertFalse(mustExist)
+            return nil
+        }
+        
+        var mlmodel: MLModel? = nil
+        ignoreExpcetion {
```
jayahariv:
```suggestion
        ignoreException {
```

Swift/Tests/CBLTestCase.swift
pasin: Thanks. Will fix this.

Swift/Tests/PredictiveQueryTest+CoreML.swift
```diff
@@ -0,0 +1,221 @@
...
+            if values.count > 3 {
+                if let v = values[3] {
+                    doc.setValue(v, forKey: "expected_price")
+                }
+            }
+            try saveDocument(doc)
+        }
+    }
+    
+    func crateDocumentWithImage(at path: String) throws {
```
jayahariv: spelling
```suggestion
    func createDocumentWithImage(at path: String) throws {
```

pasin: Thanks.

Objective-C/CBLCoreMLPredictiveModel.m
pasin: So let check nil based on Apple documentation.

Objective-C/CBLCoreMLPredictiveModel.m
pasin: So let check nil based on Apple documentation.

Swift/Tests/PredictiveQueryTest+CoreML.swift
```diff
@@ -0,0 +1,221 @@
...
+            .from(DataSource.database(db))
+            .orderBy(Ordering.expression(distance))
+        let params = Parameters()
+        params.setArray(lennon1, forName: "vectorParam")
+        q.parameters = params
+        rows = try verifyQuery(q) { (n, r) in
+            let name = r.string(at: 0)!
+            names.append(name)
+            XCTAssertNotNil(r.value(at: 1))
+            XCTAssertTrue(r.double(at: 1) >= 0)
```
jayahariv: can the distance anytime become less than zero?

pasin: No, based on https://en.wikipedia.org/wiki/Euclidean_distance.

jayahariv: can we validate the `lennon-1` distance is 0? because they are equal right.

pasin: Yes, we can add that check.

Swift/Tests/CBLTestCase.swift
pasin: I think the implementation of the method can be simplified too. I will do that.

Objective-C/Tests/PredictiveQueryTest+CoreML.m
pasin: Actually we don't add extra line here.

Objective-C/CBLCoreMLPredictiveModel.m
```diff
@@ -0,0 +1,538 @@
...
+    if (self) {
+        _model = model;
+        [self setupVisionModel];
+    }
+    return self;
+}
+
+
+- (void) setupVisionModel {
+    if (!kCBLMayUseVisionModel)
```
jayahariv: this returns every time right? not seeing anywhere this is reset.

pasin: It's defined at the top of the file.

pasin: Let's remove the check as I don't think we will ever disable it.

Objective-C/CBLCoreMLPredictiveModel.m
```diff
@@ -0,0 +1,538 @@
...
+    if (inputSpecs.count == 1 && inputSpecs.allValues[0].type == MLFeatureTypeImage) {
+        NSError* error;
+        _vnModel = [VNCoreMLModel modelForMLModel: _model error: &error];
+        if (!_vnModel) {
+            CBLWarn(Query, @"Cannot create vision model from the model: %@. "
+                            "The regular MLModel will be used instead.", error);
+            return;
+        }
+        
+        BOOL isSupported = (desc.predictedFeatureName || outputSpecs.count == 1);
```
jayahariv: not sure whether this benefit anything, but since the feature name is a String. String can be `""`
```suggestion
        BOOL isSupported = (!desc.predictedFeatureName.isEmpty || outputSpecs.count == 1);
```

pasin: I think we should do according to Apple Document that if predictedFeatureName is not null (empty string is not null as well), the observation type is VNClassificationObservation.

Objective-C/CBLCoreMLPredictiveModel.m
```diff
@@ -0,0 +1,538 @@
...
+    }
+}
+
+
+#pragma mark - CBLDictionary to Features Conversion
+
+
++ (id<MLFeatureProvider>) featuresFromDictionary: (CBLDictionary*)dictionary
+                                            spec: (NSDictionary<NSString*, MLFeatureDescription*>*)spec {
+    NSMutableDictionary* dict = [[NSMutableDictionary alloc] initWithCapacity: spec.count];
```
jayahariv: since the parameter we receive is also a `dictionary`, and this variable also is `dict` looks similar. may be `featureProviderDictionary`?? what do you think?

pasin: How about `featureProviderDict`?

Objective-C/Tests/PredictiveQueryTest+CoreML.m
jayahariv: before implementation is it only one line? but i could see CBLDocument, CBLFragment etc are all using two lines before the implementation starts.

Objective-C/CBLCoreMLPredictiveModel.m
jayahariv: yea!

Objective-C/Tests/PredictiveQueryTest+CoreML.m
pasin: I don't think it's consistent. I will could make two lines here.

pasin:
I have push up the fix. Please check. Thanks.

PR #2386, opened 2019-03-26
Convert to Swift 5.0

farktronix:
I ran the Xcode migration tool to convert this project to Swift 5.0. Only one real line of code change was needed.


pasin:
@farktronix we are using xcconfig to configure the project. Please configure the xcode version in this [file](https://github.com/couchbase/couchbase-lite-ios/blob/master/xcconfigs/Project.xcconfig#L53) instead of at the xcode project itself.

farktronix:
@pasin Thanks- I didn't catch that this was all xcconfig based. I made the change and squashed it into the previous commit.

pmusolino:
Any news on this PR?

Issue #2376, opened 2019-03-18
Test coverage improve 


jayahariv:
improve test coverage. Test coverage included the unit test files

jayahariv:
seems like we covered most of the main scenarios. Closing this.

jayahariv:
EE => ObjC = 80%; Swift = 77%
CE => ObjC = 78%; Swift = 69%

Seems like the CE-Swift is using some classes which are not covered in the test cases. Will re-open and make sure the CE-Swift is covering all valid scenarios.

PR #2387, opened 2019-03-27
unit: dict and doc type setter methods

jayahariv:
* add tests for CBLMutableDictionary
* add tests for type setter methods for CBLMutableDocument. 
* corrected the pragma. 

Ref: #2376

### DIFF
--- a/Objective-C/CBLMutableDocument.mm
+++ b/Objective-C/CBLMutableDocument.mm
@@ -99,7 +99,7 @@
 }
 
 
-#pragma mark - CBLMutableDictionary
+#pragma mark - CBLMutableDocument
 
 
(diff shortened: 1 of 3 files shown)


Objective-C/CBLMutableDocument.mm
```diff
@@ -99,7 +99,7 @@ - (CBLMutableDocument*) mutableCopyWithZone: (NSZone *)zone {
 }
 
 
-#pragma mark - CBLMutableDictionary
+#pragma mark - CBLMutableDocument
```
jayahariv:
```suggestion
#pragma mark - CBLMutableDictionary
```

PR #2392, opened 2019-03-28
unit: query params set method

jayahariv:
* fix issue with the Date in QueryParameter. 
* wrote a loadStudents method for validating different types.
* remaining are dictionary, blob and array which will be added as separate function

Ref: #2376

### DIFF
--- a/Objective-C/Tests/QueryTest.m
+++ b/Objective-C/Tests/QueryTest.m
@@ -60,6 +60,108 @@
     return numbers;
 }
 
+- (void) loadStudents {
+    NSError* error;
+    NSTimeInterval aDayInterval = 24 * 60 * 60;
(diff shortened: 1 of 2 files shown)


Objective-C/Tests/QueryTest.m
```diff
@@ -922,6 +922,30 @@ - (void) testParameters {
     AssertEqual(numRows, 4u);
 }
 
+- (void) testStringQueryParameter {
```
pasin: It's it possible to rewrite and merge testParameters and testStringQueryParameter?

jayahariv: Re-wrote the function and mixed to a single function for checking string, number, integer, long long, float, double, boolean, date. Remaining are the Blob, Dictionary and Array which we will add a separate function.

PR #2398, opened 2019-04-04
unit: check parse date

jayahariv:
* check parse date with a known date string and double return value.

Ref: #2376


Objective-C/Tests/MiscTest.m
```diff
@@ -37,5 +38,12 @@ - (void) testJSONDateRoundTrip {
     AssertEqualObjects(dateStr2, dateStr1);
 }
 
+#pragma mark - Parse Date
+
+- (void) testCBLParseISO8601Date {
+    AssertEqual(CBLParseISO8601Date("1970-01-01T00:00:01.000Z"), 1.0);
+    AssertEqual(CBLParseISO8601Date("2018-10-23T11:33:01-0700"), 1540319581);
+    AssertEqual(CBLParseISO8601Date("2020-02-29T23:59:59.000000+0000"), 1583020799);
```
jayahariv: This test fails, Need to investigate. 
`AssertEqual(CBLParseISO8601Date("1970-01-01T00:00:00.123Z"), 0.123);`

jayahariv: floating point numbers needs to be rounded off!! corrected and pushed the fix

PR #2408, opened 2019-04-10
unit: check the method with init tests

jayahariv:
* check for dataWithJSONObject and stringForJSONObject

Ref: #2376

### DIFF
--- a/Objective-C/Tests/MiscTest.m
+++ b/Objective-C/Tests/MiscTest.m
@@ -113,8 +113,16 @@
 
 - (void) testStringWithJSONObject {
     NSError* error;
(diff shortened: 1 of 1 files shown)


Objective-C/Tests/MiscTest.m
```diff
@@ -95,4 +95,41 @@ - (void) testCBLParseISO8601Date {
     Assert(CBLParseISO8601Date("1970-01-01T00:00:00.123Z") - 0.123 < 0.0001);
 }
 
+#pragma mark - CBLJSON
+
+- (void) testDataWithJSONObject {
+    NSError* error;
+    NSArray* inputs = @[@"couchbase-mobile", @2021, @[@"abcd"]];
```
pasin: We could add more data types here but this is good enough for now.

PR #2410, opened 2019-04-11
unit: add misc cpp tests

jayahariv:
* there are a few tests in CPP, which needs to be done from cpp test class itself. so kept a common test file with cpp tests in it.
* convertError method was not tested with some valid and invalid test scenarios.

Ref: #2376

### DIFF
--- a/Objective-C/Tests/MiscCppTest.mm
+++ b/Objective-C/Tests/MiscCppTest.mm
@@ -1,0 +1,88 @@
+//
+//  MiscCppTest.mm
+//  CBL ObjC Tests
+//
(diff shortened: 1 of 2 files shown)


Objective-C/Tests/MiscCppTest.mm
```diff
@@ -0,0 +1,88 @@
+//
+//  MiscCppTest.mm
+//  CBL ObjC Tests
+//
+//  Created by Jayahari Vavachan on 4/10/19.
+//  Copyright © 2019 Couchbase. All rights reserved.
+//
```
pasin: Please fix the header.

PR #2412, opened 2019-04-12
unit: add unary missing checks

jayahariv:
* validated the result when checking for not null or not missing with  function.

Ref: #2376

### DIFF
--- a/Objective-C/Tests/QueryTest.m
+++ b/Objective-C/Tests/QueryTest.m
@@ -28,6 +28,7 @@
 #import "CBLQueryResultArray.h"
 #import "CBLValueExpression.h"
 #import "CBLQueryExpression+Internal.h"
+#import "CBLUnaryExpression.h"
 
 #define kDOCID      [CBLQuerySelectResult expression: [CBLQueryMeta id]]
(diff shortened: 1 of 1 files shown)


Objective-C/Tests/QueryTest.m
```diff
@@ -2601,4 +2602,43 @@ - (void) testValueExpression {
...
+    NSDate* nw = [NSDate date];
+    CBLMutableDocument* doc1 = [self createDocument];
+    [doc1 setDate: nw forKey: @"now"];
+    [self saveDocument: doc1];
+    
+    CBLMutableDocument* doc2 = [self createDocument];
+    [self saveDocument: doc2];
+    
+    CBLUnaryExpression* notNull;
+    CBLUnaryExpression* notMiss;
```
pasin: CBLUnaryExpression is an internal API. Is using notNullOrMissing enough to test it?

jayahariv: Even though it is internal, we are not using the `not null` or `not missing` checks. For `isNotNullOrMissing`, we are using `negate: nullOrMissing`, which is also right but code is not getting validated for the result of NotNull and NotMissing. 

Impact of this test is not much in terms of exposed APIs, but more on an independent class basis. We are validating the CBLUnaryExpression for all possible usages.

PR #2414, opened 2019-04-12
refact: more readable

jayahariv:
* more readable and clear statements.
* include inner join and join. check for number2 as well as number1.
* add comments to the select statements, to distinguish between select and selectDistinct.

Ref: #2376

### DIFF
--- a/Objective-C/Tests/QueryTest.m
+++ b/Objective-C/Tests/QueryTest.m
@@ -567,36 +567,36 @@
     [joinmeCopy setValue: @42 forKey: @"theone"];
     [self saveDocument: joinmeCopy];
     
-    CBLQuerySelectResult* MAIN_DOC_ID =
-        [CBLQuerySelectResult expression: [CBLQueryMeta idFrom: @"main"]];
+    CBLQueryExpression* propNum1 = [CBLQueryExpression property: @"number1" from: @"main"];
+    CBLQueryExpression* propNum2 = [CBLQueryExpression property: @"number2" from: @"main"];
+    CBLQueryExpression* propTheOne = [CBLQueryExpression property: @"theone" from: @"secondary"];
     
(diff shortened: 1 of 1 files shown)


Objective-C/Tests/QueryTest.m
```diff
@@ -567,36 +567,36 @@ - (void) testJoin {
...
+    CBLQueryExpression* propTheOne = [CBLQueryExpression property: @"theone" from: @"secondary"];
     
-    CBLQueryExpression* on = [[CBLQueryExpression property: @"number1" from: @"main"]
-                              equalTo: [CBLQueryExpression property:@"theone" from:@"secondary"]];
     CBLQueryJoin* join = [CBLQueryJoin join: [CBLQueryDataSource database: self.db as: @"secondary"]
-                                         on: on];
-    CBLQuery* q = [CBLQueryBuilder select: @[MAIN_DOC_ID]
+                                         on: [propNum1 equalTo: propTheOne]];
+    CBLQueryJoin* innerJoin;
+    innerJoin = [CBLQueryJoin join: [CBLQueryDataSource database: self.db as: @"secondary"]
```
pasin: What is the different b/w innerJoin and join above?

jayahariv: Oops! i thought about making this a innerJoin, just for using two different kind of statements(but same end results). But ended up being writing the same.

jayahariv: what do you think, using inner-join and join with select and select-distinct, is that okay? or should we make it join in both cases? 

I used because of no particular reason. inner-join and join will give us same result, for the sake of using different statement i have used in this context.

pasin: Yes, you can use a different one. Please add maybe one line comment on that as well.

jayahariv: added the comment.

PR #2417, opened 2019-04-12
doc: update readme 2.5

jayahariv:
* update version for all samples, documentation etc to 2.5

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,7 +10,7 @@ Couchbase Lite 2.5 has a completely new set of APIs. The implementation is on to
 
 ## Requirements
 - iOS 9.0+ | macOS 10.11+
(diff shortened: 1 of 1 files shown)


pasin:
Let merge this when we release 2.5. We should also bump the minimum Xcode version as well. I'm not sure what is the minimum now. I guessed XCode 10.1 but we need to test.

jayahariv:
Yea, I was waiting for that to ask. XCode 10.0 is fine. since the test include `Int.random(range)`[ref](https://developer.apple.com/documentation/swift/int/2995648-random). 

[log test setup](https://github.com/couchbase/couchbase-lite-ios/blob/master/Swift/Tests/LogTest.swift#L36)

PR #2422, opened 2019-04-19
fix: query column title correction

jayahariv:
* removed the workaround and added the fix for the query column title issue.
* fixed the issue with selectResult with alias in Swift.

### DIFF
--- a/Swift/SelectResult.swift
+++ b/Swift/SelectResult.swift
@@ -111,7 +111,7 @@ public final class SelectResult {
 /* Internal */ class QuerySelectResultFrom: QuerySelectResult, SelectResultFrom {
     
     public func from(_ alias: String?) -> SelectResultProtocol {
-        return QuerySelectResult(expression: Expression.all().from(alias), alias: nil)
+        return QuerySelectResult(expression: Expression.all().from(alias), alias: alias)
     }
     
 }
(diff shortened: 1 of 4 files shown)


Swift/SelectResult.swift
```diff
@@ -111,7 +111,7 @@ public final class SelectResult {
 /* Internal */ class QuerySelectResultFrom: QuerySelectResult, SelectResultFrom {
     
     public func from(_ alias: String?) -> SelectResultProtocol {
-        return QuerySelectResult(expression: Expression.all().from(alias), alias: nil)
+        return QuerySelectResult(expression: Expression.all().from(alias), alias: alias)
```
jayahariv: if we revert this change, the test will in Swift. Since the [SelectResult asJSON] is returning without alias. 

Query JSON without alias:
```
{"WHAT":[[".main."],[".secondary."]],"DISTINCT":true,"FROM":[{"AS":"main"},{"ON":["=",[".main.number1"],[".secondary.theOne"]],"AS":"secondary","JOIN":"INNER"}]}
```

Swift/Tests/QueryTest.swift
```diff
@@ -1047,6 +1047,32 @@ class QueryTest: CBLTestCase {
     }
     
     
+    func testSelectAllWithDatabaseAliasWithMultipleSources() throws {
+        Database.log.console.domains = .all
```
jayahariv: need to remove

PR #2423, opened 2019-04-19
refact: query test class

jayahariv:
* separated the class into three, (i) QueryWithMeta will include all tests on meta, (ii) QueryWithJoins include all tests for Join statements and (iii) QueryTests include all the remaining.
* couldn't find anything else to keep them separate, as the tests are getting increased and the class is growing quickly. made us to reduce the size from ~2500 to ~1800.


jayahariv:
@pasin can you check this once again? 

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

Issue #2427, opened 2019-04-25
Replicator Conflict Resolver Implementation


jayahariv:
- [x] Define ConflictResolver interface
- [x] Add ReplicatorConfiguration’s conflictResolver property
- [x] Extract default conflict resolution into ConflictResolution.default
- [x] Implement custom conflict resolution logic in the replicator
  - Ensure that the custom conflict resolver is called outside transaction
  - The conflict resolver might be called multiple time if the conflict occurs again when saving the resolved document into the database
  - Ensure that the custom conflict resolver is called from a dedicated non-serialize queue / work-executor
- [x] Handle when there is an exception thrown from the custom resolver

PR #2431, opened 2019-04-26
feat: custom conflict resolution implementation in ObjC

jayahariv:
* adds interface for ConflictResolver, ConflictResolution
* adds the default conflict resolution static function.
* adds property to the replication configuration instance. not handled the property for now.
* add Conflict class with internal constructor and readOnly properties exposed. 
* implemented the conflict resolution case.
* exception handling add

Ref:  #2427


Objective-C/CBLConflictResolution.m
```diff
@@ -0,0 +1,40 @@
+//
+//  CBLConflictResolution.m
+//  CBL ObjC
+//
+//  Created by Jayahari Vavachan on 4/26/19.
+//  Copyright © 2019 Couchbase. All rights reserved.
```
jayahariv: will change this!

jayahariv: done!!

jayahariv:
taking this back, since the Conflict class seems to be declared separately. so that we can define the unavailable init, and make the implementation internal.

Objective-C/CBLConflict.m
```diff
@@ -0,0 +1,39 @@
...
+
+@implementation CBLConflict
+
+@synthesize localDocument=_localDocument, remoteDocument=_remoteDocument;
+
+# pragma mark - Internal
+
+- (instancetype) initWithLocalDocument:(CBLDocument *)localDoc
+                        remoteDocument:(CBLDocument *)remoteDoc {
+    NSParameterAssert(localDoc != nil && remoteDoc != nil);
```
jayahariv: it is nullable!!

jayahariv:
```suggestion
```

Objective-C/CBLConflict.h
```diff
@@ -0,0 +1,41 @@
+//
+//  CBLConflict.h
+//  CouchbaseLite
+//
+//  Copyright (c) 2017 Couchbase, Inc All rights reserved.
```
pasin: should be 2019.

Objective-C/CBLConflictResolution.h
```diff
@@ -0,0 +1,35 @@
+//
+//  CBLConflictResolution.h
+//  CouchbaseLite
+//
+//  Copyright (c) 2018 Couchbase, Inc All rights reserved.
```
pasin: Should be 2019

Objective-C/CBLConflictResolver.h
```diff
@@ -0,0 +1,35 @@
+//
+//  CBLConflictResolver.h
+//  CouchbaseLite
+//
+//  Copyright (c) 2017 Couchbase, Inc All rights reserved.
```
pasin: 2019

Objective-C/CBLDatabase.mm
```diff
@@ -950,102 +952,105 @@ - (BOOL) saveDocument: (CBLDocument*)document
...
         
         // Resolve conflict:
         CBLLogInfo(Sync, @"Resolving doc '%@' (mine=%@ and theirs=%@)",
                    docID, localDoc.revID, remoteDoc.revID);
+        CBLDocument* resolvedDoc = [conflictResolver resolve: conflict];
         
-        CBLDocument* resolvedDoc = [self resolveConflictBetweenLocalDoc: localDoc
-                                                           andRemoteDoc: remoteDoc];
+        if (resolvedDoc == nil) {
+            resolvedDoc = remoteDoc.isDeleted ? remoteDoc : localDoc;
```
pasin: It's possible that both remoteDoc and localDoc are not deleted doc. There might not be a real use case but the API allows it. I think you could just pass null to saveResolvedDocument() then saveResolvedDocument() will determine it as a deleted doc.

Objective-C/CBLDatabase.mm
```diff
@@ -950,102 +952,105 @@ - (BOOL) saveDocument: (CBLDocument*)document
...
-        // Read the conflicting remote revision:
-        CBLDocument* remoteDoc = [[CBLDocument alloc] initWithDatabase: self
-                                                            documentID: docID
-                                                        includeDeleted: YES
-                                                                 error: outError];
-        if (!remoteDoc || ![remoteDoc selectConflictingRevision])
-            return false;
+        conflictResolver = conflictResolver ?: [CBLConflictResolution default];
+        CBLConflict* conflict = [[CBLConflict alloc] initWithLocalDocument: localDoc
+                                                            remoteDocument: remoteDoc];
```
pasin: Need to check whether localDoc or remoteDoc is deleted doc or not. If it's a deleted doc, need to pass null. You may also add an assertion to make sure that both of them are not null.

Objective-C/CBLReplicator.mm
```diff
@@ -73,6 +73,7 @@ @interface CBLReplicator ()
 @implementation CBLReplicator
 {
     dispatch_queue_t _dispatchQueue;
+    dispatch_queue_t _otherDispatchQueue; // TODO: Naming???
```
pasin: How about _conflictResolverDispatchQueue?

Objective-C/Internal/Replicator/CBLConflict+Internal.h
```diff
@@ -0,0 +1,20 @@
+//
+//  CBLConflict+Internal.h
+//  CBL ObjC
+//
+//  Created by Jayahari Vavachan on 4/26/19.
```
pasin: Fix header.

Objective-C/CBLReplicator.mm
```diff
@@ -551,13 +556,18 @@ - (void) onDocsEnded: (NSArray<CBLReplicatedDocument*>*)docs pushing: (BOOL)push
...
-            NSError* error;
-            if ([_config.database resolveConflictInDocument: doc.id error: &error]) {
-                [doc resetError]; // Reset the error as successfully resolving the conflicts
-            } else
-                CBLWarn(Sync, @"%@: Conflict resolution of '%@' failed: %@", self, doc.id, error);
+            dispatch_async(_otherDispatchQueue, ^{
+                // Conflict pulling a document -- the revision was added but app needs to resolve it
+                CBLLogInfo(Sync, @"%@: pulled conflicting version of '%@'", self, doc.id);
+                NSError* error;
+                if ([_config.database resolveConflictInDocument: doc.id
```
pasin: Make sure to catch exception here. If you plan to do it in a separate commit, please add TODO: here.

Objective-C/CBLConflict.h
jayahariv:
```suggestion
//  Copyright (c) 2019 Couchbase, Inc All rights reserved.
```

Objective-C/CBLConflict.m
```diff
@@ -0,0 +1,38 @@
+//
+//  CBLConflict.m
+//  CouchbaseLite
+//
+//  Copyright (c) 2017 Couchbase, Inc All rights reserved.
```
jayahariv:
```suggestion
//  Copyright (c) 2019 Couchbase, Inc All rights reserved.
```

Objective-C/CBLConflictResolution.h
jayahariv:
```suggestion
//  Copyright (c) 2019 Couchbase, Inc All rights reserved.
```

Objective-C/CBLConflictResolution.m
```diff
@@ -0,0 +1,52 @@
+//
+//  CBLConflictResolution.m
+//  CouchbaseLite
+//
+//  Copyright (c) 2018 Couchbase, Inc All rights reserved.
```
jayahariv:
```suggestion
//  Copyright (c) 2019 Couchbase, Inc All rights reserved.
```

Objective-C/CBLConflictResolver.h
jayahariv:
```suggestion
//  Copyright (c) 2019 Couchbase, Inc All rights reserved.
```

Objective-C/CBLReplicator.mm
jayahariv: this can be used for any thing other than normal database serial queue right. so if we name only for conflict resolution, that might only be used for one purpose. Your thoughts?

NB: also how about `_concurrentQueue?`

Objective-C/CBLDatabase.mm
jayahariv: done

Objective-C/CBLDatabase.mm
jayahariv: done

Objective-C/CBLReplicator.mm
jayahariv: added TODO:

jayahariv:
assigning back to you, since i made the changes, and review whenever you get time.

Objective-C/CBLDatabase.mm
```diff
@@ -980,16 +980,39 @@ - (bool) resolveConflictInDocument: (NSString*)docID
         }
         
         conflictResolver = conflictResolver ?: [CBLConflictResolution default];
+        
+        CBLAssertNotNil(localDoc);
+        CBLAssertNotNil(remoteDoc);
```
pasin: It's possible localDoc or remoteDoc could be null.

Objective-C/CBLReplicator.mm
jayahariv: i named it `_concurrentQueue ` for now, do you have objections?

jayahariv: updated name to `_conflictResolverQueue `, till we have other purpose for concurrent operations.

jayahariv:
travis passed, but coveralls failed. unit tests will be added separately.

Objective-C/CBLReplicator.mm
```diff
@@ -565,9 +565,12 @@ - (void) onDocsEnded: (NSArray<CBLReplicatedDocument*>*)docs pushing: (BOOL)push
                                                           error: &error]) {
                     [doc resetError]; // Reset the error as successfully resolving the conflicts
```
pasin: Reset the error here has no point as the ReplicatedDocument has already been notified with the conflict error.

Objective-C/CBLReplicator.mm
```diff
@@ -565,9 +565,12 @@ - (void) onDocsEnded: (NSArray<CBLReplicatedDocument*>*)docs pushing: (BOOL)push
...
                 } else {
-                    // TODO: Handle the errors when the conflict happens, with the pull
                     CBLWarn(Sync, @"%@: Conflict resolution of '%@' failed: %@",
                             self, doc.id, error);
+                    id replication = [[CBLDocumentReplication alloc] initWithReplicator: self
+                                                                                 isPush: pushing
+                                                                              documents: @[doc]];
+                    [_docReplicationNotifier postChange: replication];
```
pasin: I'm not quite sure about the logic here as the ReplicatedDocument has already been notified with the conflict error. Now it will get notified again with another error.

I think we need to revise the whole logic around here. It could be that we will not notify ReplicatedDocument until the conflict gets resolved or has an error. Can you implement this logic and create a new issue to review the logic?

Objective-C/CBLDatabase.mm
```diff
@@ -950,102 +952,127 @@ - (BOOL) saveDocument: (CBLDocument*)document
...
-        CBLDocument* remoteDoc = [[CBLDocument alloc] initWithDatabase: self
-                                                            documentID: docID
-                                                        includeDeleted: YES
-                                                                 error: outError];
-        if (!remoteDoc || ![remoteDoc selectConflictingRevision])
-            return false;
+        Assert(localDoc == nil && remoteDoc == nil, @"Local and remote document shouldn't be empty \
+               at same time, when resolving conflict.");
+        localDoc = localDoc.isDeleted ? nil : localDoc;
+        remoteDoc = remoteDoc.isDeleted ? nil : remoteDoc;
```
pasin: I think localDoc or remoteDoc here should be set to nil because the log messages in line 994 and the logic in line 1018 will not be correct. You can do this when creating CBLConflict directly.

Objective-C/CBLDatabase.mm
```diff
@@ -950,102 +952,127 @@ - (BOOL) saveDocument: (CBLDocument*)document
...
+        CBLConflict* conflict = [[CBLConflict alloc] initWithLocalDocument: localDoc
+                                                            remoteDocument: remoteDoc];
         
         // Resolve conflict:
-        CBLLogInfo(Sync, @"Resolving doc '%@' (mine=%@ and theirs=%@)",
-                   docID, localDoc.revID, remoteDoc.revID);
+        CBLDocument* resolvedDoc;
+        @try {
+            CBLLogInfo(Sync, @"Resolving doc '%@' (mine=%@ and theirs=%@)",
```
pasin: change mine to local and theirs to remote to make it consistency with the naming in CBLConflict.

Objective-C/CBLDatabase.mm
```diff
@@ -950,102 +952,127 @@ - (BOOL) saveDocument: (CBLDocument*)document
...
+                       docID, localDoc.revID, remoteDoc.revID);
+            
+            resolvedDoc = [conflictResolver resolve: conflict];
+            
+            if (resolvedDoc.id != docID) {
```
pasin: resolvedDoc could be nil when returning a deleted doc.

Objective-C/CBLDatabase.mm
```diff
@@ -950,102 +952,127 @@ - (BOOL) saveDocument: (CBLDocument*)document
...
+                [NSException raise: NSInternalInconsistencyException
+                            format: @"Resolved docID '%@' is not matching with docID '%@'",
+                 resolvedDoc.id, docID];
+            }
+            
+            if (resolvedDoc.database != self) {
```
pasin: resolvedDoc could be nil so this check is not valid if resolved doc is nil.

Objective-C/CBLDatabase.mm
```diff
@@ -950,102 +952,127 @@ - (BOOL) saveDocument: (CBLDocument*)document
...
-                               mergedBody,
-                               mergedFlags,
-                               &c4err)
-        || !c4doc_save(rawDoc, 0, &c4err)) {
-        return convertError(c4err, outError);
+    CBL_LOCK(self) {
+        C4Transaction t(_c4db);
+        if (!t.begin())
+            return convertError(t.error(), outError);
+
```
pasin: If resolvedDoc is nil, we could try to match it to localDoc or remoteDoc only if one of them is a deletedDoc here.

Objective-C/CBLDatabase.mm
```diff
@@ -950,102 +952,127 @@ - (BOOL) saveDocument: (CBLDocument*)document
...
+        CBLStringBytes winningRevID = remoteDoc.revID;
+        CBLStringBytes losingRevID = localDoc.revID;
+        
+        alloc_slice mergedBody;
+        C4RevisionFlags mergedFlags = 0;
+        if (resolvedDoc != remoteDoc) {
+            // Unless the remote revision is being used as-is, we need a new revision:
+            mergedBody = [resolvedDoc encode: outError];
+            if (!mergedBody)
+                return false;
```
pasin: if resolvedDoc == nil, we don't need to call encode: and we shouldn't skip the check here too.

Objective-C/CBLReplicator.mm
```diff
@@ -108,6 +110,9 @@ - (instancetype) initWithConfig: (CBLReplicatorConfiguration *)config {
         _progressLevel = kCBLProgressLevelBasic;
         _changeNotifier = [CBLChangeNotifier new];
         _docReplicationNotifier = [CBLChangeNotifier new];
+        
+        NSString* qName = $sprintf(@"DB-Replicator-Other <%@>", config.database.name);
```
pasin: What does the 'Other' mean?

Objective-C/CBLReplicator.mm
jayahariv: I can add a separate PR for whole error handling. keeping this same logic with a TODO comment saying to revert this.

Objective-C/CBLReplicator.mm
jayahariv: other process, outside transaction. I can rename this to, `Conflict resolver`

Objective-C/CBLDatabase.mm
jayahariv: Yea!! Seems like, I need to revisit whole PR again with nil handling.

Objective-C/CBLReplicator.mm
jayahariv: i am not changing this for now, will do this handling in separate PR.

Objective-C/CBLReplicator.mm
jayahariv: added `Conflict resolver <dbName>`

Objective-C/CBLDatabase.mm
```diff
@@ -950,102 +952,138 @@ - (BOOL) saveDocument: (CBLDocument*)document
...
+        if (!t.begin())
+            return convertError(t.error(), outError);
+        
+        if (!resolvedDoc) {
+            if (!localDoc && remoteDoc)
+                resolvedDoc = remoteDoc; // local deleted, keep remote
+            
+            if (!remoteDoc && localDoc)
+                resolvedDoc = localDoc; // remote deleted, keep local
+        }
```
pasin: localDoc and remoteDoc will never be nil here. I think the logic should be something like below (You may need to clean it up):

```
if (!resolvedDoc) {
   if (localDoc.isDeleted)
       resolved = localDoc
   else if (remoteDoc.isDeleted)
      resolved = remoteDoc
}
```

Objective-C/CBLDatabase.mm
pasin: I think the logic here might be something like this (You may need to clean it up):

```
if (!resolvedDoc) {
    mergedFlags |= kRevDeleted;
} else if (resolvedDoc != remoteDoc) {
    mergedBody = [resolvedDoc encode: outError];
    if (!mergedBody)
        return false;
    if (resolvedDoc.isDeleted)
        mergedFlags |= kRevDeleted;
}
```

pasin: or 

```
if (resolvedDoc != remoteDoc) {
    if (resolvedDoc) {
        mergedBody = [resolvedDoc encode: outError];
        if (!mergedBody)
            return false;
    }

    BOOL isDeleted = !resolvedDoc || resolvedDoc.isDeleted
    if (isDeleted)
        mergedFlags |= kRevDeleted;
}
```

Objective-C/CBLDatabase.mm
jayahariv: we are assigning nil, if the localDoc or remoteDoc is deleted in the above function right? line 986, 987.

Objective-C/CBLDatabase.mm
pasin: I have a typo in the original comment. Can you review check it again?

Objective-C/CBLDatabase.mm
jayahariv: is there a chance resolvedDoc is nil, even though the local and remote is not deleted? if any of them is deleted, then the resolvedDoc will not be nil. they will be populated by the logic above 

if that is the case, we will mark the remote as deleted by replacing the local contents right?

pasin: > is there a chance resolvedDoc is nil, even though the local and remote is not deleted?

Yes. The API doesn't prevent at this moment.

> if any of them is deleted, then the resolvedDoc will not be nil. they will be populated by the logic above

This is correct.

> if that is the case, we will mark the remote as deleted by replacing the local contents right? which is wrong!

I couldn't follow this. Can you explain more?

Objective-C/CBLDatabase.mm
pasin: Yes

Objective-C/CBLDatabase.mm
jayahariv:
```
        alloc_slice mergedBody;
        C4RevisionFlags mergedFlags = 0;
        // resolvedDoc= nil
        // remote != nil
        // local != nil
        if (resolvedDoc != remoteDoc) {
            BOOL isDeleted = YES;
            if (resolvedDoc) {
                // Unless the remote revision is being used as-is, we need a new revision:
                mergedBody = [resolvedDoc encode: outError];
                if (!mergedBody)
                    return false;
                isDeleted = resolvedDoc.isDeleted;
            }
            
            if (isDeleted) // if resolved doc = nil, meaning the mergedFlags includes kRevDeleted
                mergedFlags |= kRevDeleted;
        }
        // mergedBody  = empty, since resolvedDoc = nil
        // winningRevId is remote.

//  remote will win, and body will be empty. 
```

pasin: This is correct. Remote should always win. I think empty body is fine because the content of the deleted doc will be empty anyway. If LiteCore doesn't allow nil, we will need to create an empty body.

Objective-C/Internal/Replicator/CBLConflict+Internal.h
```diff
@@ -0,0 +1,31 @@
...
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+//
+
+#import "CBLConflict.h"
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface CBLConflict (Internal)
```
pasin: If declaring (Internal) category, don't you need to add implementation with (Internal) marked?

jayahariv: implementation functions are in the CBLConflict.m similar to the CBLDatabase, CBLReplicator

jayahariv: i think for the sake of consistency, i should rename that to `CBLConflict ()`

jayahariv:
```suggestion
@interface CBLConflict ()
```

pasin: Normally when using category, I remember that we need to do implementation with category name as well. Maybe it doesn't really need to do that.

Objective-C/CBLReplicator.mm
jayahariv: https://github.com/couchbase/couchbase-lite-ios/issues/2444

Issue #2429, opened 2019-04-25
Database Save Conflict Handler Implementation


jayahariv:
- [x] Define ConflictHandler Interface or Block
- [x] Implement a new save method that takes the conflict handler
  - Ensure that the custom conflict handler is called outside transaction / lock
  - The conflict handler might be called multiple time if the conflict occurs again when saving the resolved document into the database
- [x] Ensure to capture an exception and re-throw to the save() method itself.

PR #2432, opened 2019-04-27
feat: database save conflict handler implementation in ObjC

jayahariv:
* add method to save with conflict handler.

Ref: #2429

### DIFF
--- a/Objective-C/CBLDatabase.mm
+++ b/Objective-C/CBLDatabase.mm
@@ -233,7 +233,6 @@ static void dbObserverCallback(C4DatabaseObserver* obs, void* context) {
                                     error: &err];
         // if it's a conflict, we will use the conflictHandler to resolve.
         if (!success && $equal(err.domain, CBLErrorDomain) && err.code == CBLErrorConflict) {
-            
             CBLDocument* latestDoc;
             CBL_LOCK(self) {
                 C4Transaction transaction(_c4db);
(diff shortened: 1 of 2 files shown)


Objective-C/CBLDatabase.mm
```diff
@@ -219,6 +219,60 @@ - (BOOL) saveDocument: (CBLMutableDocument*)document
...
+- (BOOL) saveDocument: (CBLMutableDocument *)document
+      conflictHandler: (BOOL (^)(CBLMutableDocument *, CBLDocument * nullable))conflictHandler
+                error: (NSError **)error {
+    
+    CBLAssertNotNil(document);
+    CBLAssertNotNil(conflictHandler);
+    
+    while (true) {
+        NSError* err;
+        BOOL success = [self saveDocument: document
```
pasin: When saving the document again after handling the conflict, the document should be saved on top of the oldDoc (as base).

Objective-C/CBLDatabase.mm
```diff
@@ -213,12 +213,69 @@ - (BOOL) saveDocument: (CBLMutableDocument*)document
...
+                                                         error: error];
+                if (!oldDoc)
+                    return NO;
+                
+                if (!transaction.commit())
+                    return convertError(transaction.error(), error);
+            }
+            
+            @try {
+                if (conflictHandler(document, oldDoc)) {
```
pasin: We also need to check if oldDoc is a deleted doc or not. If it's a deleted doc, the null value should be given to conflictHandler.

jayahariv: done

PR #2435, opened 2019-05-01
feat: database save conflict handler implementation in swift

jayahariv:
* save conflict method calleed from swift
* implemented the save conflict method in swift
* not tested

Ref: #2429


Swift/Database.swift
```diff
@@ -118,6 +118,35 @@ public final class Database {
...
+    /// - Throws: An error on a failure.
+    @discardableResult public func saveDocument(
+        _ document: MutableDocument, conflictHandler: @escaping (MutableDocument, Document?) -> Bool
+        ) throws -> Bool {
+        do {
+            let conflictHandler = { (current: CBLMutableDocument, old: CBLDocument?) -> Bool in
+                if let oldDoc = old {
+                    return conflictHandler(MutableDocument(current), Document(oldDoc))
+                }
+                return conflictHandler(MutableDocument(current), nil)
```
pasin: Can you simplify this? I think we can have only one return.

pasin: I think MutableDocument(current) is not correct. It should be just document.

Swift/Database.swift
```diff
@@ -118,6 +118,35 @@ public final class Database {
...
+            }
+            
+            try _impl.save(document._impl as! CBLMutableDocument, conflictHandler: conflictHandler)
+            return true
+        } catch let err as NSError {
+            throw err
```
pasin: The logic here should be the same as the saveDocument() above.

Swift/Database.swift
jayahariv: `func handle(document: MutableDocument, oldDocument: Document?) -> Bool`??

Swift/Database.swift
jayahariv: doubt: if it is already handling the conflict, is there a chance for getting the `CBLErrorConflict`? exceptions inside conflict handler will also not return the CBLErrorConflict right.

Swift/Database.swift
```diff
@@ -118,6 +118,36 @@ public final class Database {
...
+    /// - Returns: True if successful. False on failure
+    /// - Throws: An error on a failure.
+    @discardableResult public func saveDocument(
+        _ document: MutableDocument, conflictHandler: @escaping (MutableDocument, Document?) -> Bool
+        ) throws -> Bool {
+        do {
+            try _impl.save(
+                document._impl as! CBLMutableDocument,
+                conflictHandler: { (cur: CBLMutableDocument, old: CBLDocument?) -> Bool in
+                    return conflictHandler(MutableDocument(cur), old != nil ? Document(old!) : nil)
```
pasin: Instead of MutableDocument(cur), use the original `document` object itself. We want it to be the same document object that users save.

Swift/Database.swift
```diff
@@ -118,6 +118,36 @@ public final class Database {
...
+        _ document: MutableDocument, conflictHandler: @escaping (MutableDocument, Document?) -> Bool
+        ) throws -> Bool {
+        do {
+            try _impl.save(
+                document._impl as! CBLMutableDocument,
+                conflictHandler: { (cur: CBLMutableDocument, old: CBLDocument?) -> Bool in
+                    return conflictHandler(document, old != nil ? Document(old!) : nil)
+                }
+            )
+            return true
```
pasin: I think it needs to return the result from `_impl.save()` instead return `true`. Sorry that I didn't notice this before.

pasin: NVM, this is already correct.

PR #2436, opened 2019-05-01
doc: update script to reflect the latest README

jayahariv:
not tested, ideally this will use the latest README from the repo instead of the `Scripts/Support/Docs/README.md`. 

do you remember any reason why we use separate README?


pasin:
I don't remember why we need two different one. Please remove the readme in Scripts/Support/Docs/README.md as well.

Issue #2437, opened 2019-05-01
Replicator could take a long time to close when network connection is bad


pasin:
This is a ticket to investigate the behavior based on[ CM-163](https://issues.couchbase.com/browse/CM-163).

PR #2447, opened 2019-05-06
Improved replicator timeouts

snej:
LiteCore is handling WebSocket timeouts now. It will time out:
* if connection doesn’t open in 15 sec
* if no PONG response to PING received in 10 sec
* if no response received to CLOSE in 5 sec
The last of these should fix the problem of replicators not closing when told to stop.

Note: The fixes are in LiteCore; the CBL code change is to _remove_ the existing timeout code in CBLWebSocket, since it's now redundant.


snej:
Oops, this is for #2437.

PR #2456, opened 2019-05-15
unit: verify the sequence-id to check whether to push resolved 

jayahariv:
* revision-id for the remote wins should always be same.
* sequence should always be remote greater than local doc.

### DIFF
--- a/Objective-C/Tests/ReplicatorTest+CustomConflict.m
+++ b/Objective-C/Tests/ReplicatorTest+CustomConflict.m
@@ -18,6 +18,7 @@
 //
 
 #import "ReplicatorTest.h"
+#import "CBLDocument+Internal.h"
 
 @interface TestConflictResolver: NSObject<CBLConflictResolver>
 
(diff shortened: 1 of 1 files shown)


Objective-C/Tests/ReplicatorTest+CustomConflict.m
```diff
@@ -139,7 +147,13 @@ - (void) testConflictResolverLocalWins {
...
     // Check that it was resolved:
     AssertEqual(self.db.count, 1u);
-    AssertEqualObjects([self.db documentWithID: docId].toDictionary, localData);
+    CBLDocument* localSavedDoc = [self.db documentWithID: docId];
+    CBLDocument* remoteSavedDoc = [otherDB documentWithID: docId];
+    AssertEqualObjects(localSavedDoc.toDictionary, localData);
+    
+    // validate the local doc is ready to push to remote
+    Assert(localSavedDoc.revID != remoteSavedDoc.revID);
+    Assert(localSavedDoc.sequence > remoteSavedDoc.sequence);
```
snej: It doesn't make any sense to compare sequence numbers between two databases; those numbers are purely local.

To check that the local doc will be pushed, you should check that its sequence is greater than the sequence number it had when the last replication finished. (You actually want to compare it against the replicator's local checkpoint, but there's no API to get that.)

jayahariv: I could see in both remote win and local win, sequence of the document is increased than the last replication sequence. In local win, sequence increase by two and if remote win, which increase by one. 

Guess this cannot be verified through local-db sequence number then. We can do it by pushing the local to the remote and check whether the sequence number in remote has changed or not. which seems like not changed in case of remote win, and increase by one in case of local win.

Objective-C/Tests/ReplicatorTest+CustomConflict.m
```diff
@@ -160,8 +178,14 @@ - (void) testConflictResolverNullDoc {
     
     // Check whether the document is deleted, and returns null.
     AssertEqual(self.db.count, 0u);
-    CBLDocument* savedDoc = [self.db documentWithID: @"doc"];
-    AssertNil(savedDoc);
+    NSError* error;
+    UInt64 sequenceBeforePush = [otherDB documentWithID: docId].sequence;
+    [self run: [self config: kCBLReplicatorTypePush] errorCode: 0 errorDomain: nil];
```
jayahariv: this push will crash due to #2470

jayahariv: fixed!!

Issue #2428, opened 2019-04-25
Replicator Conflict Resolver Unit Tests


jayahariv:
- [x] Make sure the resolver is called when setting a resolver in the ReplicatorConfiguration, and saved the correct document based on what the resolver returns: local document, remote document, new document, and null document (deleted).
- [x] Make sure the default conflict resolve is used if the resolver is not set in the ReplicatorConfiguration.
- [x] Make sure the conflict resolution property is readonly after the config is set. (May be adding one more field to existing test )
- [x] When resolve causes an exception, make sure the replicator gets stopped. And make sure it gets resolved on next time. 
- [x] Make sure to throw an exception when a document with wrong ID is returned.
- [x] Make sure to throw an exception when a document from a different database instance is returned.

PR #2457, opened 2019-05-16
unit: add exception when resolution cases

jayahariv:
* all are disabled, since the test needs the resolving when the replicator restarts, which is not yet done. Will enable when it is ready. 
* exception from resolve method seems to be causing memory leaks, assuming this is due to the @try @catch in objc causing memory issues. Need to discuss this. 

Ref: #2428


Objective-C/Tests/ReplicatorTest+CustomConflict.m
```diff
@@ -233,6 +233,129 @@ - (void) testConflictResolverMergeDoc {
...
+    resolver = [[TestConflictResolver alloc] initWithResolver: ^CBLDocument* (CBLConflict* con) {
+        return con.remoteDocument;
+    }];
+    pullConfig.conflictResolver = resolver;
+    [self run: pullConfig errorCode: 0 errorDomain: nil];
+    AssertEqualObjects([self.db documentWithID: docId].toDictionary, remoteData);
+}
+
+// TODO: disable test since the unresolved conflict side is still under development.
+- (void) _testConflictResolverThrowingException {
```
jayahariv: this test is causing memory leak. which says the [localDoc](https://github.com/couchbase/couchbase-lite-ios/blob/master/Objective-C/CBLDatabase.mm#L1028) and [remoteDoc](https://github.com/couchbase/couchbase-lite-ios/blob/master/Objective-C/CBLDatabase.mm#L1034) which are created is not getting removed form memory. Tried assigning them in the [catch](https://github.com/couchbase/couchbase-lite-ios/blob/master/Objective-C/CBLDatabase.mm#L1068), as well as finally, but no luck. Is this due to objc program interruption and causing ARC not to release some scope variables?

snej: No, ARC should work correctly with exceptions.
You can find what's still referencing those objects by turning on Malloc Stack logging (in the scheme 'Run' panel, Diagnostics tab, you have to turn off sanitizers first). Set a breakpoint after the time the objects are leaked, then in the Debug navigator set the rightmost popup to "View Memory Graph Hierarchy". Then you can find instances of CBLDocument and it'll show you a graph of which other objects reference them.

jayahariv: used the technique, and could see that the CBLConflict and its properties(localDocument and remoteDocument) are causing it. CBLConflict was holding a strong reference to them, which was not released in case of exception(which i am still not clear why only for exception, since it is not happening in other unit tests).

I made them `__weak`, so now memory issue is gone.

Objective-C/Tests/ReplicatorTest+CustomConflict.m
```diff
@@ -233,6 +233,127 @@ - (void) testConflictResolverMergeDoc {
...
+    // should be solved when the replicator runs next time!!
+    resolver = [[TestConflictResolver alloc] initWithResolver: ^CBLDocument* (CBLConflict* con) {
+        return con.remoteDocument;
+    }];
+    pullConfig.conflictResolver = resolver;
+    [self run: pullConfig errorCode: 0 errorDomain: nil];
+    AssertEqualObjects([self.db documentWithID: docId].toDictionary, remoteData);
+}
+
+// TODO: enable this and handle known memory leak expectations
```
```suggestion
// TODO: enable this and handle expected memory leak in tests. 
```

Objective-C/Tests/ReplicatorTest+CustomConflict.m
jayahariv: making `__weak` seems to be wrong. so updated with strong. and disabled test since the objc seems to be not exception-safe. which is ignoring the retained variable when exception happens. Added a TODO in the so that, if we can handle an expected memory warning for any future test. 

[clang.llvm.org - doc](http://clang.llvm.org/docs/AutomaticReferenceCounting.html#exceptions)

PR #2458, opened 2019-05-16
unit: check replicator listener events

jayahariv:
* when conflicts happens, make sure there is only single event is fired. 
* validated when exception happens and also when resolution succeeds. 

Ref: #2428


Objective-C/Tests/ReplicatorTest+CustomConflict.m
```diff
@@ -233,6 +233,48 @@ - (void) testConflictResolverMergeDoc {
     AssertEqualObjects(savedDoc.toDictionary, exp);
 }
 
+- (void) testDocumentReplicationEventForConflictedDocs {
+    TestConflictResolver* resolver;
+    
+    // when resolution throws exception
```
snej: It doesn't throw an exception, it just returns nil...

```suggestion
    // when document with wrong-id is returned, it will throw exception.
```

jayahariv: it throws an exception, when we supply a wrong doc id. updated the comment

Objective-C/Tests/ReplicatorTest+CustomConflict.m
```diff
@@ -259,12 +259,55 @@ - (void) testConflictResolverMergeDoc {
...
+    [replicator removeChangeListenerWithToken: token];
+}
+
 - (void) testConflictResolverCalledTwice {
     NSString* docId = @"doc";
     NSDictionary* localData = @{@"key1": @"value1"};
     NSDictionary* remoteData = @{@"key2": @"value2"};
     [self makeConflictFor: docId withLocal: localData withRemote: remoteData];
     
+
```
```suggestion
```

PR #2462, opened 2019-05-24
unit: non-blocking resolver

jayahariv:
* validate the non blocking by making the first document to sleep.
* then second document should resolve in between.
* order should be doc1(whichever is resolved first) will start, doc2(whichever resolves second) will start, doc2 ends, doc1 ends.

ref:  #2428

### DIFF
--- a/Objective-C/Tests/ReplicatorTest+CustomConflict.m
+++ b/Objective-C/Tests/ReplicatorTest+CustomConflict.m
@@ -519,7 +519,7 @@
     [self makeConflictFor: @"doc2" withLocal: localData withRemote: remoteData];
     
     TestConflictResolver* resolver;
-    CBLReplicatorConfiguration* pullConfig = [self pullConfig];
+    CBLReplicatorConfiguration* pullConfig = [self config: kCBLReplicatorTypePull];
     
     NSMutableArray* order = [NSMutableArray array];
     resolver = [[TestConflictResolver alloc] initWithResolver: ^CBLDocument* (CBLConflict* con) {
@@ -528,10 +528,11 @@
         }
         if (order.count == 1) {
             [NSThread sleepForTimeInterval: 0.5];
-            [ex fulfill];
         }
-        @synchronized (order) {
-            [order addObject: con.localDocument.id];
+        [order addObject: con.localDocument.id];
+        
+        if (order.count == 4) {
+            [ex fulfill];
         }
         return con.remoteDocument;
     }];
--- a/Objective-C/Tests/ReplicatorTest+CustomConflict.m
+++ b/Objective-C/Tests/ReplicatorTest+CustomConflict.m
@@ -263,7 +263,7 @@
     [self waitForExpectationsWithTimeout: 5 handler: nil];
     
     // make sure, first doc starts resolution but finishes last.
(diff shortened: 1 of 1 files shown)


Objective-C/Tests/ReplicatorTest+CustomConflict.m
```diff
@@ -233,6 +233,37 @@ - (void) testConflictResolverMergeDoc {
...
+    [self makeConflictFor: @"doc1" withLocal: localData withRemote: remoteData];
+    [self makeConflictFor: @"doc2" withLocal: localData withRemote: remoteData];
+    
+    TestConflictResolver* resolver;
+    CBLReplicatorConfiguration* pullConfig = [self pullConfig];
+    
+    NSMutableArray* order = [NSMutableArray array];
+    resolver = [[TestConflictResolver alloc] initWithResolver: ^CBLDocument* (CBLConflict* con) {
+        [order addObject: con.localDocument.id];
+        if ([con.localDocument.id isEqualToString: order.firstObject]) {
```
snej: I think it would be clearer just to say `if (order.count == 1)`.

Objective-C/Tests/ReplicatorTest+CustomConflict.m
```diff
@@ -233,6 +233,37 @@ - (void) testConflictResolverMergeDoc {
...
+    NSDictionary* remoteData = @{@"key2": @"value2"};
+    [self makeConflictFor: @"doc1" withLocal: localData withRemote: remoteData];
+    [self makeConflictFor: @"doc2" withLocal: localData withRemote: remoteData];
+    
+    TestConflictResolver* resolver;
+    CBLReplicatorConfiguration* pullConfig = [self pullConfig];
+    
+    NSMutableArray* order = [NSMutableArray array];
+    resolver = [[TestConflictResolver alloc] initWithResolver: ^CBLDocument* (CBLConflict* con) {
+        [order addObject: con.localDocument.id];
```
snej: NSMutableArray is not thread-safe, so you need to synchronize accesses to`order` inside the resolver block.

Objective-C/Tests/ReplicatorTest+CustomConflict.m
```diff
@@ -233,6 +233,41 @@ - (void) testConflictResolverMergeDoc {
...
+    resolver = [[TestConflictResolver alloc] initWithResolver: ^CBLDocument* (CBLConflict* con) {
+        @synchronized (order) {
+            [order addObject: con.localDocument.id];
+        }
+        if (order.count == 1) {
+            [NSThread sleepForTimeInterval: 0.5];
+            [ex fulfill];
+        }
+        @synchronized (order) {
+            [order addObject: con.localDocument.id];
```
pasin: Could there be a race b/w adding the docID here and checking order assertion after the expectation is fulfill?

jayahariv: you are right, there is no need for this. But now i wonder, whether we should fulFill the expectation after adding the last docID.

jayahariv: udpated

PR #2463, opened 2019-05-24
unit: default resolution check

jayahariv:
* check for generation, revID and deleted scenarios with default resolution.

Ref: #2428

### DIFF
--- a/Objective-C/Tests/ReplicatorTest+CustomConflict.m
+++ b/Objective-C/Tests/ReplicatorTest+CustomConflict.m
@@ -466,7 +466,6 @@
             AssertNil(error);
             Assert(successDoc.isDeleted);
         } else {
-            
             // if generations are different
             AssertEqual(doc1.generation, [self.db documentWithID: doc1.id].generation);
             Assert(doc1.generation > doc2.generation);
--- a/Objective-C/Tests/ReplicatorTest+CustomConflict.m
+++ b/Objective-C/Tests/ReplicatorTest+CustomConflict.m
@@ -438,15 +438,13 @@
     docID = @"doc2";
     [self makeConflictFor: docID withLocal: localData withRemote: remoteData];
     [self.db deleteDocument: [self.db documentWithID: docID] error: &error];
-    [conflictedDocs addObject: @[[NSNull null],
-                                 [otherDB documentWithID: docID]]];
+    [conflictedDocs addObject: @[[NSNull null], [otherDB documentWithID: docID]]];
     
     // delete remote
     docID = @"doc4";
     [self makeConflictFor: docID withLocal: localData withRemote: remoteData];
     [otherDB deleteDocument: [otherDB documentWithID: docID] error: &error];
-    [conflictedDocs addObject: @[[self.db documentWithID: docID],
-                                 [NSNull null]]];
+    [conflictedDocs addObject: @[[self.db documentWithID: docID], [NSNull null]]];
     
     
     CBLReplicatorConfiguration* pullConfig = [self pullConfig];
--- a/Objective-C/Tests/ReplicatorTest+CustomConflict.m
+++ b/Objective-C/Tests/ReplicatorTest+CustomConflict.m
@@ -438,7 +438,7 @@
     docID = @"doc2";
     [self makeConflictFor: docID withLocal: localData withRemote: remoteData];
     [self.db deleteDocument: [self.db documentWithID: docID] error: &error];
-    [conflictedDocs addObject: @[[self.db documentWithID: docID] ?: [NSNull null],
+    [conflictedDocs addObject: @[[NSNull null],
                                  [otherDB documentWithID: docID]]];
     
     // delete remote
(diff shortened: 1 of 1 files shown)


Objective-C/Tests/ReplicatorTest+CustomConflict.m
```diff
@@ -82,13 +84,19 @@ - (void) makeConflictFor: (NSString*)docID
...
     [doc1 setData: localData];
-    Assert([self.db saveDocument: doc1 error: &error]);
     
     // pass the remote revision
     CBLMutableDocument* doc2 = [[otherDB documentWithID: docID] toMutable];
     Assert(doc2);
     [doc2 setData: remoteData];
-    Assert([otherDB saveDocument: doc2 error: &error]);
+    
+    if (localFirst) {
```
snej: What's this for? It shouldn't make any difference at all which order they're saved in, since they're in different databases. The revIDs and sequences are the same in either order.

Objective-C/Tests/ReplicatorTest+CustomConflict.m
```diff
@@ -233,6 +241,74 @@ - (void) testConflictResolverMergeDoc {
...
+                                                                      error: &error];
+            AssertNil(error);
+            Assert(successDoc.isDeleted);
+            continue;
+        }
+        
+        if (doc1.generation != doc2.generation)
+            Assert(doc1.generation > doc2.generation);
+        else
+            Assert([doc1.revID compare: doc2.revID]);
```
snej: This is just checking that their revIDs aren't equal; is that what you meant? If so, calling `-isEqualToString:` would be clearer. Or did you mean to test whether doc1.revID is ≥ doc2.revID?

jayahariv: doc1.revID is > doc2.revID

Objective-C/Tests/ReplicatorTest+CustomConflict.m
jayahariv: i am assuming if we save a document second, its revID will be greater than the previous(lexical order). hoping some kind of logic to link the revID with timestamp. 

My tests are working fine. unless it might fail in different machine.

jayahariv: my assumptions were wrong, got it clarified now. Cool, I will re-write this.

Objective-C/Tests/ReplicatorTest+CustomConflict.m
jayahariv: since this is arbitrary, this is useless comparison.

jayahariv:
I was under the impression the default conflict resolver logic from past, was using last-write-wins. And the revIDs were created with some timestamp. Anyways, all got corrected. 

Now verifying 
(a) doc with revisions of different generation, and
(b) any one of the revision is deleted.

Objective-C/Tests/ReplicatorTest+CustomConflict.m
```diff
@@ -418,6 +419,60 @@ - (void) _testConflictResolverThrowingException {
...
+    [doc setValue: @"value3" forKey: @"key3"];
+    [self saveDocument: doc];
+    [conflictedDocs addObject: @[[self.db documentWithID: docID],
+                                 [otherDB documentWithID: docID]]];
+    
+    // delete local
+    docID = @"doc2";
+    [self makeConflictFor: docID withLocal: localData withRemote: remoteData];
+    [self.db deleteDocument: [self.db documentWithID: docID] error: &error];
+    [conflictedDocs addObject: @[[self.db documentWithID: docID] ?: [NSNull null],
```
pasin: Is there a case that [self.db documentWithID: docID] is not null here?

Objective-C/Tests/ReplicatorTest+CustomConflict.m
```diff
@@ -418,6 +419,60 @@ - (void) _testConflictResolverThrowingException {
...
+                                 [otherDB documentWithID: docID]]];
+    
+    // delete remote
+    docID = @"doc4";
+    [self makeConflictFor: docID withLocal: localData withRemote: remoteData];
+    [otherDB deleteDocument: [otherDB documentWithID: docID] error: &error];
+    [conflictedDocs addObject: @[[self.db documentWithID: docID],
+                                 [otherDB documentWithID: docID] ?: [NSNull null]]];
```
pasin: Is there a case that [otherDB documentWithID: docID] is not null here?

Objective-C/Tests/ReplicatorTest+CustomConflict.m
jayahariv: No there is not. we can directly assign null

```suggestion
    [conflictedDocs addObject: @[[NSNull null],
```

Objective-C/Tests/ReplicatorTest+CustomConflict.m
jayahariv: same. we can directly assign null.

jayahariv:
```suggestion
                                 [NSNull null]]];
```

Issue #2430, opened 2019-04-25
Database Save Conflict Handler Unit Tests


jayahariv:
- [x] Check whether the save conflict is calling the conflict handler and saving the appropriate document.
- [x] ConflictHandler returning true and false, making sure, both are handled correctly. 
- [x] After one conflict is resolved, there could be another conflict happen. Make sure that the conflict handler is called again.

PR #2464, opened 2019-05-24
unit: test conflict resolver

jayahariv:
* check if return true and when some updates are done.

ref: #2430

### DIFF
--- a/Objective-C/Tests/DatabaseTest.m
+++ b/Objective-C/Tests/DatabaseTest.m
@@ -582,6 +582,59 @@
     [self cleanDB];
 }
 
+#pragma mark Save Conflict Handler
+
+- (void) testConflictHandler {
+    NSString* docID = @"doc1";
+    CBLMutableDocument* doc = [[CBLMutableDocument alloc] initWithID: docID];
+    [doc setString: @"Tiger" forKey: @"firstName"];
+    [self saveDocument: doc];
+    AssertEqual([self.db documentWithID: docID].generation, 1u);
+    
+    CBLMutableDocument* doc1a = [[self.db documentWithID: docID] toMutable];
+    CBLMutableDocument* doc1b = [[self.db documentWithID: docID] toMutable];
+    
+    [doc1a setString: @"Scotty" forKey: @"nickName"];
+    [self saveDocument: doc1a];
+    AssertEqual([self.db documentWithID: docID].generation, 2u);
+    
+    NSError* error;
+    [doc1b setString: @"Scott" forKey: @"nickName"];
+    [self.db saveDocument: doc1b
+          conflictHandler:^BOOL(CBLMutableDocument * cur, CBLDocument * old) {
+              AssertEqualObjects(doc1b.toDictionary, cur.toDictionary);
+              AssertEqualObjects(doc1a.toDictionary, old.toDictionary);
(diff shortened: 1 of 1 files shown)


Objective-C/Tests/DatabaseTest.m
```diff
@@ -582,6 +582,59 @@ - (void) testSaveDocWithDeletedConflictUsingConcurrencyControl: (int)concurrency
...
+    [doc1b setString: @"Scott" forKey: @"nickName"];
+    [self.db saveDocument: doc1b
+          conflictHandler:^BOOL(CBLMutableDocument * cur, CBLDocument * old) {
+              AssertEqualObjects(doc1b.toDictionary, cur.toDictionary);
+              AssertEqualObjects(doc1a.toDictionary, old.toDictionary);
+              AssertEqual(cur.generation, 2u);
+              AssertEqual(old.generation, 2u);
+              return YES;
+          } error: &error];
+    AssertNil(error);
```
snej: Again, test the return value of `saveDocument`, don't test `error`.

jayahariv: intention here is to check return value of the `saveDocument` and error = nil. So that no unnecessary error is returned in case of success. 

```
NSError* error = nil;
....
Assert([save document: &error]);
AssertNil(error);
```

Objective-C/Tests/DatabaseTest.m
```diff
@@ -582,6 +582,58 @@ - (void) testSaveDocWithDeletedConflictUsingConcurrencyControl: (int)concurrency
...
+    CBLMutableDocument* doc1b = [[self.db documentWithID: docID] toMutable];
+    
+    [doc1a setString: @"Scotty" forKey: @"nickName"];
+    [self saveDocument: doc1a];
+    AssertEqual([self.db documentWithID: docID].generation, 2u);
+    
+    NSError* error;
+    [doc1b setString: @"Scott" forKey: @"nickName"];
+    Assert([self.db saveDocument: doc1b
+                 conflictHandler:^BOOL(CBLMutableDocument * cur, CBLDocument * old) {
```
pasin: Change `cur` to doc to match the API.

pasin: Also check if cur is the same instance as doc1b.

Objective-C/Tests/DatabaseTest.m
```diff
@@ -582,6 +582,60 @@ - (void) testSaveDocWithDeletedConflictUsingConcurrencyControl: (int)concurrency
...
+    
+    [doc1a setString: @"Scotty" forKey: @"nickName"];
+    [self saveDocument: doc1a];
+    AssertEqual([self.db documentWithID: docID].generation, 2u);
+    
+    NSError* error;
+    [doc1b setString: @"Scott" forKey: @"nickName"];
+    Assert([self.db saveDocument: doc1b
+                 conflictHandler:^BOOL(CBLMutableDocument * document, CBLDocument * old) {
+                     AssertEqualObjects(doc1b, document);
```
pasin: I think we can just do Assert(doc1b == document). I understand that using AssertEqualObjects will call equals method?

jayahariv: Yes, true. Agree.

PR #2465, opened 2019-05-24
unit: cancel conflict handler

jayahariv:
* check with/without update to the current document if we return false from the conflict handler, how the save behaves.

ref: #2430

### DIFF
--- a/Objective-C/Tests/DatabaseTest.m
+++ b/Objective-C/Tests/DatabaseTest.m
@@ -599,9 +599,9 @@
     NSError* error;
     [doc1b setString: @"Scott" forKey: @"nickName"];
     AssertFalse([self.db saveDocument: doc1b
-                      conflictHandler:^BOOL(CBLMutableDocument * cur, CBLDocument * old) {
-                          AssertEqualObjects(doc1b.toDictionary, cur.toDictionary);
-                          AssertEqualObjects(doc1a.toDictionary, old.toDictionary);
+                      conflictHandler:^BOOL(CBLMutableDocument * savingDoc, CBLDocument * dbDoc) {
+                          AssertEqualObjects(doc1b.toDictionary, savingDoc.toDictionary);
+                          AssertEqualObjects(doc1a.toDictionary, dbDoc.toDictionary);
                           return NO;
                       } error: &error]);
     AssertEqual(error.code, CBLErrorConflict);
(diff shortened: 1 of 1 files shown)


Objective-C/Tests/DatabaseTest.m
```diff
@@ -582,6 +582,57 @@ - (void) testSaveDocWithDeletedConflictUsingConcurrencyControl: (int)concurrency
...
+    CBLMutableDocument* doc1b = [[self.db documentWithID: docID] toMutable];
+    
+    [doc1a setString: @"Scotty" forKey: @"nickName"];
+    [self saveDocument: doc1a];
+    AssertEqual([self.db documentWithID: docID].generation, 2u);
+    
+    NSError* error;
+    [doc1b setString: @"Scott" forKey: @"nickName"];
+    [self.db saveDocument: doc1b
+          conflictHandler:^BOOL(CBLMutableDocument * cur, CBLDocument * old) {
```
snej: The names `cur` and `old` confused me -- `old` isn't old, it's the newer revision in the database. So I had them backwards. Maybe `savingDoc` and `dbDoc` or something like that?

Objective-C/Tests/DatabaseTest.m
```diff
@@ -582,6 +582,57 @@ - (void) testSaveDocWithDeletedConflictUsingConcurrencyControl: (int)concurrency
...
+              AssertEqualObjects(doc1b.toDictionary, cur.toDictionary);
+              AssertEqualObjects(doc1a.toDictionary, old.toDictionary);
+              return NO;
+          } error: &error];
+    AssertNil(error);
```
snej: Isn't `saveDocument` supposed to return a Conflict error if the handler returns NO?

Also, testing `error` is the wrong way to check for success/failure. In Cocoa APIs, the return value indicates success/failure. The NSError has a value only on failure; on success _its value is undefined_. (Yeah, you initialized it to nil, but it's possible that the method you called used it itself for other calls that failed, but it recovered from those and returned YES, but left the error non-nil.)

jayahariv: will assert the `db saveDocument` method. But should we also make sure, the provided error is empty when success

jayahariv: yea, this was wrong, checking error as empty. which is corrected.

Objective-C/Tests/DatabaseTest.m
jayahariv: As per the design spec this variables are `document` and `oldDocument`. We can discuss about this when Pasin is back.

jayahariv: corrected to use the design-spec, with `document` and `oldDocument`

PR #2466, opened 2019-05-24
unit: deleted old doc

jayahariv:
* keeping the new doc when old doc is deleted.
* keeping the old deleted doc, and ignoring the new changes.

ref:  #2430

### DIFF
--- a/Objective-C/Tests/DatabaseTest.m
+++ b/Objective-C/Tests/DatabaseTest.m
@@ -595,12 +595,12 @@
     
     NSError* error;
     [doc1b setString: @"value1" forKey: @"key1"];
-    [self.db saveDocument: doc1b
-          conflictHandler:^BOOL(CBLMutableDocument * cur, CBLDocument * old) {
-              AssertNil(old);
-              AssertNotNil(cur);
-              return YES;
-          } error: &error];
+    Assert([self.db saveDocument: doc1b
+                 conflictHandler:^BOOL(CBLMutableDocument * cur, CBLDocument * old) {
+                     AssertNil(old);
+                     AssertNotNil(cur);
+                     return YES;
+                 } error: &error]);
     AssertNil(error);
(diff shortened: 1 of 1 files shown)


Objective-C/Tests/DatabaseTest.m
```diff
@@ -582,6 +582,48 @@ - (void) testSaveDocWithDeletedConflictUsingConcurrencyControl: (int)concurrency
...
+    
+    NSError* error;
+    [doc1b setString: @"value1" forKey: @"key1"];
+    Assert([self.db saveDocument: doc1b
+                 conflictHandler:^BOOL(CBLMutableDocument * cur, CBLDocument * old) {
+                     AssertNil(old);
+                     AssertNotNil(cur);
+                     return YES;
+                 } error: &error]);
+    AssertNil(error);
```
snej: Take this out -- on success, the value of `error` is undefined.

Objective-C/Tests/DatabaseTest.m
```diff
@@ -582,6 +582,48 @@ - (void) testSaveDocWithDeletedConflictUsingConcurrencyControl: (int)concurrency
...
+    [self deleteDocument: doc1a concurrencyControl: kCBLConcurrencyControlLastWriteWins];
+    
+    [doc1b setString: @"value2" forKey: @"key2"];
+    AssertFalse([self.db saveDocument: doc1b
+                      conflictHandler:^BOOL(CBLMutableDocument * cur, CBLDocument * old) {
+                          AssertNil(old);
+                          AssertNotNil(cur);
+                          return NO;
+                      } error: &error]);
+    AssertNil(error);
```
snej: If `saveDocument` returned NO, then `error` should be non-nil (it should be a conflict error.)

Objective-C/Tests/DatabaseTest.m
jayahariv: ok

PR #2467, opened 2019-05-24
fix: on save handler false - return conflict error

jayahariv:
* on save handler false - return conflict error
* it was not added in objc. 

ref: #2429

### DIFF
--- a/Objective-C/CBLDatabase.mm
+++ b/Objective-C/CBLDatabase.mm
@@ -259,6 +259,8 @@ static void dbObserverCallback(C4DatabaseObserver* obs, void* context) {
                 if (conflictHandler(document, oldDoc.isDeleted ? nil : oldDoc)) {
                     CBLAssertNotNil(document);
                     continue;
+                } else {
+                    return createError(CBLErrorConflict, error);
                 }
             } @catch(NSError* conflictHandlerError) {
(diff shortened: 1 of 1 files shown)


Objective-C/CBLDatabase.mm
```diff
@@ -259,6 +259,8 @@ - (BOOL) saveDocument: (CBLMutableDocument *)document
                 if (conflictHandler(document, oldDoc.isDeleted ? nil : oldDoc)) {
                     CBLAssertNotNil(document);
```
snej: Why this assertion? The conflictHandler can't change the value of `document`.

jayahariv: It was added to make sure, the document is never released or not became dangling pointer. in which case, this will fail instead of other unexpected places.

snej: Nothing's changing the value of `document`, so it still has whatever value it originally had. It can't possibly be nil unless it started out that way. (Deallocing the _object_ doesn't affect a _pointer_ to the object.)

(Actually I guess there is the special case of a weak pointer, which gets set to nil if the object is dealloced. But that's not happening here.)

jayahariv: done!

PR #2477, opened 2019-05-31
unit: resolver called twice

jayahariv:
* swift test to validate when resolver is conflicted and called twice working fine.
* added helper methods which needs to be removed before merging. or atleast one of these set of methods should be merged.

ref: #2428


Swift/Tests/ReplicatorTest.swift
```diff
@@ -840,5 +839,6 @@ class ReplicatorTest_Main: ReplicatorTest {
         XCTAssertEqual(otherDB.count, 3)
         XCTAssertEqual(db.count, 2)
     }
+
```
jayahariv:
```suggestion
```

PR #2478, opened 2019-05-31
unit: exception cases handled

jayahariv:
* exception cases handled with wrong-doc-id, different-db, and exception thrown inside custom conflict method.
* unlike objc, swift is not showing memory issue with conflict handler throwing method.

ref: #2428


Swift/Tests/ReplicatorTest.swift
```diff
@@ -839,6 +839,5 @@ class ReplicatorTest_Main: ReplicatorTest {
         XCTAssertEqual(otherDB.count, 3)
         XCTAssertEqual(db.count, 2)
     }
```
jayahariv:
```suggestion
    }
    
```

PR #2479, opened 2019-05-31
unit: conflict handler called twice

jayahariv:
* add conflict handler to be called twice test.
* confusion on either naming, may be the old is not old, as Jens suggested. since this test wil confuse. and cur is not cur, but the savedDoc.

ref: #2430

### DIFF
--- a/Objective-C/Tests/DatabaseTest.m
+++ b/Objective-C/Tests/DatabaseTest.m
@@ -686,6 +686,58 @@
     AssertEqual(error.code, CBLErrorConflict);
 }
 
+- (void) testConflictHandlerCalledTwice {
+    NSString* docID = @"doc1";
+    CBLMutableDocument* doc = [[CBLMutableDocument alloc] initWithID: docID];
+    [doc setString: @"Tiger" forKey: @"firstName"];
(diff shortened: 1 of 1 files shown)


Objective-C/Tests/DatabaseTest.m
```diff
@@ -784,6 +784,59 @@ - (void) testDeleteDocWithConflictUsingConcurrencyControl: (int)concurrencyContr
...
+                         AssertEqual(old.generation, 2u);
+                         AssertEqual(cur.generation, 2u);
+                         [doc1c setBoolean: YES forKey: @"secondUpdate"];
+                         [self saveDocument: doc1c];
+                         AssertEqual([self.db documentWithID: docID].generation, 3u);
+                     }
+                     
+                     // "secondUpdate" is going to be erased, since the cur doesn't have the latest
+                     // doc from DB. We will be merging the old and cur here...
+                     // whether this should be due to variable naming may be!!!!
```
jayahariv: attention: this might be something to notice.

jayahariv: discussed, and this assumption is right. Issue is with the naming of parameter.

PR #2483, opened 2019-06-04
fix: lock result set

jayahariv:
* previous result set needs to be locked when copying, else end up returning corrupted list.


jayahariv:
RS = rs1

- threadA: updates rs1 to rs2
- threadB: updates rs1 to rs3

Both gets the RS = rs1 and first one enters write lock and second one waits for the first to finish writing.

RS will be one executing last. It will result in rs2, or rs3. 

But if the rs.refresh() will give the latest. This won't be an issue.

jayahariv:
seems like this is not required. Since the same c4enum will return the latest from the db. 

```
    CBL_LOCK(db) {
        newEnum = c4queryenum_refresh(_c4enum, &c4error);
    }
```

PR #2485, opened 2019-06-07
unit: swift unit tests for replication events

jayahariv:
* validated whether there is only single events are fired when document replication custom conflict resolution happens, and when CCR exception is thrown(skips the ccr).


Swift/Tests/ReplicatorTest+CustomConflict.swift
```diff
@@ -305,6 +305,49 @@ class ReplicatorTest_CustomConflict: ReplicatorTest {
         XCTAssert(["docType": "new-with-same-ID"] == db.document(withID: docID)!.toDictionary())
     }
     
+    func testDocumentReplicationEventForConflictedDocs() throws {
+        var resolver: TestConflictResolver!
+        
+        // when resolution is skipped: here wrong doc-id throws an exception & skips it
```
pasin: The comment is not updated.

PR #2499, opened 2019-06-12
CBL-72: unit: throws an exception in save-handler

jayahariv:
* when exception thrown in save handler, it should show a warning log. and fail the save.
* add warning message.

#2430

### DIFF
--- a/Swift/Database.swift
+++ b/Swift/Database.swift
@@ -127,7 +127,7 @@ public final class Database {
     ///   - document: The document.
     ///   - conflictHandler: The conflict handler closure which can be used to resolve it.
     /// - Returns: True if successful. False on failure
-    /// - Throws: An error on a failure.
+    /// - Throws: An error, if an exception happens.
     @discardableResult public func saveDocument(
         _ document: MutableDocument, conflictHandler: @escaping (MutableDocument, Document?) -> Bool
         ) throws -> Bool {
--- a/Swift/Database.swift
+++ b/Swift/Database.swift
@@ -118,16 +118,19 @@ public final class Database {
     }
     
     
-    /// Saves a document to the database. When write operations are executed concurrently,
-    /// and if conflicts occur, conflict handler will be called. Use the handler to directly
-    /// edit the document. Returning true, will save the document. Returning false, will cancel
-    /// the save operation.
-    ///
+    /// Saves a document to the database. When write operations are executed concurrently and if
+    /// conflicts occur, the conflict handler will be called. Use the conflict handler to directly
+    /// edit the document to resolve the conflict. When the conflict handler returns 'true', the
+    /// save method will save the edited document as the resolved document. If the conflict handler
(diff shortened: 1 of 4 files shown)


Swift/Database.swift
```diff
@@ -127,7 +127,7 @@ public final class Database {
     ///   - document: The document.
     ///   - conflictHandler: The conflict handler closure which can be used to resolve it.
     /// - Returns: True if successful. False on failure
-    /// - Throws: An error on a failure.
+    /// - Throws: An error, if an exception happens.
```
jayahariv:
```
    /// - Returns: True if successful. False on failure
    /// - Throws: An error on a failure.
```

From the comments, feeling like, when there is a failure of save, it will throw an exception. And when there is any exception other than conflict, it will throw an exception. But if it is a conflict, and it will return false without an error.

Swift/Tests/DatabaseTest.swift
```diff
@@ -412,6 +412,27 @@ class DatabaseTest: CBLTestCase {
...
+        try db.saveDocument(doc)
+        
+        let doc1a = db.document(withID: doc.id)!.toMutable()
+        let doc1b = db.document(withID: doc.id)!.toMutable()
+        doc1a.setString("Scott", forKey: "firstName")
+        try db.saveDocument(doc1a)
+        
+        // conflicting save
+        doc1b.setString("Lion", forKey: "middleName")
+        XCTAssertFalse(try db.saveDocument(doc1b) { (cur, old) -> Bool in
```
pasin: What does happen to the exception thrown? Crash?

jayahariv: It won't crash, unless they force unwrap try like this.`try! db.saveDocument(doc1a) { (doc, old) -> Bool in`

jayahariv: We already have a test which gets back the `CBLErrorNotFound` error. 
https://github.com/couchbase/couchbase-lite-ios/blob/0f2cd0edfecffcec1212fa0aa108a9019fced8eb/Swift/Tests/DatabaseTest.swift#L452

Swift/Database.swift
pasin: Just saw your comment. The comment seems wrong. The comment should match or similar to saveDocument with concurrencyControl. Maybe something like below.

```
Saves a document to the database. When write operations are executed concurrently and if conflicts occur, 
the conflict handler will be called. Use the conflict handler to directly edit the document to resolve the conflict. When the conflict handler returns 'true', the save method will save the edited document as the resolved document. If the conflict handler returns 'false', the save operation will be canceled with 'false' value returned as the conflict wasn't resolved. 
...
/// - Returns: True if successful. False if there is a conflict, but the conflict wasn't resolved as the conflict handler returns 'false' value.
/// - Throws: An error on a failure.
```

PR #2503, opened 2019-06-13
Make conflict resolution concurrent and defer stopped until done 

pasin:
1. Refactored the replicator to use an enum state to control start, stop, and suspend.
2. Made conflict resolution concurrent and defer stopped until done. 
3. Moved the logic to remove from the active replicator list to the stopped method so that the logic could go along with updating/posting stopped status.


pasin:
I will close and reopen the PR to avoid confusion.

PR #2507, opened 2019-06-14
Remove suspending cancel state

pasin:
Removed suspending cancel state and change the logic to use _cancelSuspending variable to signal resuming after suspended.

### DIFF
--- a/Objective-C/CBLReplicator.mm
+++ b/Objective-C/CBLReplicator.mm
@@ -88,18 +88,18 @@ typedef enum {
     dispatch_queue_t _conflictQueue;
     C4Replicator* _repl;
     NSString* _desc;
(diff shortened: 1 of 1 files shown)


Objective-C/CBLReplicator.mm
```diff
@@ -100,6 +99,7 @@ @implementation CBLReplicator
     CBLChangeNotifier<CBLReplicatorChange*>* _changeNotifier;
     CBLChangeNotifier<CBLDocumentReplication*>* _docReplicationNotifier;
     BOOL _resetCheckpoint;
+    BOOL _cancelSuspending;
```
jayahariv: may be add a single line of comment, describing when this will be set and reset. or purpose of this.

PR #2515, opened 2019-06-19
fix: only one warning about skip resolution

jayahariv:
* only one log per conflict skipping.

### DIFF
--- a/Objective-C/CBLReplicator.mm
+++ b/Objective-C/CBLReplicator.mm
@@ -596,7 +596,7 @@ static void onDocsEnded(C4Replicator* repl,
     BOOL success = [_config.database resolveConflictInDocument: doc.id
                                           withConflictResolver: _config.conflictResolver
                                                          error: &error];
-    if (success || !error) {
+    if (success && !error) {
         [doc resetError];
     } else {
         CBLWarn(Sync, @"%@: Conflict resolution of '%@' failed: %@", self, doc.id, error);
--- a/Objective-C/CBLReplicator.mm
+++ b/Objective-C/CBLReplicator.mm
@@ -593,14 +593,16 @@ static void onDocsEnded(C4Replicator* repl,
 - (void) _resolveConflict: (CBLReplicatedDocument*)doc {
     CBLLogInfo(Sync, @"%@: Resolve conflicting version of '%@'", self, doc.id);
     NSError* error;
-    if ([_config.database resolveConflictInDocument: doc.id
-                               withConflictResolver: _config.conflictResolver
-                                              error: &error])
+    BOOL success = [_config.database resolveConflictInDocument: doc.id
+                                          withConflictResolver: _config.conflictResolver
+                                                         error: &error];
+    if (success || !error) {
         [doc resetError];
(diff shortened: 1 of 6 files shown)


Objective-C/CBLReplicator.mm
```diff
@@ -597,7 +597,7 @@ - (void) _resolveConflict: (CBLReplicatedDocument*)doc {
                                withConflictResolver: _config.conflictResolver
                                               error: &error])
         [doc resetError];
-    else {
+    else if (error) {
```
pasin: This means that if the error is nil, we have never reseted the error. I think we probably don't need the `resetError` method if we always update the error at the end of the `_resolveConflict:` method.

jayahariv: then i feel like we should be returning success, else seems like we will end up with 


Screen Shot 2019-06-19 at 10 14 43 AM

jayahariv: updated the change: https://github.com/couchbase/couchbase-lite-ios/pull/2515/commits/154b2e7240103a90d68086556f46fc3c282a705b

Objective-C/CBLReplicator.mm
```diff
@@ -593,14 +593,16 @@ - (void) resolveConflict: (CBLReplicatedDocument*)doc {
 - (void) _resolveConflict: (CBLReplicatedDocument*)doc {
     CBLLogInfo(Sync, @"%@: Resolve conflicting version of '%@'", self, doc.id);
     NSError* error;
-    if ([_config.database resolveConflictInDocument: doc.id
-                               withConflictResolver: _config.conflictResolver
-                                              error: &error])
+    BOOL success = [_config.database resolveConflictInDocument: doc.id
+                                          withConflictResolver: _config.conflictResolver
+                                                         error: &error];
+    if (success && !error) {
```
pasin: Is there a case of success and having an error?

pasin: Can we do something this?

```
- (void) _resolveConflict: (CBLReplicatedDocument*)doc {
    CBLLogInfo(Sync, @"%@: Resolve conflicting version of '%@'", self, doc.id);
    NSError* error = nil;
    BOOL success = [_config.database resolveConflictInDocument: doc.id
                                          withConflictResolver: _config.conflictResolver
                                                         error: &error];
    if (!success)
        CBLWarn(Sync, @"%@: Conflict resolution of '%@' failed: %@", self, doc.id, error);
    
    [doc updateError: error];
    [self logErrorOnDocument: doc pushing: NO];
    [self postDocumentReplications: @[doc] pushing: NO];
}
```

jayahariv: i think, if we return true from the database, this is enough.

jayahariv: > Is there a case of success and having an error?
there is none i can think of. i think, only checking success is fine

pasin: Then in updateError, if error is nil, you can internally resetError so don't need two methods, resetError and updateError.

PR #2517, opened 2019-06-20
fix: change resolution to resolve

jayahariv:
* updated class to adopt the naming of ConflictResolve.default
* removed the ConflictResolution class, which we copied to ConflictResolver class.
* swift we named the protocol with a suffix, and objc kept same name for both.
* updated tests with changes


Swift/ConflictResolver.swift
```diff
@@ -20,9 +20,38 @@
...
-public protocol ConflictResolver {
+public protocol ConflictResolverProtocol {
     
     /// The callback resolve method, if conflict occurs.
     func resolve(conflict: Conflict) -> Document?
     
 }
+
+
+/// Conflict Resolution class
```
pasin: Maybe `ConflictResolver provides access to the default conflict resolver used by the replicator`.

Swift/ConflictResolver.swift
```diff
@@ -20,9 +20,38 @@
...
+public class ConflictResolver {
```
pasin: Let's make the class final.

Swift/ConflictResolver.swift
```diff
@@ -20,9 +20,38 @@
...
+    
+    /// The default conflict resolver used by the replicator.
+    public static var `default`: ConflictResolver {
+        return DefaultResolver()
```
pasin: Can we do public static let `default` so that there will be only one single instance of the DefaultResolver?

Objective-C/CBLConflictResolver.m
```diff
@@ -17,24 +17,23 @@
...
+@interface CBLDefaultConflictResolver : NSObject <CBLConflictResolver>
 
 @end
 
-@implementation CBLConflictResolution
+@implementation CBLConflictResolver
 
 + (id) default {
-    return [[CBLDefaultConflictResolution alloc] init];
+    return [[CBLDefaultConflictResolver alloc] init];
```
pasin: Can we make the default singleton?

PR #2523, opened 2019-06-26
CBL-98: Create a separate modulemap file for accessing Objective-C API

pasin:
Instead of using a submodule to the main module, create a separate module map file for accessing Objective-C API. By doing it this way, we don’t need to run the post build script to remove the submodule and private headers.

Reference: CBL-98


pasin:
Closed as not working.

PR #2525, opened 2019-07-03
CBL-138: fix: multiple retry logic

jayahariv:
* cancel the previous retry before starting a new one.
* only allow to retry from specific two states, suspended or offline.
* replicator can start from stopped state but that shouldn't come through a retry but start itself.

### DIFF
--- a/Objective-C/CBLReplicator.mm
+++ b/Objective-C/CBLReplicator.mm
@@ -200,17 +200,19 @@ typedef enum {
 }
 
 - (void) retry {
-    dispatch_async(_dispatchQueue, ^{
-        CBL_LOCK(self) {
-            CBLLogInfo(Sync, @"%@: Retrying...", self);
-            if (_repl || _state <= kCBLStateStopping) {
-                CBLLogInfo(Sync, @"%@: Ignore retrying (state = %d, status = %d)",
-                           self, _state, _rawStatus.level);
-                return;
+    if (_state == kCBLStateOffline || _state == kCBLStateSuspended) {
+        dispatch_async(_dispatchQueue, ^{
+            CBL_LOCK(self) {
+                CBLLogInfo(Sync, @"%@: Retrying...", self);
+                if (_repl || _state <= kCBLStateStopping) {
+                    CBLLogInfo(Sync, @"%@: Ignore retrying (state = %d, status = %d)",
+                               self, _state, _rawStatus.level);
+                    return;
+                }
+                [self _start];
             }
-            [self _start];
-        }
-    });
+        });
+    }
 }
 
 - (void) _start {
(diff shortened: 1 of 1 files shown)


Objective-C/CBLReplicator.mm
```diff
@@ -165,17 +165,53 @@ - (void) start {
...
+                   afterDelay: delay];
+    });
+}
+
+- (void) resetRetryCount {
+    CBLLogVerbose(Sync, @"%@: Resetting the retry count", self);
+    _retryCount = 0;
+}
+
+- (void) cancelPrviousRetry {
```
pasin: misspelled the method name.

Objective-C/CBLReplicator.mm
```diff
@@ -165,17 +165,53 @@ - (void) start {
...
+            [NSObject cancelPreviousPerformRequestsWithTarget: strongSelf
+                                                     selector: @selector(retry)
+                                                       object: nil];
         }
-        [self _start];
+    });
+}
+
+- (void) retry {
+    if (_state == kCBLStateOffline || _state == kCBLStateSuspended) {
```
pasin: The _state is accessed without lock.

Objective-C/CBLReplicator.mm
```diff
@@ -165,17 +165,53 @@ - (void) start {
...
+        dispatch_async(_dispatchQueue, ^{
```
pasin: Why do we need to call in the dispatch queue?

Objective-C/CBLReplicator.mm
```diff
@@ -505,14 +541,7 @@ - (bool) handleError: (C4Error)c4err {
...
                    self, error.localizedDescription, delay);
-        dispatch_after(dispatch_time(DISPATCH_TIME_NOW, (int64_t)(delay * NSEC_PER_SEC)),
-                       _dispatchQueue, ^{
-                           CBL_LOCK(self) {
-                               // Retry if still in offline state:
-                               if (_state == kCBLStateOffline)
-                                   [self retry: NO];
-                           }
-                       });
+        [self scheduleRetry: delay];
```
pasin: Also we don't want the retry to be executed if the replicator is suspended.

Objective-C/CBLReplicator.mm
pasin: What do you think about using dispatch_source_set_timer instead? Will it be cleaner?

Objective-C/CBLReplicator.mm
jayahariv: else this will be called from main queue

Objective-C/CBLReplicator.mm
pasin: Another way to implement this is to cancel the previous retry before suspending so we don't need to check here.

Objective-C/CBLReplicator.mm
jayahariv: agree. `dispatch_source_set_timer` is a better option here. We can leave the run loop selection part to GCD side. And we can wrap that to a class, which can be reuse in document expiry logic.

jayahariv:
Closing this PR and create a separate PR.

PR #2532, opened 2019-08-16
feat: add dispatch source set timer

jayahariv:
* create a wrapper for dispatch source set timer, for all cbl timer tasks.
* for a test trial, check by replacing the document expiry code block.

Not sure, why those extra project file changes showing up. is it XCode automatically removed them?

### DIFF
--- a/Objective-C/Internal/CBLTimer.h
+++ b/Objective-C/Internal/CBLTimer.h
@@ -24,10 +24,24 @@ NS_ASSUME_NONNULL_BEGIN
 
 @interface CBLTimer : NSObject
 
+/**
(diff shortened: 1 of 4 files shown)


Objective-C/Internal/CBLTimer.h
```diff
@@ -0,0 +1,35 @@
...
+//
+
+
+#import <Foundation/Foundation.h>
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface CBLTimer : NSObject
+
++ (dispatch_source_t) scheduleIn: (nullable dispatch_queue_t)queue
```
pasin: Can you add the comment about using global_queue when queue is nil?

PR #2541, opened 2019-09-09
CBL-342: fix: race condition

jayahariv:
* when replicator starts before its handle being created the nullptr was returned.
* fix by returning replicator handle, and then start the replicator
* agree with Blake on this. https://github.com/couchbase/couchbase-lite-java/pull/139#discussion_r322407475

### DIFF
--- a/Objective-C/CBLReplicator.mm
+++ b/Objective-C/CBLReplicator.mm
@@ -288,8 +288,8 @@ typedef enum {
         if (!_repl) {
             NSError *error = nil;
             convertError(err, &error);
-            CBLLogInfo(Sync, @"%@: Replicator cannot be created: %@", self,
-                       error.localizedDescription);
+            CBLWarnError(Sync, @"%@: Replicator cannot be created: %@",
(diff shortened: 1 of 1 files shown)


Objective-C/CBLReplicator.mm
```diff
@@ -283,6 +284,15 @@ - (void) _start {
     C4Error err;
     CBL_LOCK(_config.database) {
         _repl = c4repl_new(_config.database.c4db, addr, dbName, otherDB.c4db, params, &err);
+        
+        if (!_repl) {
+            NSError *error = nil;
+            convertError(err, &error);
+            CBLLogInfo(Sync, @"%@: Replicator cannot be created: %@", self,
```
pasin: Should be error not info.

pasin:
Need to fix here in EE as well:
https://github.com/couchbaselabs/couchbase-lite-ios-ee/blob/master/Sources/Objective-C/Replicator/CBLMessageEndpointListener.mm#L96

PR #2544, opened 2019-09-11
CBL-333: feat: xcframework script

jayahariv:
* xcframework script to generate the release package script. 
* Reference ticket: CBL-333
* include the flag `--xcframework` with the `generate_release_zip.sh`

### DIFF
--- a/CouchbaseLite.xcodeproj/project.pbxproj
+++ b/CouchbaseLite.xcodeproj/project.pbxproj
@@ -3078,7 +3078,7 @@
 				27F961961ED777EA0060F804 /* libSupport.a */,
 				9398D94A1E0347B600464432 /* LiteCoreCppTests */,
 				9398D94C1E0347B600464432 /* C4Tests */,
-				9398D94E1E0347B600464432 /* libLiteCoreJNI.dylib */,
+				9398D94E1E0347B600464432 /* LiteCoreJNI.dylib */,
 				27DF7D841F424C840022F3DF /* libSQLite.a */,
 				9398D9501E0347B600464432 /* libTokenizer.a */,
 				2749B97F1EB25A660068DBF9 /* libcivetweb.a */,
@@ -4416,10 +4416,10 @@
 			remoteRef = 9398D94B1E0347B600464432 /* PBXContainerItemProxy */;
 			sourceTree = BUILT_PRODUCTS_DIR;
 		};
-		9398D94E1E0347B600464432 /* libLiteCoreJNI.dylib */ = {
+		9398D94E1E0347B600464432 /* LiteCoreJNI.dylib */ = {
 			isa = PBXReferenceProxy;
 			fileType = "compiled.mach-o.dylib";
-			path = libLiteCoreJNI.dylib;
+			path = LiteCoreJNI.dylib;
 			remoteRef = 9398D94D1E0347B600464432 /* PBXContainerItemProxy */;
 			sourceTree = BUILT_PRODUCTS_DIR;
 		};
@@ -5560,6 +5560,7 @@
 			isa = XCBuildConfiguration;
 			baseConfigurationReference = 27BE3B451E4D63AF0012B74A /* CBL Swift.xcconfig */;
 			buildSettings = {
+				BUILD_LIBRARY_FOR_DISTRIBUTION = YES;
 			};
 			name = Release;
 		};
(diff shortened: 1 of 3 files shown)


CouchbaseLite.xcodeproj/project.pbxproj
```diff
@@ -5565,6 +5567,7 @@
 			isa = XCBuildConfiguration;
 			baseConfigurationReference = 27BE3B451E4D63AF0012B74A /* CBL Swift.xcconfig */;
 			buildSettings = {
+				BUILD_LIBRARY_FOR_DISTRIBUTION = YES;
```
pasin: Do you know what is this config? Should it go to the xcconfig file?

Scripts/build_xcframework.sh
```diff
@@ -0,0 +1,112 @@
...
+  xcodebuild archive -scheme "${SCHEME}" -configuration "${CONFIGURATION}" -destination "generic/platform=${DESTINATION}" ${BUILD_VERSION} ${BUILD_NUMBER} -archivePath ${ARCHIVE_PATH} "ONLY_ACTIVE_ARCH=NO" "BITCODE_GENERATION_MODE=bitcode" "CODE_SIGNING_REQUIRED=NO" "CODE_SIGN_IDENTITY=" "clean"  ${QUIET} "SKIP_INSTALL=NO"
+  echo "Finished archiving ${DESTINATION}."
+done
+
+FRAMEWORK_LOCATION=${BIN_NAME}.xcarchive/Products/Library/Frameworks/${BIN_NAME}.framework
+
+# create xcframework
+echo "Make Objective-C framework zip file ..."
+mkdir -p "${OUTPUT_DIR}/${SCHEME}"
+xcodebuild -create-xcframework -output "${OUTPUT_DIR}/${SCHEME}/${BIN_NAME}.xcframework" -framework ${BUILD_DIR}/iOS/${FRAMEWORK_LOCATION} -framework ${BUILD_DIR}/macOS/${FRAMEWORK_LOCATION} -framework ${BUILD_DIR}/iOS_Simulator/${FRAMEWORK_LOCATION}
```
pasin: Can the the three -framework params be constructed in the for-loop above?

Scripts/build_xcframework.sh
```diff
@@ -0,0 +1,112 @@
...
+
+# remove all related files
+for DESTINATION in "${DESTINATIONS[@]}"
+do
+  FOLDER_NAME=$(echo ${DESTINATION} | sed 's/ /_/g')
```
pasin: This hasn't bee used.

CouchbaseLite.xcodeproj/project.pbxproj
jayahariv: This is necessary when we make it xcframework

Scripts/build_xcframework.sh
jayahariv: Yea true. It can be removed.

Scripts/build_xcframework.sh
jayahariv: After creating the archive from above for loop, we are supplying them as args to -create-xcframework command.

CouchbaseLite.xcodeproj/project.pbxproj
pasin: At least, it needs to be in xcconfig file, not in the xcode project. The following should be find out (maybe look to see how other libraries are doing).
1. Should BUILD_LIBRARY_FOR_DISTRIBUTION be enabled on both Swift and Objective-C target?
2. Should BUILD_LIBRARY_FOR_DISTRIBUTION be enabled on both debug and release build.
3. Should we just enable it at the xcodebuild command?

Scripts/build_xcframework.sh
jayahariv: removed, loop.

Scripts/build_xcframework.sh
jayahariv: ha! you found it. this was a hack, just to get rid of too many logic to save the paths etc. In the create framework statement, we are hard coding the paths, since we already know where the paths going to be in. 

Will correct this. Create the `-framework PATH` from above for-loop and then passing to the `create-framework` will be more scalable/less code.

CouchbaseLite.xcodeproj/project.pbxproj
jayahariv: as per the understanding this will create the `<arch-type>.swiftinterface` file for swift target. which will be used to compatible with future swift compiler versions.

jayahariv: I am getting an error called "No 'swiftinterface' files found within <path to module map>" when using as a command line argument in archive.

PR #2547, opened 2019-09-20
CBL-373: fix: swift coverage

jayahariv:
* array object set and add method verification tests added
* conflict resolver default
* CBL-373

### DIFF
--- a/CouchbaseLite.xcodeproj/project.pbxproj
+++ b/CouchbaseLite.xcodeproj/project.pbxproj
@@ -151,6 +151,7 @@
 		27F9619A1ED8D9440060F804 /* CBLReachability.h in Headers */ = {isa = PBXBuildFile; fileRef = 27F961971ED8D9440060F804 /* CBLReachability.h */; };
 		27F9619B1ED8D9440060F804 /* CBLReachability.m in Sources */ = {isa = PBXBuildFile; fileRef = 27F961981ED8D9440060F804 /* CBLReachability.m */; };
 		27F9619C1ED8D9440060F804 /* CBLReachability.m in Sources */ = {isa = PBXBuildFile; fileRef = 27F961981ED8D9440060F804 /* CBLReachability.m */; };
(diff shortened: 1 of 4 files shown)


Swift/Tests/ArrayTest.swift
```diff
@@ -84,5 +112,120 @@ class ArrayTest: CBLTestCase {
...
+        array.addDouble(data[3] as! Double)
+        array.addFloat(Float(data[3] as! Double))
+        
+        array.addDate(data[4] as? Date)
+        let dict = MutableDictionaryObject(data: data[5] as! [String: String])
+        array.addDictionary(dict)
+        let subarray = MutableArrayObject(data: data[6] as! [String])
+        array.addArray(subarray)
+        array.addBlob(data[7] as? Blob)
+        array.addValue(NSNull())
```
pasin: Wondering why not adding NSNull() to the arrayOfAllTypes() as well?

CouchbaseLite.xcodeproj/project.pbxproj
```diff
@@ -5246,6 +5247,7 @@
 				9343F192207D636300F19A89 /* DatabaseTest.swift in Sources */,
 				9343F193207D636300F19A89 /* CBLTestHelper.m in Sources */,
 				1AA91DC722B0356000BF0BDE /* CustomLogger.swift in Sources */,
+				69845B0723354D0A00CC16BB /* DateTimeQueryFunctionTest.swift in Sources */,
```
pasin: Is this a new test?

jayahariv: No, it was an old one, but not added to EE test target.

Swift/Tests/ArrayTest.swift
jayahariv: when i added, felt like not worth adding, parsing and comparing. But seems like, you are right, adding it to the array is better for consistency, else this will stay as separate and may miss in case a new test is added.

pasin: I think it's better to add it as it should be consistent when comparing between 2 things.

PR #2554, opened 2019-10-10
CBL-443: feat: common error message 

jayahariv:
```
kCBLErrorMessageCreateDBDirectoryFailed = @"Unable to create database directory."
kCBLErrorMessageDocumentNotFound = @"The document doesn't exist in the database."
kCBLErrorMessageResolvedDocFailedLiteCore = @"LiteCore failed resolving conflict."
kCBLErrorMessageNoDocumentRevision = @"No revision data on the document!"
kCBLErrorMessageFragmentPathNotExist = @"Specified fragment path does not exist in object; cannot set value."
kCBLErrorMessageInvalidValueToBeDeserialized = @"Non-string or null key in data to be deserialized."
kCBLErrorMessageCannotSetLogLevel = @"Cannot set logging level without a configuration."
kCBLErrorMessageQueryParamNotAllowedContainCollections = @"Query parameters are not allowed to contain collections."
kCBLErrorMessageMissASforJoin = @"Missing AS clause for JOIN."
kCBLErrorMessageMissONforJoin = @"Missing ON statement for JOIN."
kCBLErrorMessageExpressionsMustBeIExpressionOrString = @"Expressions must either be %1$@ or String."
kCBLErrorMessageInvalidExpressionValueBetween = @"Invalid expression value for expression of Between(%1$@)."
kCBLErrorMessageResultSetAlreadyEnumerated = @"This result set has already been enumerated. Please re-execute the original query."
kCBLErrorMessageExpressionsMustContainOnePlusElement = @"%1$@ expressions must contain at least one element."
kCBLErrorMessageDuplicateSelectResultName = @"Duplicate select result named %1$@."
kCBLErrorMessageNoAliasInJoin = @"The default database must have an alias in order to use a JOIN statement (Make sure your data source uses the As() function)."
kCBLErrorMessageInvalidQueryDBNull = @"Invalid query: The database is null."
kCBLErrorMessageInvalidQueryMissingSelectOrFrom = @"Invalid query: missing Select or From."
```

These are the un-used error codes in iOS.


Objective-C/CBLDatabase.mm
```diff
@@ -1054,8 +1047,12 @@ - (BOOL) saveResolvedDocument: (CBLDocument*)resolvedDoc
                                                 userInfo: @{NSLocalizedDescriptionKey: ex.description}];
                     return false;
                 }
-                if (!mergedBody)
+                if (!mergedBody) {
+                    *outError = [NSError errorWithDomain: CBLErrorDomain
+                                                    code: CBLErrorUnexpectedError
+                                                userInfo: @{NSLocalizedDescriptionKey: kCBLErrorMessageResolvedDocContainsNull}];
```
jayahariv: added this extra!

Objective-C/CBLDocument.mm
```diff
@@ -177,11 +178,14 @@ - (FLSliceResult) encode: (NSError**)outError {
 
 - (BOOL) selectConflictingRevision {
     CBL_LOCK(self) {
+        if (!_c4Doc) {
+            [NSException raise: NSInternalInconsistencyException
+                        format: @"%@", kCBLErrorMessageNoDocumentRevision];
+        }
```
jayahariv: added this extra

Objective-C/CBLDatabase.mm
bmeike: you da man!

PR #2555, opened 2019-10-14
CBL-437: fix: deprecated setlog level method

jayahariv:
* before deprecated means, no op. now we set console.log.domain and console.log.level from it.

### DIFF
--- a/Objective-C/CBLDatabase.mm
+++ b/Objective-C/CBLDatabase.mm
@@ -462,6 +462,9 @@ static void dbObserverCallback(C4DatabaseObserver* obs, void* context) {
 + (void) setLogLevel: (CBLLogLevel)level domain: (CBLLogDomain)domain {
     CBLWarn(Database, @"This method has been deprecated. "
             "Please use CBLDatabase.log.console instead of -setLogLevel:domain:.");
+    
+    CBLDatabase.log.console.domains =  domain;
+    CBLDatabase.log.console.level = level;
 }
 
 + (CBLLog*) log {
--- a/Objective-C/CBLDatabase.mm
+++ b/Objective-C/CBLDatabase.mm
@@ -463,7 +463,7 @@ static void dbObserverCallback(C4DatabaseObserver* obs, void* context) {
     CBLWarn(Database, @"This method has been deprecated. "
             "Please use CBLDatabase.log.console instead of -setLogLevel:domain:.");
     
-    CBLDatabase.log.console.domains =  domain;
+    CBLDatabase.log.console.domains = kCBLLogDomainAll;
     CBLDatabase.log.console.level = level;
 }
 
--- a/Swift/Database.swift
+++ b/Swift/Database.swift
@@ -371,7 +371,7 @@ public final class Database {
         Log.log(domain: .database, level: .warning, message:
             "This method has been deprecated. Please use Database.log.console instead of setLogLevel(_, domain:)")
         
-        Database.log.console.domains = LogDomains(rawValue: Int(domain.rawValue))
+        Database.log.console.domains = LogDomains(rawValue: LogDomain.all.hashValue)
         Database.log.console.level = level
     }
     
--- a/Swift/Database.swift
+++ b/Swift/Database.swift
@@ -370,6 +370,7 @@ public final class Database {
     public class func setLogLevel(_ level: LogLevel, domain: LogDomain) {
         Log.log(domain: .database, level: .warning, message:
             "This method has been deprecated. Please use Database.log.console instead of setLogLevel(_, domain:)")
+        
         Database.log.console.domains = LogDomains(rawValue: Int(domain.rawValue))
         Database.log.console.level = level
     }
--- a/Swift/Database.swift
+++ b/Swift/Database.swift
@@ -370,6 +370,8 @@ public final class Database {
     public class func setLogLevel(_ level: LogLevel, domain: LogDomain) {
         Log.log(domain: .database, level: .warning, message:
             "This method has been deprecated. Please use Database.log.console instead of setLogLevel(_, domain:)")
+        Database.log.console.domains = LogDomains(rawValue: Int(domain.rawValue))
+        Database.log.console.level = level
     }
     
     /// Log object used for configuring console, file, and custom logger.


Swift/Database.swift
```diff
@@ -370,6 +370,8 @@ public final class Database {
     public class func setLogLevel(_ level: LogLevel, domain: LogDomain) {
         Log.log(domain: .database, level: .warning, message:
             "This method has been deprecated. Please use Database.log.console instead of setLogLevel(_, domain:)")
+        Database.log.console.domains = LogDomains(rawValue: Int(domain.rawValue))
```
jayahariv:
```suggestion
        
        Database.log.console.domains = LogDomains(rawValue: Int(domain.rawValue))
```

Objective-C/CBLDatabase.mm
```diff
@@ -462,6 +462,9 @@ + (BOOL) copyFromPath: (NSString*)path
 + (void) setLogLevel: (CBLLogLevel)level domain: (CBLLogDomain)domain {
     CBLWarn(Database, @"This method has been deprecated. "
             "Please use CBLDatabase.log.console instead of -setLogLevel:domain:.");
+    
+    CBLDatabase.log.console.domains =  domain;
```
pasin: extra space b/w = and domain.

Objective-C/CBLDatabase.mm
```diff
@@ -462,6 +462,9 @@ + (BOOL) copyFromPath: (NSString*)path
...
+    CBLDatabase.log.console.level = level;
```
pasin: Normally this method will be used as follows:

[CBLDatabase setLogLevel: verbose domain: database];
[CBLDatabase setLogLevel: verbose domain: query];
[CBLDatabase setLogLevel: verbose domain: replicator];

With this change, only replicator domain will be enabled. I feel like changing as one of the followings is better:

1. Ignore domain which means that always set to all domains.

2. Add domain to the set of domains as most of the time, users will have the same level for each domain.

Please discuss with the others.

bmeike: @Pasin, I think that the discussion was that normally, this method will not be used at all: it is deprecated.  I definitely thought that, even given the problems you mentioned, this was the implementation we agreed (warts and all).

pasin: I understand that the reason that we want to add the implementation to this method because some users / customers are still using the method and was confusing that there were no log messages printed out so to avoid the confusion, we are trying to add the implementation back.

bmeike: Does the iOS implementation have a way of saying "all domains"?  The Java impl does.  If there is no way to do that in iOS, perhaps this is less than ideal.

pasin: iOS can do all domains. I don't think none of the solution is perfect but at least if we set it to all domains, users will still have logs. We have one CBSE ticket that users couldn't see log messages and it turned out they are using the old setLogLevel: method.

pasin: Yes, using both methods could be weird.

bmeike: Maybe we should chat about this.  The corresponding Java function takes a single domain (not a set of domains) as an argument.  In order to fix this, someone introduced a new domain called "ALL" which, one presumes, should behave like the set of all domains.
The function in Java is exactly like this one.  Because of that weird domain, though, the client can set logging to be any one domain at some level, or all domains at some level.
That seems entirely acceptable, to me.

borrrden:
I am not on the iOS code review list, is there something that specifically needs me here?

Objective-C/CBLDatabase.mm
jayahariv:
```suggestion
    CBLDatabase.log.console.domains = domain;
```

Swift/Database.swift
```diff
@@ -370,6 +370,9 @@ public final class Database {
     public class func setLogLevel(_ level: LogLevel, domain: LogDomain) {
         Log.log(domain: .database, level: .warning, message:
             "This method has been deprecated. Please use Database.log.console instead of setLogLevel(_, domain:)")
+        
+        Database.log.console.domains = LogDomains(rawValue: Int(domain.rawValue))
```
jayahariv:
```suggestion
        Database.log.console.domains = LogDomains(rawValue: LogDomain.all.hashValue)
```

Objective-C/CBLDatabase.mm
borrrden: .NET takes an enum, which means you can flag together any combination you want.  So it is expected that this call is only used once for all the domains that are desired.

jayahariv: this means, if the user set verbose to database and info to replicator, end up in setting, all domains to info. which is also confusing, since user expects verbose from database. Last set level will only be applicable. 

[CBLDatabase setLogLevel: verbose domain: database];
[CBLDatabase setLogLevel: info domain: replicator];

borrrden: There is no way that this API can behave exactly the same as before, so we cannot evaluate it against that standard.

pasin: > this means, if the user set verbose to database and info to replicator, end up in setting, all domains to info. which is also confusing, since user expects verbose from database. Last set level will only be applicable.

I think this option has least confusing; it's better than not having database log at all. Also I believe most of the users will not have different log level for each domains; most likely they will have the same for all domains that they specified.

jayahariv: FYI @bmeike @borrrden @Sandychuang8 @pasin 

iOS will follow the below logic, 
```
CBLDatabase.log.console.domains = kCBLLogDomainAll;
CBLDatabase.log.console.level = level;
```

Sandychuang8: @borrrden should .Net follow iOS?

bmeike: Java is doing something different.  I can't track the discussion that got me there, but there was discussion.  Lots of it.
Java does:
- affects only the console
- sets the console logger to log exactly the passed domain.  That domain can be `ALL`, or any one specific domain.
- sets the console logger to exactly the passed level.
No affect on file logging or custom logging.  No way to choose multiple but not all domains.
It is a *deprecated* method and will interact with the new logging facility in really odd ways, no matter what we do.  I vote we move on...

PR #2558, opened 2019-10-25
CBL-488: Defer notification for C4Idle status

jayahariv:
- in case of idle state, we defer the notification till we finish resolving all conflicts.

### DIFF
--- a/Objective-C/CBLReplicator.mm
+++ b/Objective-C/CBLReplicator.mm
@@ -99,7 +99,7 @@ typedef enum {
     BOOL _resetCheckpoint;          // Reset the replicator checkpoint
     BOOL _cancelSuspending;         // Cancel the current suspending request
     unsigned _conflictCount;        // Current number of conflict resolving tasks
-    BOOL _deferStoppedNotification; // Defer the stopped until finishing all conflict resolving tasks
+    BOOL _deferReplicatorNotification; // Defer the replicator notificayion until finishing all conflict resolving tasks
     dispatch_source_t _retryTimer;
 }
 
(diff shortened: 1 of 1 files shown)


Objective-C/CBLReplicator.mm
```diff
@@ -612,10 +616,20 @@ - (void) resolveConflict: (CBLReplicatedDocument*)doc {
     dispatch_async(_conflictQueue, ^{
         [self _resolveConflict: doc];
         CBL_LOCK(self) {
-            if (--_conflictCount == 0 && _deferStoppedNotification) {
-                Assert(_rawStatus.level == kC4Stopped);
-                Assert(_state == kCBLStateStopping);
-                [self stopped];
+            if (--_conflictCount == 0) {
+                if (_deferReplicatorNotification) {
```
pasin: The original code looks cleaner.

jayahariv:
taking back! will submit again

Objective-C/CBLReplicator.mm
```diff
@@ -502,10 +501,15 @@ - (void) c4StatusChanged: (C4ReplicatorStatus)c4Status {
...
-                                  "to finish before notifying.", self, _conflictCount);
-                _deferStoppedNotification = YES;
+        }
+        
+        // handle replicator status callback
+        if (_conflictCount > 0 && (c4Status.level == kC4Stopped || c4Status.level == kC4Idle)) {
+            CBLLogInfo(Sync, @"%@: Status = %d, but waiting for conflict resolution (pending = %d) "
+            "to finish before notifying.", self, c4Status.level, _conflictCount);
+            
+            _deferReplicatorNotification = YES;
```
pasin: It's also possible that the IDLE status could be changed to Busy so the deferred IDLE might not valid anymore. For that case, we can reset _deferReplicatorNotification = NO.

pasin: Please also review other scenarios of the status change from IDLE to something else to see the behavior of deferring the idle notification.

pasin:
I did add some comments.

PR #2564, opened 2019-10-30
CBL-449: Pending Document ID ObjC

jayahariv:
* add implementation for getting the Pending Document ID in ObjC
* add one unit test

### DIFF
--- a/Objective-C/CBLReplicator.mm
+++ b/Objective-C/CBLReplicator.mm
@@ -456,34 +456,29 @@ static C4ReplicatorValidationFunction filter(CBLReplicationFilter filter, bool i
 }
 
 - (NSSet<NSString*>*) pendingDocumentIds: (NSError**)error {
-    // FIXME: Invalidate pending doc Id
-    if (!_pendingDocumentIds)
-        return _pendingDocumentIds;
-    
     if (_config.replicatorType > 1) {
-        *error = [NSError errorWithDomain: CBLErrorDomain
-                                     code: CBLErrorUnsupported
-                                 userInfo: @{NSLocalizedDescriptionKey: kCBLErrorMessagePullOnlyPendingDocIDs}];
+        if (error)
+            *error = [NSError errorWithDomain: CBLErrorDomain
+                                         code: CBLErrorUnsupported
+                                     userInfo: @{NSLocalizedDescriptionKey: kCBLErrorMessagePullOnlyPendingDocIDs}];
         return [NSSet set];
     }
     
     
-    if (!_repl) {
+    if (!_repl)
         return [NSSet set];
-    }
+
     
     C4Error err;
     C4SliceResult result = c4repl_getPendingDocIDs(_repl, &err);
+    
+    if (!result.size)
+        return [NSSet set];
+        
     FLValue val = FLValue_FromData(C4Slice(result), kFLTrusted);
-    FLArray arr = FLValue_AsArray(val);
-    uint32_t count = FLArray_Count(arr);
-    NSMutableSet* docIds = [[NSMutableSet alloc] initWithCapacity: count];
-    for (uint32_t i = 0; i < count; i++) {
-        FLValue item = FLArray_Get(arr, i);
-        FLString strId = FLValue_AsString(item);
-        [docIds addObject: slice2string(strId)];
-    }
-    _pendingDocumentIds = [NSSet setWithSet: docIds];
+    NSArray<NSString*>* list = FLValue_GetNSObject(val, nullptr);
+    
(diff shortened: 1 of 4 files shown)


Objective-C/CBLReplicator.mm
```diff
@@ -454,6 +455,33 @@ - (void) removeChangeListenerWithToken: (id<CBLListenerToken>)token {
...
+            *error = [NSError errorWithDomain: CBLErrorDomain
+                                         code: CBLErrorUnsupported
+                                     userInfo: @{NSLocalizedDescriptionKey: kCBLErrorMessagePullOnlyPendingDocIDs}];
+        return [NSSet set];
+    }
+    
+    
+    if (!_repl)
+        return [NSSet set];
+
```
jayahariv:
```suggestion
```

Objective-C/CBLReplicator.mm
```diff
@@ -454,6 +455,32 @@ - (void) removeChangeListenerWithToken: (id<CBLListenerToken>)token {
...
 
+- (NSSet<NSString*>*) pendingDocumentIds: (NSError**)error {
+    if (_config.replicatorType > 1) {
+        if (error)
+            *error = [NSError errorWithDomain: CBLErrorDomain
+                                         code: CBLErrorUnsupported
+                                     userInfo: @{NSLocalizedDescriptionKey: kCBLErrorMessagePullOnlyPendingDocIDs}];
+        return [NSSet set];
+    }
+
```
jayahariv:
```suggestion
```

Objective-C/Tests/ReplicatorTest+PendingDocIds.m
```diff
@@ -0,0 +1,113 @@
...
+    
+    doc = [self createDocument: @"4"];
+    [doc setString:@"Smokey" forKey:@"name"];
+    [self saveDocument: doc];
+    
+    doc = [self createDocument: @"5"];
+    [doc setString:@"Smokey" forKey:@"name"];
+    [self saveDocument: doc];
+    
+
```
jayahariv:
```suggestion
```

Objective-C/CBLReplicator.mm
```diff
@@ -454,6 +455,36 @@ - (void) removeChangeListenerWithToken: (id<CBLListenerToken>)token {
     }
 }
 
+- (NSSet<NSString*>*) pendingDocumentIds: (NSError**)error {
+    if (_config.replicatorType > 1) {
+        if (error)
```
bmeike: Is this right?  If there is already an error, replace it?

jayahariv: this is to make sure, the passed in argument to this function is valid, then only we will assign value to the pointed location.

bmeike: All good, then!

Objective-C/CBLReplicator.mm
```diff
@@ -454,6 +455,38 @@ - (void) removeChangeListenerWithToken: (id<CBLListenerToken>)token {
...
+            *error = [NSError errorWithDomain: CBLErrorDomain
+                                         code: CBLErrorUnsupported
+                                     userInfo: @{NSLocalizedDescriptionKey: kCBLErrorMessagePullOnlyPendingDocIDs}];
+        return [NSSet set];
+    }
+    
+    if (!_repl) {
+        CBLLogInfo(Sync, @"Trying to fetch pending documentIds without a c4replicator %@", _repl);
+        return [NSSet set];
```
pasin: I understand that we should be able to return the pending docs even when before starting the replicator and when the replicator is stopped. We probably need to change to be able to create the replicator on demand (lazy) and make sure to also free them in destructor.

Objective-C/CBLReplicator.mm
```diff
@@ -454,6 +455,38 @@ - (void) removeChangeListenerWithToken: (id<CBLListenerToken>)token {
...
+    
+    C4Error c4err = {};
+    C4SliceResult result = c4repl_getPendingDocIDs(_repl, &c4err);
+    if (c4err.code > 0) {
+        convertError(c4err, error);
+        CBLWarnError(Sync, @"Error while fetching pending documentIds: %d/%d", c4err.domain, c4err.code);
+        return [NSSet set];
+    }
+    
+    if (!result.size)
```
pasin: I feel easier to understand when doing result.size > 0.

jayahariv:
```suggestion
    if (result.size <= 0)
```

Objective-C/CBLReplicator.mm
jayahariv: will create a separate PR with replicator change.

PR #2566, opened 2019-10-30
CBL-517: Lazy initialize c4replicator instance

jayahariv:
* move everything which creates the c4replicator to a method. 
* if any error, returned so that the initial status can include the error.
* remove assertion to check `_repl` not exists before starting it.

### DIFF
--- a/Objective-C/CBLReplicator.mm
+++ b/Objective-C/CBLReplicator.mm
@@ -134,6 +134,79 @@ typedef enum {
     return _config;
 }
 
+- (void) dealloc {
+    c4repl_free(_repl);
+    [_reachability stop];
+}
+
+- (NSString*) description {
+    if (!_desc)
+        _desc = [NSString stringWithFormat: @"%@[%s%s%s %@]",
+                 self.class,
+                 (isPull(_config.replicatorType) ? "<" : ""),
+                 (_config.continuous ? "*" : "-"),
+                 (isPush(_config.replicatorType)  ? ">" : ""),
+                 _config.target];
+    return _desc;
+}
+
+- (void) clearRepl {
+    c4repl_free(_repl);
+    _repl = nullptr;
+}
+
+- (void) start {
+    CBL_LOCK(self) {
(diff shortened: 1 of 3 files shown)


Objective-C/CBLReplicator.mm
```diff
@@ -291,14 +294,24 @@ - (void) _start {
...
         }
-        c4repl_start(_repl);
+    }
+    
+    return _repl;
+}
+
+- (void) _start {
+    C4Error err;
+    C4Replicator* repl = [self c4repl: &err];
```
pasin: Please do warning the error and return here if repl is nil. You can move CBLWarnError from c4repl: method here.

jayahariv:
> I think this PR is not completed as it handle or create c4repl when getting pending docs or check if pending doc if _repl is nullptr. 

This part is not in this PR. I will be adding that soon. 

> The complication could be with the retry logic if the _repl is recreated again.

We will keep using the `_repl` in case of retry logic and everything else, only place we will be lazy initializing will be inside the `start method` and `pendingDocIds method`. Which makes less impact right?

pasin:
I think the logic should be completed in this PR as we want to do lazy init to handle the pendingDocIds logic.

pasin:
> We will keep using the _repl in case of retry logic and everything else, only place we will be lazy initializing will be inside the start method and pendingDocIds method. Which makes less impact right?

There is a scenario that the pendingDocIDs method might be called when the replicator is offline (the _repl is null) so the _repl will be recreated. However the current logic in the retry method checks whether _repl is null to be able to retry. That needs to be changed a bit.

jayahariv:
current PR is not making whole `_repl` instance lazy init, but it is using `c4repl`, which will be used by start and pending document ids.

Do you think we should only use `_repl`?

jayahariv:
got it, will make the change and resubmit soon.

jayahariv:
issue with this PR can be resolved by editing the destination branch to another one and back to previous.

PR #2567, opened 2019-11-01
CBL-449: swift implementation - pending document ids

jayahariv:
* swift implementation - pending document ids
* add attribute for marking nonnull error to throw in objc function declaration.

### DIFF
--- a/Objective-C/CBLReplicator.h
+++ b/Objective-C/CBLReplicator.h
@@ -158,9 +158,9 @@ typedef struct {
  change between the time the call was made and the time the call returns.
  
  @param error error On return, the error if any.
- @return A  set of document Ids, each of which has one or more pending revisions
+ @return A  set of document Ids, each of which has one or more pending revisions. If error, nil.
  */
-- (NSSet<NSString*>*) pendingDocumentIds: (NSError**)error;
+- (nullable NSSet<NSString*>*) pendingDocumentIds: (NSError**)error;
 
 /**
  Checks if the document with the given ID has revisions pending push.  This API is a snapshot and results may
--- a/Objective-C/CBLReplicator.h
+++ b/Objective-C/CBLReplicator.h
@@ -160,7 +160,7 @@ typedef struct {
  @param error error On return, the error if any.
  @return A  set of document Ids, each of which has one or more pending revisions
  */
-- (NSSet<NSString*>*) pendingDocumentIds: (NSError**)error __attribute__((swift_error(nonnull_error)));
+- (NSSet<NSString*>*) pendingDocumentIds: (NSError**)error NS_SWIFT_NOTHROW;
 
 /**
  Checks if the document with the given ID has revisions pending push.  This API is a snapshot and results may
@@ -171,7 +171,7 @@ typedef struct {
  @return true if the document has one or more revisions pending, false otherwise
  
  */
-- (BOOL) isDocumentPending: (NSString*)documentID error: (NSError**)error __attribute__((swift_error(nonnull_error)));
+- (BOOL) isDocumentPending: (NSString*)documentID error: (NSError**)error NS_SWIFT_NOTHROW;
 
 @end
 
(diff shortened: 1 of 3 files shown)


Objective-C/CBLReplicator.h
```diff
@@ -160,7 +160,7 @@ typedef struct {
  @param error error On return, the error if any.
  @return A  set of document Ids, each of which has one or more pending revisions
  */
-- (NSSet<NSString*>*) pendingDocumentIds: (NSError**)error;
+- (NSSet<NSString*>*) pendingDocumentIds: (NSError**)error __attribute__((swift_error(nonnull_error)));
```
jayahariv: i couldn't find any Apple documentation, but clang doc has [this](https://github.com/apple/swift-clang/blob/383859a9c4b964af3d127b5cc8abd0a8f11dd164/include/clang/Basic/AttrDocs.td#L1800-L1819), also can seen in FBSDK

Objective-C/CBLReplicator.h
```diff
@@ -160,7 +160,7 @@ typedef struct {
  @param error error On return, the error if any.
  @return A  set of document Ids, each of which has one or more pending revisions
  */
-- (NSSet<NSString*>*) pendingDocumentIds: (NSError**)error;
+- (NSSet<NSString*>*) pendingDocumentIds: (NSError**)error NS_SWIFT_NOTHROW;
```
jayahariv: better solution, and found recommended Apple way of doing 
https://developer.apple.com/documentation/swift/cocoa_design_patterns/about_imported_cocoa_error_parameters

jayahariv:
in case of `- (BOOL) isDocumentPending`
> If an error producing Objective-C method returns a BOOL value to indicate the success or failure of a method call, Swift changes the return type of the function to Void

in case of ` - (NSSet<NSString*>*) pendingDocumentIds`
If the result is non-nullable, it cannot convert to throwable.

Swift/Replicator.swift
```diff
@@ -175,6 +175,35 @@ public final class Replicator {
...
+    }
+    
+    /// Checks if the document with the given ID has revisions pending push.  This API is a snapshot and
+    /// results may change between the time the call was made and the time the call returns.
+    ///
+    /// - Parameter documentID: The ID of the document to check
+    /// - Returns: true if the document has one or more revisions pending, false otherwise
+    public func isDocumentPending(_ documentID: String) throws -> Bool {
+        var error: NSError?
+        let result = _impl.isDocumentPending(documentID, error: &error)
```
pasin: I have never used `NS_SWIFT_NOTHROW` before. Can you explain why we don't let the objective-c thrown the exception so we don't need to check the error?

pasin: Never mind, I got your point here.

Swift/Replicator.swift
```diff
@@ -175,6 +175,35 @@ public final class Replicator {
         _impl.removeChangeListener(with: token._impl)
     }
     
+    /// Gets a set of document Ids, who have revisions pending push. This API is a snapshot and results
+    /// may change between the time the call was made and the time the call returns.
+    ///
+    /// - Returns: A  set of document Ids, each of which has one or more pending revisions
+    public func pendingDocumentIds() throws -> Set<String> {
+        var error: NSError?
+        let result = _impl.pendingDocumentIds(&error)
```
pasin: The objective-C -pendingDocumentIds should return null if there is an error occurred. Then this place we an let the objective-c thrown in swift so we don't need to check the error.

jayahariv: that makes sense

Objective-C/CBLReplicator.mm
```diff
@@ -457,24 +457,24 @@ - (void) removeChangeListenerWithToken: (id<CBLListenerToken>)token {
...
     if (c4err.code > 0) {
         convertError(c4err, error);
         CBLWarnError(Sync, @"Error while fetching pending documentIds: %d/%d", c4err.domain, c4err.code);
-        return [NSSet set];
+        return nil;
     }
     
     if (result.size <= 0)
-        return [NSSet set];
+        return nil;
```
pasin: If result's size == 0, it should return an empty set. This is not an error case.

PR #2569, opened 2019-11-02
CBL-517: Lazy C4Replicator init

jayahariv:
Same as #2566 

Sorry for multiple rebasing, somehow the PR was not showing me the exact changes, kept rebasing, and ended up same commits 3 times, on top of same master branch.

### DIFF
--- a/Objective-C/CBLReplicator.mm
+++ b/Objective-C/CBLReplicator.mm
@@ -134,6 +134,79 @@ typedef enum {
     return _config;
 }
 
+- (void) dealloc {
+    c4repl_free(_repl);
+    [_reachability stop];
+}
+
+- (NSString*) description {
+    if (!_desc)
+        _desc = [NSString stringWithFormat: @"%@[%s%s%s %@]",
+                 self.class,
+                 (isPull(_config.replicatorType) ? "<" : ""),
+                 (_config.continuous ? "*" : "-"),
+                 (isPush(_config.replicatorType)  ? ">" : ""),
+                 _config.target];
+    return _desc;
+}
+
+- (void) clearRepl {
+    c4repl_free(_repl);
+    _repl = nullptr;
+}
+
+- (void) start {
+    CBL_LOCK(self) {
+        CBLLogInfo(Sync, @"%@: Starting...", self);
+        if (_state != kCBLStateStopped && _state != kCBLStateSuspended) {
+            CBLWarn(Sync, @"%@ has already started (state = %d, status = %d); ignored.",
+                    self,  _state, _rawStatus.level);
+            return;
+        }
+        
+        _retryCount = 0;
+        [self _start];
+    }
+}
+
+- (void) resetAndScheduleRetry {
+    [self resetRetryCount];
+    [self scheduleRetry: 0];
+}
+
+- (void) scheduleRetry: (NSTimeInterval)delayInSeconds {
+    [self cancelRetry];
+    
+    __weak CBLReplicator *weakSelf = self;
+    _retryTimer = [CBLTimer scheduleIn: _dispatchQueue
+                                 after: delayInSeconds
+                                 block: ^{
+        [weakSelf _retry];
+    }];
+}
+
+- (void) resetRetryCount {
+    CBLLogVerbose(Sync, @"%@: Resetting the retry count", self);
+    _retryCount = 0;
+}
+
+- (void) _retry {
+    CBL_LOCK(self) {
+        CBLLogInfo(Sync, @"%@: Retrying...", self);
+        if (_repl || _state <= kCBLStateStopping) {
+            CBLLogInfo(Sync, @"%@: Ignore retrying (state = %d, status = %d)",
+                       self, _state, _rawStatus.level);
+            return;
+        }
+        [self _start];
+    }
+}
+
+- (void) cancelRetry {
+    [CBLTimer cancel: _retryTimer];
+    _retryTimer = nil;
+}
+
 - (C4Replicator*) c4repl: (nullable C4Error*)c4err {
     if (_repl)
         return _repl;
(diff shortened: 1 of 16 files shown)


Objective-C/CBLReplicator.mm
```diff
@@ -193,7 +195,7 @@ - (void) resetRetryCount {
 - (void) _retry {
     CBL_LOCK(self) {
         CBLLogInfo(Sync, @"%@: Retrying...", self);
-        if (_repl || _state <= kCBLStateStopping) {
+        if ((_repl && _state != kCBLStateCreated) || _state <= kCBLStateStopping) {
```
pasin: Probably change to if not offline nor suspended?

Objective-C/CBLReplicator.mm
```diff
@@ -350,7 +364,7 @@ - (void) stop {
 - (void) _stop {
     [self cancelRetry];
     
-    if (_repl) {
+    if (_repl && _state != kCBLStateCreated) {
```
pasin: what happens when calling stop on non starting _repl? is there anything we can get from _repl itself that it hasn't been started yet?

jayahariv: it will crash, since the rawStatus.level after creating is connecting. It will fail somewhere [here](https://github.com/couchbaselabs/BLIP-Cpp/blob/8bbba50dcb3d023eb07df0f0649389e7c8c003a9/include/blip_cpp/LoopbackProvider.hh#L252)

Objective-C/CBLReplicator.mm
```diff
@@ -75,7 +75,7 @@ static NSTimeInterval retryDelay(unsigned retryCount) {
     kCBLStateSuspending,        ///< The replicator was asked to suspend but in progress.
     kCBLStateOffline,           ///< The replicator is offline due to a transient or network error.
     kCBLStateRunning,           ///< The replicator is running which is either idle or busy.
-    kCBLStateStarting           ///< The replicator was asked to start but in progress.
+    kCBLStateStarting          ///< The replicator was asked to start but in progress.
```
jayahariv:
```suggestion
    kCBLStateStarting           ///< The replicator was asked to start but in progress.
```

Objective-C/CBLReplicator.mm
```diff
@@ -137,6 +137,7 @@ - (CBLReplicatorConfiguration*) config {
 - (void) dealloc {
     c4repl_free(_repl);
     [_reachability stop];
+    _repl = nullptr;
```
jayahariv:
```suggestion
```

Objective-C/CBLReplicator.mm
```diff
@@ -350,7 +360,7 @@ - (void) stop {
 - (void) _stop {
     [self cancelRetry];
     
-    if (_repl) {
+    if (_repl && (_rawStatus.level != kC4Connecting || _rawStatus.level != kC4Offline)) {
```
pasin: Can you explain about the condition here? I couldn't follow.

Objective-C/CBLReplicator.mm
```diff
@@ -460,13 +470,19 @@ - (void) removeChangeListenerWithToken: (id<CBLListenerToken>)token {
...
+        CBLWarnError(Sync, @"%@: Replicator cannot be created: %@", self, err.localizedDescription);
+        if (error)
+            *error = err;
         return nil;
     }
     
-    C4Error c4err = {};
-    C4SliceResult result = c4repl_getPendingDocIDs(_repl, &c4err);
+    c4err = {};
+    C4SliceResult result = c4repl_getPendingDocIDs(repl, &c4err);
```
pasin: Should we call this in CBL_LOCK(self) to avoid race condition if repl is freed when stopped or offline?

Objective-C/CBLReplicator.mm
```diff
@@ -494,14 +510,20 @@ - (BOOL) isDocumentPending: (NSString*)documentID error: (NSError**)error {
...
+        if (error)
+            *error = err;
         return false;
     }
     
-    C4Error c4err = {};
+    c4err = {};
     CBLStringBytes docID(documentID);
-    BOOL isPending = c4repl_isDocumentPending(_repl, docID, &c4err);
+    BOOL isPending = c4repl_isDocumentPending(repl, docID, &c4err);
```
pasin: Should we call this in CBL_LOCK(self) to avoid race condition if repl is freed when stopped or offline?

Objective-C/CBLReplicator.mm
jayahariv: i take it back, this won't happen. Since `[replicator stop]` function will not allow it to pass. Since when creating a new c4replicator, without setting the `_state`, it defaults to `kCBLStateStopped`.

Objective-C/CBLReplicator.mm
```diff
@@ -233,7 +237,7 @@ - (void) _start {
         Assert(remoteURL, @"Endpoint has no URL");
 #endif
     }
-
+
```
jayahariv:
```suggestion

```

Objective-C/CBLReplicator.mm
```diff
@@ -233,7 +237,6 @@ - (void) _start {
         Assert(remoteURL, @"Endpoint has no URL");
 #endif
     }
```
jayahariv:
```suggestion
    }

```

Objective-C/CBLReplicator.mm
```diff
@@ -193,7 +193,7 @@ - (void) resetRetryCount {
 - (void) _retry {
     CBL_LOCK(self) {
         CBLLogInfo(Sync, @"%@: Retrying...", self);
-        if (_repl || _state <= kCBLStateStopping) {
+        if ((_repl && (_state != kCBLStateOffline || _state != kCBLStateSuspended)) || _state <= kCBLStateStopping) {
```
pasin: Hope we could simplify this logic but not the big deal. 

Another idea to make it readable is to do:
```
- (BOOL) isOffline { return (_state != kCBLStateOffline || _state != kCBLStateSuspended); }

if ((_repl && [self isOffline]) || _state <= kCBLStateStopping)
```
OR 
define IS_OFFLINE (_state != kCBLStateOffline || _state != kCBLStateSuspended)

jayahariv: when c4replicator is lazily initialized, its state will be stopped. this means, 
1. if replicator goes offline. 
2. user check for pending-document-ids 
3. comes to online, and retries, it won't allow to start again(since _state = stopped). unless we mark `_state` to something else.

Checking for `_repl && ![self isOffline]` condition will always be false, since we always clear the replicator just before setting both `kCBLStateOffline` and `kCBLStateSuspended` state. 

am i right or missing something?

jayahariv: I will verify this using a unit test and see.

jayahariv: nevermind. i am wrong

jayahariv: there is something wrong with this. I will investigate further

jayahariv:
closing this for now!!

PR #2576, opened 2019-11-13
fix: memory leak due to holding fleecedata

jayahariv:
* previously we were using Doc instance, which was expecting a alloc_slice, but we supply a FLSliceResult which was already owned, and it again retained and owned in the constructor.
* also the encode result FLSliceResult was not released after update/create the c4Doc.

### DIFF
--- a/Objective-C/CBLDatabase.mm
+++ b/Objective-C/CBLDatabase.mm
@@ -925,8 +925,9 @@ static C4DatabaseConfig c4DatabaseConfig (CBLDatabaseConfiguration *config) {
         CBLStringBytes docID(document.id);
         *outDoc = c4doc_create(_c4db, docID, (FLSlice)body, revFlags, &err);
     }
-    
+
     FLSliceResult_Free(body);
+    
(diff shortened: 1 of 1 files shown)


Objective-C/CBLDatabase.mm
```diff
@@ -900,29 +900,33 @@ - (BOOL) saveDocument: (CBLDocument*)document
...
     if (c4Doc) {
-        *outDoc = c4doc_update(c4Doc, body, revFlags, &err);
+        *outDoc = c4doc_update(c4Doc, (FLSlice)body, revFlags, &err);
     } else {
         CBLStringBytes docID(document.id);
-        *outDoc = c4doc_create(_c4db, docID, body, revFlags, &err);
+        *outDoc = c4doc_create(_c4db, docID, (FLSlice)body, revFlags, &err);
     }
     
+    FLSliceResult_Free(body);
```
pasin: Can you add an empty line after this so that it is not in the same group of checking error?

pasin: The rest is good.

PR #2577, opened 2019-11-13
CBL-514: Add logging to try to track down P2P failures

borrrden:


pasin:
@borrrden What do you want to do with this PR?

borrrden:
It was never my intention to merge this.  This was for a problem that only seemed to happen on the PR validation machine.  What is the status of this problem?

pasin:
The PR validation seems to be fine.

borrrden:
I know, I wasn't able to repro the problem.

PR #2578, opened 2019-11-13
CBL-520: include revision-id with filters

jayahariv:
* include revision-id with document
* check c4doc exists, if not return the revision-id
* include revision-id check with all filter unit tests
* update the revision-id signature form lite-core.
* when tries to create the c4doc_singleRevision, but deleted document returns nil with NotFound error. so in that case, we might again needs to save the revision-id and return. so not possible to create c4doc with revision-id, document-id and other info.

### DIFF
--- a/Objective-C/CBLDocument.mm
+++ b/Objective-C/CBLDocument.mm
@@ -125,6 +125,9 @@ using namespace fleece;
 }
 
 - (CBLMutableDocument*) toMutable {
+    if (_revID)
+        [NSException raise: NSInternalInconsistencyException
+                    format: @"%@", kCBLErrorMessageNoDocEditInReplicationFilter];
     return [self mutableCopy];
 }
 
--- a/Objective-C/CBLDocument.mm
+++ b/Objective-C/CBLDocument.mm
@@ -46,6 +46,7 @@ using namespace fleece;
     if (self) {
         _database = database;
         _id = documentID;
+        _revID = nil;
         [self setC4Doc: c4Doc];
     }
     return self;
@@ -60,6 +61,7 @@ using namespace fleece;
         _database = database;
         _id = documentID;
         _fleeceData = body;
+        _revID = nil;
         [self updateDictionary];
     }
     return self;
@@ -90,6 +92,7 @@ using namespace fleece;
     self = [self initWithDatabase: database documentID: documentID c4Doc: nil];
     if (self) {
         _database = database;
+        _revID = nil;
         CBLStringBytes docId(documentID);
         C4Error err;
         auto doc = c4doc_get(database.c4db, docId, true, &err);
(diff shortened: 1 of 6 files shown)


pasin:
There is an edge case that this change might not be working perfectly. Basically the doc from the filter cannot be used to edit and save correctly as the c4doc doesn't actually exist. This seems to be OK with the pull filter because the doc doesn't actually exist on the local database but a bit wrong with the push filter.

pasin:
I could think of two ways to deal with this now:
1. Document to the API doc that the Document from the filter is just a placeholder to observe information. This means that editing and saving the doc will have the same effect as saving a newly created document which could cause a conflict.
2. Trying to be smart about it for push filter by loading the C4Doc. I'm not sure about the comment from @jayahariv that loading the deleted doc causes NOT_FOUND error. For pull filter, we still need to document and print out the warning similar to the first method.

@borrrden what do you think?

borrrden:
2 is a rabbit hole that borders on absurdity.  We are providing a filter function to examine documents, not to make changes to them an save them.  That feels like an abuse.  With this change it's actually possible to detect that we are creating a document for the purpose of a filter so I vote for an option 3:  Throwing an exception if this sort of document is attempted to be saved.

pasin:
I agreed with @borrrden. I think the best now is to not allowing them to do toMutable() so that they cannot edit and save. I have also checked with @rajagp about the issue. Let's throw an exception when users to toMutable() with the message such as "The documents from the replication filter are not allowed to edit". We should update the API doc regarding this as well. I will update the design doc.

Sandychuang8:
Let's make this message "The documents from the replication filter are not allowed to edit." global through out the platforms...  
https://github.com/couchbaselabs/couchbase-mobile-tools/blob/master/gen_errors/cbl-errors.json#L39

bmeike:
I would say that we might be able to do better that: "The documents from the replication filter are not allowed to edit."   Perhaps: "Documents from a replication filter may not be edited."?

pasin:
Let's use the message that @bmeike suggested.

bmeike:
Hey... I as I implemented this, I came to the conclusion that it is a class invariant that: `c4doc == null || refId == null`.  I think you might want to enforce that.  I think that `setC4Doc` should set `refId = null`

jayahariv:
iOS tests will fail, if we don't allow in case of `c4doc == nil`. 
we could do `_revID != nil && c4Doc == nil`

bmeike:
Yeah... definitely not suggesting that you require c4doc == nil.  That doesn't make any sense, for sure.  But you *do* allow setting c4doc.  If that happens when revId != nil, the instance is in an inconsistent state.  I think you should set revId = null when you set c4doc.

jayahariv:
if i understand correctly, that is set in all constructors, which are the only places from c4Doc or revIds are set.

bmeike:
Oh!  In java there is a `setC4Doc` that, you know, sets the `c4doc` ;-P   I thought I saw one in the Obj-C code, too.

jayahariv:
Yes, there is one in ObjC too. But I could see one more constructor which also requires `revID` to be nil.

PR #2583, opened 2019-11-18
revert pending documents id 

jayahariv:


bmeike:
Ouch!  I hope you have that on a branch, somewhere.  Is there no way, in Obj-C, simply to make it invisible in the API?

jayahariv:
everything in the git history

PR #2585, opened 2019-11-19
fix: leak with collator

jayahariv:
* _operator was assigned with QueryInstance([here](https://github.com/couchbase/couchbase-lite-ios/blob/6b09f2d2f51b1dbcb406fbdddd46bd46b0841f41/Objective-C/CBLQueryExpression.m#L310)), and was stuck without getting released when the collator itself was released. 
* cbl-540

### DIFF
--- a/Objective-C/Internal/CBLCollationExpression.m
+++ b/Objective-C/Internal/CBLCollationExpression.m
@@ -41,4 +41,9 @@
     return @[@"COLLATE", [_collation asJSON], [_operand asJSON]];
 }
 
+- (void) dealloc {
+    _collation = nil;
+    _operand = nil;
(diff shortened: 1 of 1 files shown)


Objective-C/Internal/CBLCollationExpression.m
```diff
@@ -41,4 +41,9 @@ - (id) asJSON {
     return @[@"COLLATE", [_collation asJSON], [_operand asJSON]];
 }
 
+- (void) dealloc {
+    _collation = nil;
+    _operand = nil;
```
pasin: I understand that it will be automatically released. Do we need to set them to nil?

PR #2586, opened 2019-11-20
CBL-518: CBL-519: blob

jayahariv:
* cbl-518 
* cbl-519

* when blob content not used, it is not closed. so release it in dealloc
* return nil, in case of contentStream property asks for it before getting the contents.
* make it in sync with .NET and Android

### DIFF
--- a/Objective-C/CBLBlob.mm
+++ b/Objective-C/CBLBlob.mm
@@ -135,6 +135,12 @@ static NSString* const kBlobType = @kC4ObjectType_Blob;
     return self;
 }
 
+- (void) dealloc {
+    if (_initialContentStream)
+        [_initialContentStream close];
+    _initialContentStream = nil;
+}
+
 - (NSDictionary*) properties {
     if (_properties) {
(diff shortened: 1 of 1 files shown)


Objective-C/CBLBlob.mm
```diff
@@ -135,6 +135,12 @@ - (instancetype) initWithDatabase: (CBLDatabase*)db
     return self;
 }
 
+- (void) dealloc {
```
bmeike: wooo.... I better confirm that I do the same!

bmeike: Whew.  I do.

pasin:
Some tests failed. Please check.

bmeike:
"No space left on device"

jayahariv:
I have just reverted to investigate the failures, seems to be legit failure with swift test. Will resubmit again soon.

jayahariv:
https://github.com/couchbase/couchbase-lite-ios/pull/2588

PR #2588, opened 2019-11-21
CBL-518: CBL-519: fix: blob

jayahariv:
* corrected objc test for blob with object stream
* when blob content not used, it is not closed. so release it in dealloc
* return nil, in case of contentStream property asks for it before getting the contents.
* make it in sync with .NET and Android

### DIFF
--- a/Objective-C/CBLBlob.mm
+++ b/Objective-C/CBLBlob.mm
@@ -280,7 +280,7 @@ static NSString* const kBlobType = @kC4ObjectType_Blob;
         uint8_t buffer[kReadBufferSize];
         NSInteger bytesRead = 0;
         _length = 0;
-        NSInputStream *contentStream = [self contentStream];
+        NSInputStream *contentStream = _initialContentStream;
         [contentStream open];
         success = true;
         while(success && (bytesRead = [contentStream read:buffer maxLength: kReadBufferSize]) > 0) {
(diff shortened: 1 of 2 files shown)


Objective-C/CBLBlob.mm
```diff
@@ -275,7 +280,7 @@ - (BOOL) installInDatabase: (CBLDatabase*)db error:(NSError**)outError {
         uint8_t buffer[kReadBufferSize];
         NSInteger bytesRead = 0;
         _length = 0;
-        NSInputStream *contentStream = [self contentStream];
+        NSInputStream *contentStream = _initialContentStream;
```
pasin: I think this is not correct. For example, if blob was initiated with content (Data). Probably need to get from _initialContentStream or [self contentStream] if _initialContentStream is null.

pasin: Can we also check if we have unit tests to cover all constructors?

jayahariv: it is working fine with the updated unit test. Since if it is initialized with content(Data), it will initiate via `success = c4blob_create(store, data2slice(_content), nullptr, &key, &err);`

pasin: That is right.

pasin:
Umm seems like tests still failed.

PR #2591, opened 2019-11-25
CBL-556: Only retry if there is any network issues 

jayahariv:
* Only retry if there is any network issues
* previously we were setting observer in case of all errors.


jayahariv:
i will wait for the CI to  finish and merge this PR

borrrden:
This won't help if the server side hangs up, the reachability start will just simply reconnect immediately right?

jayahariv:
true, if the error is not a transient this will start again.

jayahariv:
I will be creating another PR to include a fix for that.

jayahariv:
https://github.com/couchbase/couchbase-lite-ios/pull/2592 fix for the issue

PR #2592, opened 2019-11-26
CBL-556: When permanent failure

jayahariv:
* fix when permanent failure happens, it shouldn't  invoke network reachability. 
* not tested, since i cannot able to reproduce a permanent failure from my side.

### DIFF
--- a/Objective-C/CBLReplicator.mm
+++ b/Objective-C/CBLReplicator.mm
@@ -550,11 +550,14 @@ static void statusChanged(C4Replicator *repl, C4ReplicatorStatus status, void *c
         CBLLogInfo(Sync, @"%@: Transient error (%@); will retry in %.0f sec...",
                    self, error.localizedDescription, delay);
         [self scheduleRetry: delay];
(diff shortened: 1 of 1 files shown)


Objective-C/CBLReplicator.mm
```diff
@@ -550,11 +550,14 @@ - (bool) handleError: (C4Error)c4err {
         CBLLogInfo(Sync, @"%@: Transient error (%@); will retry in %.0f sec...",
                    self, error.localizedDescription, delay);
         [self scheduleRetry: delay];
-    } else {
+    } else if (!_reachability.reachable) {
```
jayahariv: is there any possibility that reachability can be turned back on, when the control reaches here? which means it will stay disconnected.

pasin: Permanent error when not come here. It should be already returned.

jayahariv:
This PR is anyways invalid. closing and will open if anything else fails.

PR #2597, opened 2020-01-06
CBL-563: refac: reachability initialized on start

jayahariv:
* previously we were only initializing the replicator when there is some transient error happened.
* now refactored to initialize on start and start the observer when there is any transient error.
* move all reachability functions to a pragma mark.
* seems only the startReachability on dispatchQueue is only used, and run-loop is not used in code.
* convert the global variable "allowReachability" to a local one, since we are initializing on start.
* in case the reachability instance is not initialized, added a verbose log to print, for debugging.

### DIFF
--- a/Objective-C/CBLReplicator.mm
+++ b/Objective-C/CBLReplicator.mm
@@ -430,6 +430,8 @@ static C4ReplicatorValidationFunction filter(CBLReplicationFilter filter, bool i
         return;
     
     _reachability = [[CBLReachability alloc] initWithURL: remoteURL];
+    __weak auto weakSelf = self;
+    _reachability.onChange = ^{ [weakSelf reachabilityChanged]; };
 }
 
 - (void) startReachabilityObserver {
@@ -438,8 +440,6 @@ static C4ReplicatorValidationFunction filter(CBLReplicationFilter filter, bool i
         return;
     }
     
-    __weak auto weakSelf = self;
-    _reachability.onChange = ^{ [weakSelf reachabilityChanged]; };
     [_reachability startOnQueue: _dispatchQueue];
 }
 
--- a/Objective-C/CBLReplicator.mm
+++ b/Objective-C/CBLReplicator.mm
@@ -552,7 +552,7 @@ static void statusChanged(C4Replicator *repl, C4ReplicatorStatus status, void *c
     convertError(c4err, &error);
     
     // Transient error or maybe still reachable:
-    if (transient || _reachability.reachableWithoutStart) {
+    if (transient || _reachability.reachable) {
         // On transient error, retry periodically, with exponential backoff:
         auto delay = retryDelay(++_retryCount);
         CBLLogInfo(Sync, @"%@: Transient error (%@); will retry in %.0f sec...",
--- a/Objective-C/CBLReplicator.mm
+++ b/Objective-C/CBLReplicator.mm
@@ -552,7 +552,7 @@ static void statusChanged(C4Replicator *repl, C4ReplicatorStatus status, void *c
     convertError(c4err, &error);
     
     // Transient error or maybe still reachable:
-    if (transient || _reachability.reachable) {
+    if (transient || _reachability.reachableWithoutStart) {
         // On transient error, retry periodically, with exponential backoff:
         auto delay = retryDelay(++_retryCount);
         CBLLogInfo(Sync, @"%@: Transient error (%@); will retry in %.0f sec...",
--- a/Objective-C/CBLReplicator.mm
+++ b/Objective-C/CBLReplicator.mm
@@ -91,7 +91,6 @@ typedef enum {
     CBLReplicatorState _state;
     C4ReplicatorStatus _rawStatus;
     unsigned _retryCount;
-    BOOL _allowReachability;
     CBLReachability* _reachability;
     CBLReplicatorProgressLevel _progressLevel;
     CBLChangeNotifier<CBLReplicatorChange*>* _changeNotifier;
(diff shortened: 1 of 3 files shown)


jayahariv:
Tested `startOnDispatchQueue` code by, changing the network status in Todo App, but not validated the `startOnRunloop`, since its not used anywhere in the code. 

When reading the code, every time we start observer, `ClientCallback` will be scheduled on run-loop or dispatch-queue, and on stop, it will be unscheduled from it. 

Seems fine to me, @pasin do you think anything else I need to double check?

pasin:
@jayahariv Can you remind me again what the specific problem that this change fixes? I remembered that it could help correcting something but I forgot.

jayahariv:
1. [here](https://github.com/couchbase/couchbase-lite-ios/blob/master/Objective-C/CBLReplicator.mm#L547) even if the `_reachability.reachable` is reachable, during the first visit this will be return false, since we allocate `_reachability` after the checking [here](https://github.com/couchbase/couchbase-lite-ios/blob/master/Objective-C/CBLReplicator.mm#L557) only. 
2. second point was, no need to keep allocating the `_reachability` again and again, since we only needs to do once, and do multiple times the start/stop observer. Keeping `_reachability` instance won't be expensive, but started observer is the one which is expensive. This is more like a good practice. 
3. [Not fixed in this PR] We are notifying immediately after starting the observer [here](https://github.com/couchbase/couchbase-lite-ios/blob/master/Objective-C/Internal/Replicator/CBLReachability.m#L136z). Which is not causing anything bad, but i forgot something @borrrden  was saying about this.

Objective-C/CBLReplicator.mm
```diff
@@ -450,6 +419,45 @@ - (void) removeChangeListenerWithToken: (id<CBLListenerToken>)token {
...
+        return;
+    }
+    
+    [_reachability startOnQueue: _dispatchQueue];
+}
+
+// Should be called from _dispatchQueue
+- (void) stopReachabilityObserver {
+    [_reachability stop];
+}
```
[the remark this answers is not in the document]
jayahariv: no, we need the reachability instance to be alive as long as replicator is alive. 

this PR is supposed to keep reachability instance populated, but observer is stopped and started again when in need.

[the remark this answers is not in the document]
bmeike: Yep.  Cool.

pasin:
I think it solved the 2nd point but not the 1st point because _reachability.reachable will still return false if the reachability is not started or stopped.

borrrden:
If that's the case then this basically solves nothing.  The bad situation is this:

1. We are online, but reachability says `false` because of this flaw
2. We start reachability.  
3. Reachability calls back immediately with `true` (the value that it should have been) because it never was `false` in the first place and we get an immediate retry instead of a delayed one.

pasin:
@jayahariv If the reachability is unknown, you can try to get the current reachability without start listening by calling SCNetworkReachabilityGetFlags().

jayahariv:
let me double check the `_reachability.reachable`, without starting it.

jayahariv:
yes, `SCNetworkReachabilityGetFlags` will return the correct value. 

How about including a new property to `CBLReachability` say, `reachableWithoutStart` which can check for reachability with `SCNetworkReachability` and without the `_reachabilityKnown` or `_reachabilityFlags`

Objective-C/CBLReplicator.mm
```diff
@@ -544,7 +552,7 @@ - (bool) handleError: (C4Error)c4err {
     convertError(c4err, &error);
     
     // Transient error or maybe still reachable:
-    if (transient || _reachability.reachable) {
+    if (transient || _reachability.reachableWithoutStart) {
```
jayahariv: update: new method for check without start

borrrden:
I don't see any use for the normal `reachable` anymore in this scenario.  Why not just change `reachable` to do what `reachableWithoutStart` does?  Or maybe do it only if unknown.

jayahariv:
When network status changes, `reachabilityChanged` function will get called, inside the function, `_reachability.reachable` is checked([here](https://github.com/couchbase/couchbase-lite-ios/blob/c75f0ac0a970d999231f448c4be44855964eb892/Objective-C/CBLReplicator.mm#L403))

pasin:
Instead of having two methods, I think what @borrrden suggested is a better fix. It should be fine that the _reachability.reachable is also accessed in the `reachabilityChanged`.

Objective-C/Internal/Replicator/CBLReachability.m
```diff
@@ -205,9 +205,11 @@ - (NSString*) description {
 
 - (BOOL) reachable {
     // We want 'reachable' flag to be on, but not if user intervention is required (like PPP login)
-    return _reachabilityKnown
-        &&  (_reachabilityFlags & kSCNetworkReachabilityFlagsReachable)
-        && !(_reachabilityFlags & kSCNetworkReachabilityFlagsInterventionRequired);
+    SCNetworkReachabilityFlags flag;
+    if (SCNetworkReachabilityGetFlags(_ref, &flag))
```
pasin: I think if the _rechabilityKnow is true, we can just use _reachabilityFlags. You will need to change the implementation of the status method as well.

pasin: BTW, The CBLReachability is not thread-safe by its implementation which is fine as long as the CBLReplicator doesn't call it from multiple threads (please double check as well).

jayahariv: all the `_reachability` instances are called from lock.

PR #2600, opened 2020-01-10
CBL-534: update enterprise license(master)

jayahariv:
* updated the license from `LA v5: 03262019` to `LA v7: 12Nov2019`


jayahariv:
Grab it from documents repo, that would make us to add doc repo as submodule, which seems to be an overkill right?

Usage of this license text is rare, only when someone creates the release zip for enterprise from other than build machines.

PR #2604, opened 2020-02-07
fix: CBL-649: handle encode error

jayahariv:
* previously when blob throws an nsexception it was returned back to CBL side, recently the Fleece APIs are noexcept, so exceptions thrown can be ignored and only look for error object.
* remove try..catch since the methods are no longer throwable.
* use utility method when creating the CBLError.
* fetch Fleece Error Message before finishing the Encoding, so if there is any error happened, we can pass the error message too.
* related PR: https://github.com/couchbaselabs/fleece/pull/69

### DIFF
--- a/Objective-C/CBLMutableDocument.mm
+++ b/Objective-C/CBLMutableDocument.mm
@@ -183,9 +183,10 @@
         return {};
     }
     FLError flErr;
+    const char* errMessage = FLEncoder_GetErrorMessage(encoder);
     FLSliceResult body = FLEncoder_Finish(encoder, &flErr);
     if (!body.buf)
-        convertError(flErr, outError);
+        createError(flErr, [NSString stringWithUTF8String: errMessage], outError);
     return body;
 }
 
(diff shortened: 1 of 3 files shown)


Objective-C/Tests/ReplicatorTest+CustomConflict.m
```diff
@@ -867,6 +867,7 @@ - (void)testConflictResolverReturningBlobFromDifferentDB {
     [replicator removeChangeListenerWithToken: token];
     
     // using blob from remote document of user's- which is a different database
+
```
```suggestion
```

Objective-C/CBLMutableDocument.mm
```diff
@@ -183,9 +183,10 @@ - (C4SliceResult) encode: (NSError**)outError {
         return {};
     }
     FLError flErr;
+    const char* errMessage = FLEncoder_GetErrorMessage(encoder);
     FLSliceResult body = FLEncoder_Finish(encoder, &flErr);
     if (!body.buf)
```
bmeike: Shouldn't you look up the error message only if you need it?

jayahariv: there is an issue with it, we knew if there is any error once the `FLEncoder_Finish` function finishes(from FLError), and once it is finish, it will reset the encoder, so the previous recorded error message will be empty.

PR #2606, opened 2020-02-12
Merge lite core changes

jayahariv:
partial merge to master (with latest lite core)

* includes API change in lite core, fleece etc..  release instead of free
* new replicator changes like, 
  * remove retry logic(move to lite core)
  * remove suspended state(move to lite core)
  * remove error handling(moved to lite core)
* disable some tests since, they are failing and not yet analyzed, will commit them via separate PR to master. 
* mbedtls, mbedcrypto, mbedx509 flags included.


Objective-C/CBLReplicator.mm
```diff
@@ -148,11 +129,6 @@ - (NSString*) description {
     return _desc;
 }
 
-- (void) clearRepl {
-    c4repl_free(_repl);
-    _repl = nullptr;
-}
-
```
bmeike: You can still tell that it has been closed/freed?

jayahariv: i am assuming, we are not deleting the c4replicator instance, and it should be handled by lite core.

bmeike: 👍

Objective-C/CBLReplicator.mm
```diff
@@ -713,33 +607,13 @@ - (BOOL) active {
...
-                return;
-            }
-            
-            if (suspended) {
-                _state = kCBLStateSuspending;
-                [self _stop];
-            } else
-                [self resetAndScheduleRetry];
-        }
+        c4repl_setSuspended(_repl, suspended);
```
pasin: @jayahariv Have you looked if the logic to [1] ignore the suspend when the replicator is in the stop process or [2] to resume from suspending (e.g. switching fast b/w BG and FG) is still working?

At least we should create a ticket to verify after all other unit tests are passed.

jayahariv: sure, will create a ticket to verify this

jayahariv: https://issues.couchbase.com/browse/CBL-687

PR #2610, opened 2020-02-21
CBL-703: Pending Document Ids

jayahariv:
* include pending document ids API
* include unit tests

### DIFF
--- a/Objective-C/CBLReplicator.h
+++ b/Objective-C/CBLReplicator.h
@@ -153,6 +153,25 @@ typedef struct {
  */
 - (void) removeChangeListenerWithToken: (id<CBLListenerToken>)token;
 
+/**
+ Gets a set of document Ids, who have revisions pending push. This API is a snapshot and results may
+ change between the time the call was made and the time the call returns.
+ 
+ @param error error On return, the error if any.
+ @return A  set of document Ids, each of which has one or more pending revisions. If error, nil.
+ */
+- (nullable NSSet<NSString*>*) pendingDocumentIds: (NSError**)error;
+
+/**
+ Checks if the document with the given ID has revisions pending push.  This API is a snapshot and results may
+ change between the time the call was made and the time the call returns.
+ @param documentID The ID of the document to check
+ @param error error On return, the error if any.
+ @return true if the document has one or more revisions pending, false otherwise
+ 
+ */
+- (BOOL) isDocumentPending: (NSString*)documentID error: (NSError**)error;
+
 @end
 
 
--- a/Objective-C/CBLReplicator.h
+++ b/Objective-C/CBLReplicator.h
@@ -160,7 +160,7 @@ typedef struct {
  @param error error On return, the error if any.
  @return A  set of document Ids, each of which has one or more pending revisions. If error, nil.
  */
-- (nullable NSSet<NSString*>*) pendingDocumentIds: (NSError**)error;
+- (nullable NSSet<NSString*>*) pendingDocumentIDs: (NSError**)error;
 
 /**
  Checks if the document with the given ID has revisions pending push.  This API is a snapshot and results may
--- a/Objective-C/CBLReplicator.mm
+++ b/Objective-C/CBLReplicator.mm
@@ -358,7 +358,7 @@ static C4ReplicatorValidationFunction filter(CBLReplicationFilter filter, bool i
     }
 }
 
-- (NSSet<NSString*>*) pendingDocumentIds: (NSError**)error {
+- (NSSet<NSString*>*) pendingDocumentIDs: (NSError**)error {
     if (_config.replicatorType > 1) {
         if (error)
             *error = [NSError errorWithDomain: CBLErrorDomain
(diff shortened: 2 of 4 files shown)


Objective-C/CBLReplicator.mm
```diff
@@ -358,6 +358,67 @@ - (void) removeChangeListenerWithToken: (id<CBLListenerToken>)token {
...
+    if (_config.replicatorType > 1) {
+        if (error)
+            *error = [NSError errorWithDomain: CBLErrorDomain
+                                         code: CBLErrorUnsupported
+                                     userInfo: @{NSLocalizedDescriptionKey: kCBLErrorMessagePullOnlyPendingDocIDs}];
+        return nil;
+    }
+    
+    C4Error c4err = {};
+    if (![self _setupC4Replicator: &c4err]) {
```
bmeike: This starts a replicator, now?

jayahariv: it will initialize a replicator, if it already not done.

bmeike: do you have to close it or anything?

Objective-C/CBLReplicator.h
```diff
@@ -153,6 +153,25 @@ typedef struct {
...
 - (void) removeChangeListenerWithToken: (id<CBLListenerToken>)token;
 
+/**
+ Gets a set of document Ids, who have revisions pending push. This API is a snapshot and results may
+ change between the time the call was made and the time the call returns.
+ 
+ @param error error On return, the error if any.
+ @return A  set of document Ids, each of which has one or more pending revisions. If error, nil.
+ */
+- (nullable NSSet<NSString*>*) pendingDocumentIds: (NSError**)error;
```
pasin: I think it's the method is pendingDocumentIDs per the spec.

Objective-C/CBLReplicator.mm
pasin: _setupC4Replicator: should be called under the lock (same lock as the start method) for thread safety.

Objective-C/CBLReplicator.mm
```diff
@@ -358,6 +358,67 @@ - (void) removeChangeListenerWithToken: (id<CBLListenerToken>)token {
...
+        convertError(c4err, error);
+        CBLWarnError(Sync, @"%@: Replicator cannot be created: %d/%d", self, c4err.domain, c4err.code);
+        return nil;
+    }
+    
+    C4SliceResult result = c4repl_getPendingDocIDs(_repl, &c4err);
```
pasin: Question: Does the c4repl_getPendingDocIDs need to be called under a database's lock?

Objective-C/CBLReplicator.mm
```diff
@@ -358,6 +358,67 @@ - (void) removeChangeListenerWithToken: (id<CBLListenerToken>)token {
...
+    if (_config.replicatorType > 1) {
+        if (error)
+            *error = [NSError errorWithDomain: CBLErrorDomain
+                code: CBLErrorUnsupported
+            userInfo: @{NSLocalizedDescriptionKey: kCBLErrorMessagePullOnlyPendingDocIDs}];
+        return NO;
+    }
+
+    C4Error c4err = {};
+    if (![self _setupC4Replicator: &c4err]) {
```
pasin: Should be called under the same lock as the lock in the start method for thread safety.

Objective-C/CBLReplicator.mm
```diff
@@ -358,6 +358,67 @@ - (void) removeChangeListenerWithToken: (id<CBLListenerToken>)token {
...
+        convertError(c4err, error);
+        CBLWarnError(Sync, @"%@: Replicator cannot be created: %d/%d", self, c4err.domain, c4err.code);
+        return NO;
+    }
+    
+    CBLStringBytes docID(documentID);
+    BOOL isPending = c4repl_isDocumentPending(_repl, docID, &c4err);
```
pasin: Question: Does this method need be called under the database's lock?

Objective-C/CBLReplicator.h
bmeike: That would be contrary to Java naming conventions.  Do we really all have to standardize on the conventions of some language?  Not being sarcastic, here.  I don't understand the advantage of imposing some other language's conventions.  I am probably missing something...

pasin: In general, each platform can adjust the name to what makes sense to its naming conversion. For example, Objective-C has documentWithID: and Java has getDocument(String id). Please feel free to make this kind of adjustment. It might be a good idea to double check if Java has use xxxID() previously (I couldn't think anything) so we could also weight on that.

@borrrden @Sandychuang8 what do you think about this? What do you normally do on .NET especially the ID thing?

bmeike: Good point.  Java *should* use `pendingDocumentIds`, but I'll check.

Objective-C/CBLReplicator.mm
bmeike: The Java code is nearly identical.
The reference to the C4Replicator is guarded by a lock, because it could be mutated on multiple threads.  Once the code has a non-null reference, though, it presumes it can used it freely.  If that is not the case, there will have to be a *LOT* more locking in the Java code.

pasin: Yes, it's the same concept here. Basically we need to guard the C4Replicator so that it will not create multiple C4Replicator if it got called from multiple threads.

Objective-C/CBLReplicator.h
Sandychuang8: If it's a property, .Net will use PendingDocumentIds, GetPendingDocumentIds if it's a method.

Objective-C/CBLReplicator.mm
jayahariv: sure, will make the changes.

Objective-C/CBLReplicator.mm
jayahariv: Assuming this API is thread safe, able to see a mutex lock in the definition of API [here](https://github.com/couchbase/couchbase-lite-core/blob/c6f8346ea925b7ba60200ced92e3b6770d34d54c/Replicator/c4Replicator.hh#L124)

Objective-C/CBLReplicator.mm
jayahariv: same, https://github.com/couchbase/couchbase-lite-core/blob/c6f8346ea925b7ba60200ced92e3b6770d34d54c/Replicator/c4Replicator.hh#L146

Objective-C/CBLReplicator.mm
```diff
@@ -358,6 +358,73 @@ - (void) removeChangeListenerWithToken: (id<CBLListenerToken>)token {
     }
 }
 
+- (NSSet<NSString*>*) pendingDocumentIDs: (NSError**)error {
+    if (_config.replicatorType > 1) {
+        if (error)
+            *error = [NSError errorWithDomain: CBLErrorDomain
+                                         code: CBLErrorUnsupported
+                                     userInfo: @{NSLocalizedDescriptionKey: kCBLErrorMessagePullOnlyPendingDocIDs}];
+        return nil;
```
Sandychuang8: I thought we are returning empty set instead of nil?

jayahariv: Oops! i didn't see this, and you are right. it should be an empty set.

jayahariv: https://issues.couchbase.com/browse/CBL-757

PR #2616, opened 2020-02-27
CBL-717: Fix in..batch API

jayahariv:
* in batch API when thrown exception(from block) was not cancelling the transaction.
* created separate method with throwable block and start handle when swift throws exception.
* in case of objc, same issue exists. 
* include unit tests

### DIFF
--- a/Objective-C/CBLDatabase.mm
+++ b/Objective-C/CBLDatabase.mm
@@ -319,7 +319,7 @@ static void dbObserverCallback(C4DatabaseObserver* obs, void* context) {
 
 #pragma mark - BATCH OPERATION
 
-- (BOOL) inBatch: (NSError**)outError usingBlock: (void (NS_NOESCAPE ^)())block {
+- (BOOL) inBatch: (NSError**)outError usingThrowableBlock: (void (NS_NOESCAPE ^)(NSError**))block {
     CBLAssertNotNil(block);
     
     CBL_LOCK(self) {
@@ -332,7 +332,23 @@ static void dbObserverCallback(C4DatabaseObserver* obs, void* context) {
         if (!transaction.begin())
             return convertError(transaction.error(), outError);
         
-        block();
+        @try {
+            NSError* err = nil;
+            block(&err);
+            if (err) {
+                // if swift throws an exception, `err` will be populated
+                transaction.abort();
+                return createError(CBLErrorUnexpectedError,
+                                   [NSString stringWithFormat: @"%@", err.localizedDescription],
+                                   outError);
+            }
+        } @catch(NSException* e) {
+            // if objc block throws an exception, control reaches here
+            transaction.abort();
+            return createError(CBLErrorUnexpectedError,
+                               [NSString stringWithFormat: @"%@ %@ %@", e.name, e.reason, e.userInfo],
+                               outError);
+        }
         
         if (!transaction.commit())
             return convertError(transaction.error(), outError);
@@ -341,6 +357,14 @@ static void dbObserverCallback(C4DatabaseObserver* obs, void* context) {
     [self postDatabaseChanged];
     
     return YES;
+}
+
+- (BOOL) inBatch: (NSError**)outError usingBlock: (void (NS_NOESCAPE ^)())block {
+    CBLAssertNotNil(block);
+    
+    return [self inBatch: outError usingThrowableBlock:^(NSError **) {
+        block();
+    }];
 }
 
 #pragma mark - DATABASE MAINTENANCE
--- a/Objective-C/CBLDatabase.mm
+++ b/Objective-C/CBLDatabase.mm
@@ -362,9 +362,7 @@ static void dbObserverCallback(C4DatabaseObserver* obs, void* context) {
 - (BOOL) inBatch: (NSError**)outError usingBlock: (void (NS_NOESCAPE ^)())block {
     CBLAssertNotNil(block);
     
-    return [self inBatch: outError usingThrowableBlock:^(NSError **) {
-        block();
-    }];
+    return [self inBatch: outError usingThrowableBlock:^(NSError **) { block(); }];
 }
 
 #pragma mark - DATABASE MAINTENANCE
--- a/Objective-C/CBLDatabase.mm
+++ b/Objective-C/CBLDatabase.mm
@@ -319,7 +319,7 @@ static void dbObserverCallback(C4DatabaseObserver* obs, void* context) {
 
 #pragma mark - BATCH OPERATION
 
-- (BOOL) inBatch: (NSError**)outError usingThrowableBlock: (void (NS_NOESCAPE ^)(NSError**))block {
+- (BOOL) inBatch: (NSError**)outError usingBlockWithError: (void (NS_NOESCAPE ^)(NSError**))block {
     CBLAssertNotNil(block);
     
     CBL_LOCK(self) {
@@ -362,7 +362,7 @@ static void dbObserverCallback(C4DatabaseObserver* obs, void* context) {
 - (BOOL) inBatch: (NSError**)outError usingBlock: (void (NS_NOESCAPE ^)())block {
     CBLAssertNotNil(block);
     
-    return [self inBatch: outError usingThrowableBlock:^(NSError **) { block(); }];
+    return [self inBatch: outError usingBlockWithError: ^(NSError **) { block(); }];
 }
 
 #pragma mark - DATABASE MAINTENANCE
--- a/Objective-C/Internal/CBLDatabase+Swift.h
+++ b/Objective-C/Internal/CBLDatabase+Swift.h
@@ -23,4 +23,6 @@
 
 + (void) checkFileLogging: (BOOL)swift;
 
+- (BOOL) inBatch: (NSError**)error usingThrowableBlock: (void (NS_NOESCAPE ^)(NSError**))block NS_REFINED_FOR_SWIFT;
+
 @end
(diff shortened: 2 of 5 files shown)


Objective-C/CBLDatabase.mm
```diff
@@ -332,7 +332,23 @@ - (BOOL) inBatch: (NSError**)outError usingBlock: (void (NS_NOESCAPE ^)())block
         if (!transaction.begin())
             return convertError(transaction.error(), outError);
         
-        block();
+        @try {
+            NSError* err = nil;
+            block(&err);
+            if (err) {
+                // if swift throws an exception, `err` will be populated
```
bmeike: Would it be unreasonable just to throw the error, if it in not nil?  Then you wouldn't have to write that code twice.  It isn't really important to make it fast, I think: it is a failure...

Objective-C/CBLDatabase.mm
```diff
@@ -319,7 +319,7 @@ - (BOOL) purgeDocumentWithID: (NSString*)documentID error: (NSError**)error {
 
 #pragma mark - BATCH OPERATION
```
bmeike: This is an addition to the public API?  I would like to do the same thing.  Pasin?

jayahariv: this is only internal, not public.

Objective-C/Internal/CBLDatabase+Swift.h
```diff
@@ -23,4 +23,6 @@
 
 + (void) checkFileLogging: (BOOL)swift;
 
+- (BOOL) inBatch: (NSError**)error usingThrowableBlock: (void (NS_NOESCAPE ^)(NSError**))block NS_REFINED_FOR_SWIFT;
```
pasin: `usingThrowableBlock` is confusing because the block is not throwable block. How about `usingBlockWithError` or `usingBlockWithOutError`?

Objective-C/CBLDatabase.mm
bmeike: Ah.  We should put it on the agenda for 3.0 changes.  I've wanted to do the same thing.

Objective-C/CBLDatabase.mm
jayahariv: objc is not designed in a way to handle exceptions. throwing an exceptions will cause a leak, so recommended way is to grab the error and process.

Objective-C/CBLDatabase.mm
```diff
@@ -343,6 +359,12 @@ - (BOOL) inBatch: (NSError**)outError usingBlock: (void (NS_NOESCAPE ^)())block
     return YES;
 }
 
+- (BOOL) inBatch: (NSError**)outError usingBlock: (void (NS_NOESCAPE ^)())block {
+    CBLAssertNotNil(block);
+    
+    return [self inBatch: outError usingThrowableBlock:^(NSError **) { block(); }];
```
pasin: space b/w  usingThrowableBlock: and ^

Objective-C/CBLDatabase.mm
pasin: I think Java has already handled it correctly. Only Swift has an issue.

bmeike: Yes.  The point is that Java does not allow the lambda to throw.  It should. I've opened a ticket for that.

jayahariv: Swift & ObjC

pasin: I see. That is right.

Objective-C/CBLDatabase.mm
dtorres: If I may....

Swift doesn't have exceptions, what it does have is an error handling mechanism where you can "throw" them and it will return early and they must be handled by the caller with the `do {} catch {}` mechanism or marking the callsite as throwable.

These errors map directly to NSError in Objective C and vice-versa.

Exceptions in Obj-C are complicated but the gist is, you shouldn't catch them in a generic context, but only in a case by case basis where you know it's ok as most NSExceptions are thrown due to unrecoverable errors (out of memory, memory corruption, internal inconsistency, etc) which is why Swift doesn't have them, because well, you can't really do much about them but to abort the program

bmeike: Thank you for the long explanations.  I think you are missing the point.
There is an @catch at line 345, that will catch an NSException raised at line 339.   Doing this would eliminate that annoying duplicate code.

dtorres: Huh? There is no exception being raised at line 339.

If you refer to the comment:
> // if swift throws an exception, `err` will be populated

That is an NSError, which has nothing to do with the Obj-C exceptions mechanism

bmeike: Uhhh...  Oh... NVM... It's committed.

pasin: I think @dtorres made a good point about the NSException that when it got thrown from the Objective-C code, it's pretty much unrecoverable error.

jayahariv: may be the comment is misleading, `if swift throws an error, ..` instead of `throws an exception`. 

@dtorres, even though it says exception, in swift we are catching the `Error` thrown from in..bach block and send back to ObjC as `NSError`(see Swift/Database.swift). Any error happens during the batch operation, this NSError object will be populated.

And yes throwing exception from objc is unrecoverable.

bmeike: Yeah.  Really beating a dead horse, at this point.  I was just pointing out that:
1. That `@catch` catches *something*
2. The code that it executes, when it catches *something* is exactly the same code that is inside the `if` clause
3. Why not `raise` *something*, inside the `if` clause?  Then you don't have to duplicate code.

Objective-C/CBLDatabase.mm
```diff
@@ -332,7 +332,23 @@ - (BOOL) inBatch: (NSError**)outError usingBlock: (void (NS_NOESCAPE ^)())block
...
+                                   [NSString stringWithFormat: @"%@", err.localizedDescription],
+                                   outError);
+            }
+        } @catch(NSException* e) {
+            // if objc block throws an exception, control reaches here
+            transaction.abort();
+            return createError(CBLErrorUnexpectedError,
+                               [NSString stringWithFormat: @"%@ %@ %@", e.name, e.reason, e.userInfo],
+                               outError);
+        }
```
dtorres: My point is... this part of the code shouldn't exist.... particularly because of "if objc block throws an exception, control reaches here". Converting an exception (particularly one originating from code you don't control) is a crash waiting to happen down the line.

The Swift part of this is fine.

Objective-C/CBLDatabase.mm
jayahariv: Got your point Blake, but doing so is not recommended in ObjC. 
[reference objc not exceptions-safe](http://clang.llvm.org/docs/AutomaticReferenceCounting.html#exceptions)

The only way we can avoid the duplicate is to write a function, but more overhead. 

Marking this as resolved.

bmeike: I'll be darned!  TIL!  Thank you, @jayahariv !

Objective-C/CBLDatabase.mm
jayahariv: > Converting an exception (particularly one originating from code you don't control) is a crash waiting to happen down the line

If any NSExceptions are raised in client code(inside in..batch block in objc), which will get caught in this catch block rather than crash(see Objective-C/Tests/DatabaseTest.m), cancels the transaction and returns unexpected error.

rob-keepsafe: I believe what @dtorres is saying is you're very broadly swallowing _all_ exceptions here even though some of them should raised as a valid exception that stops program execution in order to prevent corruption (or very difficult to debug crashes) further down in the stack.

To reiterate his comment above:
>Exceptions in Obj-C are complicated but the gist is, you shouldn't catch them in a generic context, but only in a case by case basis where you know it's ok as most NSExceptions are thrown due to unrecoverable errors (out of memory, memory corruption, internal inconsistency, etc) 

Are there specific exceptions you're expecting here that are safe to catch versus others that should be raised and cause program execution to stop?

rob-keepsafe: See examples here of rethrowing if the local context doesn't understand / need to process what it's caught: https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/Exceptions/Tasks/NestingExceptionHandlers.html#//apple_ref/doc/uid/20000060-CJBBFDIF

And the Clang article you linked above mentions this as well:

>The standard Cocoa convention is that exceptions signal programmer error and are not intended to be recovered from. Making code exceptions-safe by default would impose severe runtime and code size penalties on code that typically does not actually care about exceptions safety.

as well as memory leaks actually being fine in this context where needed
>**Therefore, ARC-generated code leaks by default on exceptions, which is just fine if the process is going to be immediately terminated anyway. Programs which do care about recovering from exceptions should enable the option.**

rob-keepsafe: Lastly, from an old Apple dev: https://stackoverflow.com/a/3379240/308315

>Any exception that passes through system framework code will leave said framework in an undefined state.. Catching said exceptions and trying to recover from it will lead to memory leaks, undefined behaviour and crashing.

>Just because the language has syntax to support something, doesn't mean the frameworks are designed to support that thing, too. The frameworks have a very firm policy: exceptions are unrecoverable errors. Trying to catch and recover goes against framework design and, quite clearly, an exception that "jumps over" framework based stack frames is a violation of documented policy and will yield unspecified behavior.

jayahariv: got it now, thanks for pointing it out and detailed explanation. 

https://github.com/couchbase/couchbase-lite-ios/commit/ec919108a9f38deb2684f7cb9d3f8596787ac140

PR #2620, opened 2020-03-17
CBL-711: test: offline, idle tests for pending document id API

jayahariv:
* include tests to validate continuous replicator. for idle, offline states

### DIFF
--- a/Objective-C/Tests/ReplicatorTest+PendingDocIds.m
+++ b/Objective-C/Tests/ReplicatorTest+PendingDocIds.m
@@ -18,6 +18,7 @@
 //
 
 #import "ReplicatorTest.h"
+#import "CBLReplicator+Internal.h"
 
 #define kDocIdFormat @"doc-%d"
 #define kActionKey @"action-key"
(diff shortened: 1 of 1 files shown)


Objective-C/Tests/ReplicatorTest+PendingDocIds.m
```diff
@@ -236,6 +237,49 @@ - (void) testPendingDocIdsWithFilter {
...
+    
+    CBLReplicator* replicator = [[CBLReplicator alloc] initWithConfig: config];
+    __block CBLReplicator* r;
+    id<CBLListenerToken> token = [replicator addChangeListener: ^(CBLReplicatorChange* change) {
+        r = replicator;
+        NSError* err = nil;
+        NSSet* ids = [change.replicator pendingDocumentIDs: &err];
+        AssertNil(err);
+
+        if (change.status.activity == kCBLReplicatorConnecting) {
```
bmeike: I think that this test will fail if the states are updated in the wrong order.  That seems like a good thing.  I wonder, though, if you want to make that explicit: check to see that you first see CONNECTING, then IDLE, then OFFLINE.  Not necessary, I think, but it might make the test easier to read...

Objective-C/Tests/ReplicatorTest+PendingDocIds.m
```diff
@@ -353,6 +397,44 @@ - (void) testIsDocumentPendingWithPushFilter {
...
+    __block CBLReplicator* r;
+    id<CBLListenerToken> token = [replicator addChangeListener: ^(CBLReplicatorChange* change) {
+        r = replicator;
+        NSError* err = nil;
+        if (change.status.activity == kCBLReplicatorConnecting) {
+            Assert([change.replicator isDocumentPending: @"doc-1" error: &err]);
+        } else if (change.status.activity == kCBLReplicatorIdle) {
+            AssertFalse([change.replicator isDocumentPending: @"doc-1" error: &err]);
+            [r setSuspended: YES];
+
```
bmeike: Yeah... Is this synchronous?  Is there any chance you could get the OFFLINE, before you insert the document

pasin: It's asynchronous and I think it could be a race condition but pretty difficult to happen. Maybe save the doc before suspend?

bmeike: I was thinking that the test would be clearer if it asserted that the first call definitely was connecting, the second idle, and the third offline.  I agree that it isn't a big deal, though...

jayahariv: Save before suspend might also cause, replicator to push the docs before suspending. I will update to avoid the situation.

pasin: You can save and suspend in batch so that the doc will not be pushed before suspending.

jayahariv: restructured,  differently

jayahariv:
Removed the connecting state, since connecting is already covered in other tests. Made this test to cover idle and offline states alone. 

Also restructured to avoid race condition.

PR #2623, opened 2020-03-24
Move otherDB to the base test class

pasin:
* Moved otherDB to the base test class. The otherDB needs to explicitly open.
* Fixed missing header search path to fleece API.

### DIFF
--- a/Objective-C/Tests/ReplicatorTest+PendingDocIds.m
+++ b/Objective-C/Tests/ReplicatorTest+PendingDocIds.m
@@ -49,7 +49,7 @@
 }
 
 - (void) validatePendingDocumentIDs: (NSSet*)docIds count: (NSUInteger)count {
-    id target = [[CBLDatabaseEndpoint alloc] initWithDatabase: otherDB];
+    id target = [[CBLDatabaseEndpoint alloc] initWithDatabase: self.otherDB];
     id config = [self configWithTarget: target type: kCBLReplicatorTypePush continuous: NO];
 
     __block id<CBLListenerToken> token;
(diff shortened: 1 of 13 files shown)


Objective-C/Tests/ReplicatorTest+PendingDocIds.m
```diff
@@ -277,19 +281,23 @@ - (void) testPendingDocIdsWhenOffline {
...
+    NSSet* ids = [replicator pendingDocumentIDs: &err];
     AssertNil(err);
     AssertEqual(ids.count, 1);
     [ids containsObject: @"doc-2"];
     
     [replicator stop];
+    
+    // repicator is stopped
+    [self waitForExpectations: @[stopped] timeout: 5.0];
+
```
jayahariv: good catch!

Swift/Tests/MessageEndpointTest.swift
```diff
@@ -77,7 +77,7 @@ class MessageEndpointTest: CBLTestCase, MCNearbyServiceBrowserDelegate,
 MCNearbyServiceAdvertiserDelegate, MCSessionDelegate, MessageEndpointDelegate,
 MultipeerConnectionDelegate {
     
-    var otherDB: Database!
+    var oDB: Database!
```
jayahariv: name? since everywhere we named it otherDB, do you have any reason why named it oDB?

Swift/Tests/ReplicatorTest+PendingDocIds.swift
```diff
@@ -21,6 +21,9 @@ import XCTest
 import CouchbaseLiteSwift
 
 class ReplicatorTest_PendingDocIds: ReplicatorTest {
+    
+    #if COUCHBASE_ENTERPRISE
+
```
jayahariv: is this only an EE only feature? i can see its an CE feature

pasin: It is CE feature but it needs to test under EE because it uses DatabaseEndpoint (EE only).

Swift/Tests/MessageEndpointTest.swift
pasin: It's for a convenience so that it is not optional (otherDB is optional in the base case).

`oDB = otherDB!`

PR #2624, opened 2020-03-25
CBL-772 : Enhance Database Close and Delete

pasin:
* Enchance Database’s close() and delete() to stop active replicators and live quries befoer closing the database.
* Changed Database’s delete() to call close() then use static delete() to delete the database.
* Fixed live query stop logic to wait until the pending query’s execution or refreshing is done before removing itself from the database’s active live queries list.
* Not allow to start a new replicator and a live query when database is closing.


Objective-C/CBLDatabase.mm
```diff
@@ -527,8 +552,6 @@ - (void) removeChangeListenerWithToken: (id<CBLListenerToken>)token {
     CBLAssertNotNil(token);
     
     CBL_LOCK(self) {
-        [self mustBeOpen];
```
borrrden: What relevance does this have to the PR?

Objective-C/CBLDatabase.mm
```diff
@@ -360,58 +360,83 @@ - (BOOL) inBatch: (NSError**)outError usingBlock: (void (NS_NOESCAPE ^)())block
 #pragma mark - DATABASE MAINTENANCE
 
 - (BOOL) close: (NSError**)outError {
+    NSArray *activeReplicators, *activeLiveQueries = nil;
+    
     CBL_LOCK(self) {
-        if (_c4db == nullptr)
+        if ([self isClosed] || _isClosing)
```
borrrden: This seems a bit dangerous in that two rapid calls will result in the second one returning `YES` before the database is actually closed.  I think any subsequent calls should block until the close condition, and if possible report back the ultimate success of the operation from the other thread.

Objective-C/CBLReplicator.mm
```diff
@@ -301,17 +300,13 @@ - (void) stop {
 }
 
 - (void) stopped {
-    CBL_LOCK(self) {
```
borrrden: Why was this lock removed?

Objective-C/CBLDatabase.mm
pasin: Good catch. It's a mistake. Not sure why it got removed.

Objective-C/CBLReplicator.mm
pasin: Because it's already called under the same lock.

Objective-C/CBLDatabase.mm
pasin: Really good point. I will fix this and update the technical spec regarding this well.

Objective-C/CBLDatabase.mm
pasin: Actually I removed it because it doesn't need it.

Objective-C/CBLDatabase.mm
```diff
@@ -360,58 +360,83 @@ - (BOOL) inBatch: (NSError**)outError usingBlock: (void (NS_NOESCAPE ^)())block
...
         
-        [self freeC4Observer];
-        [self freeC4DB];
+        // Release database:
+        if (success)
+            [self freeC4DB];
         
-        return YES;
+        // Reset closing flag:
+        _isClosing = NO;
```
bmeike: All these flags... Would it be better to have a state machine?
CLOSED, OPEN, CLOSING.  I think that's sufficient...

Objective-C/CBLDatabase.mm
```diff
@@ -957,8 +1001,65 @@ - (BOOL) saveDocument: (CBLDocument*)document
...
+
+- (uint64_t) activeLiveQueryCount {
+    CBL_LOCK(self) {
+        return _activeLiveQueries.count;
+    }
+}
 
+#pragma mark - REPLICATOR
+
+// NOTE: These methods are called from the replicator thread.
```
bmeike: Is this true?  In this code, they seem to be called from the start methods... which is what I would expect...

Objective-C/CBLReplicator.mm
```diff
@@ -301,17 +300,13 @@ - (void) stop {
 }
 
 - (void) stopped {
```
bmeike: hmmm... this looks important.  What is it doing?

Objective-C/Internal/CBLLiveQuery.mm
```diff
@@ -32,7 +32,7 @@ @implementation CBLLiveQuery
     __weak CBLQuery* _query;
     NSTimeInterval _updateInterval;
     
-    BOOL _observing, _willUpdate;
+    BOOL _observing, _willUpdate, _updating, _stopping;
```
bmeike: Again, i wonder if a single state variable wouldn't be a better way to handle this.  Would that even be possible?

Objective-C/CBLReplicator.mm
pasin: It's just a method to set the stopped state. There are a couple of places calling that to set the replicator in stopped state.

Objective-C/CBLDatabase.mm
pasin: Yea, it's not accurate. I will change this.

Objective-C/CBLDatabase.mm
pasin: It created https://issues.couchbase.com/browse/CBL-794 to evaluate and track this.

Objective-C/Internal/CBLLiveQuery.mm
pasin: updating and stopping are independent to each other but we might be able to clean this up more.

pasin: Track it here : https://issues.couchbase.com/browse/CBL-794

pasin:
@jayahariv @Sandychuang8 Could you please review?

Objective-C/CBLReplicator.mm
bmeike: Yah... ok.  Makes perfect sense, now.  Don't know what I was thinking...

Objective-C/CBLDatabase.mm
```diff
@@ -624,9 +655,30 @@ - (nullable NSDate*) getDocumentExpirationWithID: (NSString*)documentID {
 #pragma mark - INTERNAL
 
 - (void) mustBeOpen {
-    if (_c4db == nullptr) {
+    if ([self isClosed])
         [NSException raise: NSInternalInconsistencyException
                     format: @"%@", kCBLErrorMessageDBClosed];
+}
+
+- (void) mustBeOpenLocked {
```
jayahariv: do you think this will be used other than in CBLQuery? 

if only used in CBLQuery, can we directly lock at the caller
```
CBL_LOCK(self.database) { [self.database mustBeOpen]; } 
```

Objective-C/CBLDatabase.mm
```diff
@@ -624,9 +655,30 @@ - (nullable NSDate*) getDocumentExpirationWithID: (NSString*)documentID {
...
+    CBL_LOCK(self) {
+        [self mustBeOpen];
+    }
+}
+
+- (void) mustBeOpenAndNotClosing {
```
jayahariv: probably one line of comment saying, call this function only from the lock, since `_isClosing` is getting accessed.

Objective-C/CBLDatabase.mm
jayahariv: `mustBeOpen` is called in many places inside the database classes from the lock, for consistency can we include lock inside base method `mustBeOpen` itself and remove `mustBeOpenLocked`.

Objective-C/CBLDatabase.mm
```diff
@@ -624,9 +655,30 @@ - (nullable NSDate*) getDocumentExpirationWithID: (NSString*)documentID {
...
+    if ([self isClosed] || _isClosing)
+        [NSException raise: NSInternalInconsistencyException
+                    format: @"%@", kCBLErrorMessageDBClosed];
+}
+
+- (BOOL) isClosed {
+    return _c4db == nullptr;
+}
+
+- (BOOL) isClosedLocked {
```
jayahariv: same with `isClosed`. seems all accessors are using the lock, which means we can make base method `isClosed` thread safe, and remove locked method.

Objective-C/CBLDatabase.mm
```diff
@@ -360,58 +360,89 @@ - (BOOL) inBatch: (NSError**)outError usingBlock: (void (NS_NOESCAPE ^)())block
...
-        return YES;
+        // Reset closing flag:
+        _isClosing = NO;
+        
+        return success;
     }
 }
 
+- (BOOL) isReadyToClose {
+    CBL_LOCK(self) {
```
jayahariv: wondering whether we need this lock, since the condition lock should already be locking this predicate check.

Objective-C/CBLDatabase.mm
pasin: mustBeOpen is already called inside a lock so the purpose of not having lock here because it doesn't need to lock again.

Objective-C/CBLDatabase.mm
pasin: I want to separate it out because we don't want to re-lock again.

Objective-C/CBLDatabase.mm
pasin: It's sort of having naming conversion around those methods already but I could add a comment.

Objective-C/CBLDatabase.mm
pasin: condition lock doesn't guard _activeReplicators and _activeLiveQueries from accessing by multiple threads. It is required to lock here.

jayahariv: true!

Objective-C/CBLDatabase.mm
pasin: I updated the comments. Thanks.

PR #2640, opened 2020-05-12
CBL-933: Specify VALID ARCH

jayahariv:
* only build for 64 bit simulator
* all supported device archs


jayahariv:
#2639 duplicate

PR #2642, opened 2020-05-14
CBL-817: Add unit tests for URLEndpointListener 

jayahariv:
* add test for validating  ports.
* add tests for get URLs, but test fails due to wrong scheme in URLs
* add tests for connection status, but again fails the active connection count, lite core fails.

### DIFF
--- a/Objective-C/Tests/URLEndpointListenerTest.m
+++ b/Objective-C/Tests/URLEndpointListenerTest.m
@@ -37,121 +37,116 @@ typedef CBLURLEndpointListener Listener;
 - (Listener*) listenAt: (uint16)port {
     Config* config = [[Config alloc] initWithDatabase: self.otherDB];
     config.port = port;
-    config.disableTLS = YES;
-    Listener* list = [[Listener alloc] initWithConfig: config];
+    Listener* listener = [[Listener alloc] initWithConfig: config];
     
     // start listener
     NSError* err = nil;
-    Assert([list startWithError: &err]);
+    Assert([listener startWithError: &err]);
     AssertNil(err);
     
-    return list;
+    return listener;
 }
 
 - (void) testPort {
     // initialize a listener
     Config* config = [[Config alloc] initWithDatabase: self.otherDB];
     config.port = kPort;
-    config.disableTLS = YES;
-    Listener* list = [[Listener alloc] initWithConfig: config];
-    AssertEqual(list.port, 0);
+    Listener* listener = [[Listener alloc] initWithConfig: config];
+    AssertEqual(listener.port, 0);
     
     // start listener
     NSError* err = nil;
-    Assert([list startWithError: &err]);
+    Assert([listener startWithError: &err]);
     AssertNil(err);
-    AssertEqual(list.port, kPort);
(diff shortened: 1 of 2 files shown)


Objective-C/Tests/URLEndpointListenerTest.m
```diff
@@ -0,0 +1,162 @@
...
+typedef CBLURLEndpointListener Listener;
+
+@end
+
+@implementation URLEndpointListenerTest
+
+- (Listener*) listenAt: (uint16)port {
+    Config* config = [[Config alloc] initWithDatabase: self.otherDB port: port];
+    config.disableTLS = YES;
+    Listener* list = [[Listener alloc] initWithConfig: config];
```
pasin, line 40: I feel like list is not a good variable name for this as list also means List. Can we change it to listener?

Objective-C/Tests/URLEndpointListenerTest.m
```diff
@@ -0,0 +1,162 @@
...
+- (void) testBusyPort {
+    Listener* list1 = [self listenAt: kPort];
+    
+    // initialize a listener at same port
+    Config* config = [[Config alloc] initWithDatabase: self.otherDB port: kPort];
+    config.disableTLS = YES;
+    Listener* list2 = [[Listener alloc] initWithConfig: config];
+    
+    // already in use when starting the second listener
+    [self ignoreException:^{
```
pasin, line 95: Why do we need to ignore the exception here?

Objective-C/Tests/URLEndpointListenerTest.m
```diff
@@ -0,0 +1,162 @@
...
+    }];
+    
+    // stops
+    [list1 stop];
+}
+
+// TODO: https://issues.couchbase.com/browse/CBL-948
+- (void) _testURLs {
+    Config* config = [[Config alloc] initWithDatabase: self.otherDB port: kPort];
+    config.disableTLS = YES;
```
pasin, line 109: We also need to test without disabling TLS as well.

Objective-C/Tests/URLEndpointListenerTest.m
jayahariv: may be, the exception thrown from lite core might not be caught in between? will check

Objective-C/Tests/URLEndpointListenerTest.m
jayahariv: should we enable TLS for all tests? or only a few to verify?

pasin: TLS mode is the default mode so all the tests should test with TLS. Ideally, we possible we test both TLS and non-TLS but some may not need such as testPort.

Objective-C/Tests/URLEndpointListenerTest.m
jayahariv: already existing port will throw exception from liter core. so the problem lies in hitting the exception break point

jayahariv: will keep this as same. and merging, will fix this if any issue exists.

PR #2649, opened 2020-05-22
Change per client cert auth implementation and TLSIdentity refactoring

pasin:
* Moved CBLClientCertAuthenticator to EE
* Added but disabled (for now) Client Cert Auth test
* Added createSecError() in CBLStatus
* CBL-819, CBL-820


Objective-C/Tests/URLEndpointListenerTest.m
```diff
@@ -245,4 +251,51 @@ - (void) testPaswordAuthenticator {
     [listener stop];
 }
 
+// Disable until completing change in CBLWebSocket:
+- (void) _testClientCertAuthenticatorWithBlock API_AVAILABLE(ios(10.3), macosx(10.5)) {
+    if (!self.keyChainAccessAllowed) return;
+    
+    // Listener:
+    CBLListenerCertificateAuthenticator* listenerAuth =
+    [[CBLListenerCertificateAuthenticator alloc] initWithBlock:
```
jayahariv, line 260: Not particular to this variable alone, but i am also starting to feel the class names are really big. Should we do something on API level? I mean renaming it to something more easy? 

URLEndpointListener, URLEndpointListenerConfiguration, ListenerCertificateAuthenticator etc.. can we limit to two words, somehow?

Objective-C/Tests/URLEndpointListenerTest.m
```diff
@@ -245,4 +251,51 @@ - (void) testPaswordAuthenticator {
...
+    
+    CBLReplicatorConfiguration* config = nil;
+    CBLClientCertificateAuthenticator* auth = nil;
+    auth = [[CBLClientCertificateAuthenticator alloc] initWithIdentity: identity1];
+    pinServerCert = NO;
+    config = [self configWithTarget: listener.localEndpoint
+                               type: kCBLReplicatorTypePushAndPull
+                         continuous: NO
+                      authenticator: auth];
+    [self run: config errorCode: 0 errorDomain: nil];
```
jayahariv, line 295: validate replication, at least a document between listener and replicator?

Objective-C/Tests/URLEndpointListenerTest.m
pasin: IMO, long is fine if its not too long. We tried not to abbreviate name if not necessarily in 2.0 API so name could be long.

Anyhow, if you have some suggestion about the name, can you comment in the API doc? 
For changing the name of the class  which is cross-platform change, we need to do it from the API doc and review.

Objective-C/Tests/URLEndpointListenerTest.m
pasin: I think it’s optional as the replicator is connected and caught up.

Objective-C/Tests/URLEndpointListenerTest.m
pasin: There are some other options provided during the API review. There were mixed comments during that time. URLEndpointListener is not ideal as it's little long but it matches with the URLEndpoint used for connecting to the listener. It's little late but not impossible to change the name. Please shout it loud and come up with some suggestion if you think we should change the name (either email or comments on API doc).

jayahariv: Yea, will update the API doc. I was accepted with the name at first, then each time when i see/write the class name(especially in ObjC), this starts to annoy.

Objective-C/Tests/URLEndpointListenerTest.m
jayahariv: Agreed, as long as we capture all errors thrown.

PR #2655, opened 2020-05-28
Add swift client cert auth tests

pasin:
* Added tests for Swift’s Client Cert Authentication.
* Added convenience run method that will create config and replicator at the same time.
* Fixed bugs not checking the error when running the Replicator in tests.
* Cleaned up some method signatures in tests.

### DIFF
--- a/Swift/Tests/CBLTestCase.swift
+++ b/Swift/Tests/CBLTestCase.swift
@@ -60,6 +60,8 @@ class CBLTestCase: XCTestCase {
     override func setUp() {
         super.setUp()
         
+        Database.log.console.level = .info
+        
         try? deleteDB(name: databaseName);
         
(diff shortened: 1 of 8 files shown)


Swift/Tests/CBLTestCase.swift
```diff
@@ -60,6 +60,8 @@ class CBLTestCase: XCTestCase {
     override func setUp() {
         super.setUp()
         
+        Database.log.console.level = .info
```
jayahariv, line 63: planned to remove?

pasin: I think about keeping it as now we didn't log anything when running Swift Tests.

jayahariv: logs might be noisy? but i guess it should be okay.

PR #2658, opened 2020-05-29
CBL-957: multiple replicators/listeners test

jayahariv:
* when replicating from multiple replicators to same database.
* two listener on same database

### DIFF
--- a/Objective-C/Tests/URLEndpointListenerTest.m
+++ b/Objective-C/Tests/URLEndpointListenerTest.m
@@ -560,7 +560,8 @@ typedef CBLURLEndpointListener Listener;
     AssertEqual(self.otherDB.count, 1);
 }
 
-- (void) testMultipleReplicatorsToListener {
+// TODO: https://issues.couchbase.com/browse/CBL-1033
+- (void) _testMultipleReplicatorsToListener {
     if (!self.keyChainAccessAllowed) return;
     
     [self listen]; // writable listener
@@ -577,7 +578,8 @@ typedef CBLURLEndpointListener Listener;
     [_listener stop];
 }
 
-- (void) testMultipleReplicatorsOnReadOnlyListener {
+// TODO: https://issues.couchbase.com/browse/CBL-1033
+- (void) _testMultipleReplicatorsOnReadOnlyListener {
     if (!self.keyChainAccessAllowed) return;
     
     Config* config = [[Config alloc] initWithDatabase: self.otherDB];
--- a/Objective-C/Tests/URLEndpointListenerTest.m
+++ b/Objective-C/Tests/URLEndpointListenerTest.m
@@ -114,6 +114,93 @@ typedef CBLURLEndpointListener Listener;
     return _listener;
 }
 
+// Two replicators, replicates docs to the listener; validates connection status
+- (void) validateMultipleReplicationsTo: (Listener*)listener replType: (CBLReplicatorType)type {
+    XCTestExpectation* exp1 = [self expectationWithDescription: @"replicator#1 stopped"];
+    XCTestExpectation* exp2 = [self expectationWithDescription: @"replicator#2 stopped"];
+    
+    NSUInteger existingDocsInListener = listener.config.database.count;
+    
+    // open DBs
+    NSError* err = nil;
+    Assert([self deleteDBNamed: @"db1" error: &err], @"Failed to delete db1 %@", err);
+    Assert([self deleteDBNamed: @"db2" error: &err], @"Failed to delete db2 %@", err);
+    CBLDatabase* db1 = [self openDBNamed: @"db1" error: &err];
+    AssertNil(err);
+    CBLDatabase* db2 = [self openDBNamed: @"db2" error: &err];
+    AssertNil(err);
+    
+    // For keeping the replication long enough to validate connection status, we will use blob
+    NSData* content = [@"i am a blob" dataUsingEncoding: NSUTF8StringEncoding];
+    
+    // DB#1
+    CBLBlob* blob1 = [[CBLBlob alloc] initWithContentType: @"text/plain" data: content];
+    CBLMutableDocument* doc1 = [self createDocument: @"doc-1"];
+    [doc1 setValue: blob1 forKey: @"blob"];
+    Assert([db1 saveDocument: doc1 error: &err], @"Fail to save db1 %@", err);
+    
+    // DB#2
+    CBLBlob* blob2 = [[CBLBlob alloc] initWithContentType: @"text/plain" data: content];
+    CBLMutableDocument* doc2 = [self createDocument: @"doc-2"];
+    [doc2 setValue: blob2 forKey: @"blob"];
+    Assert([db2 saveDocument: doc2 error: &err], @"Fail to save db2 %@", err);
+    
+    // replicators
+    CBLReplicatorConfiguration *config1, *config2;
+    config1 = [[CBLReplicatorConfiguration alloc] initWithDatabase: db1 target: listener.localEndpoint];
+    config2 = [[CBLReplicatorConfiguration alloc] initWithDatabase: db2 target: listener.localEndpoint];
+    config1.replicatorType = type;
+    config2.replicatorType = type;
+    config1.pinnedServerCertificate = (__bridge SecCertificateRef) listener.config.tlsIdentity.certs[0];
+    config2.pinnedServerCertificate = (__bridge SecCertificateRef) listener.config.tlsIdentity.certs[0];
+    CBLReplicator* repl1 = [[CBLReplicator alloc] initWithConfig: config1];
+    CBLReplicator* repl2 = [[CBLReplicator alloc] initWithConfig: config2];
+    
+    // get listener status
+    __block Listener* weakListener = _listener;
+    __block uint64_t maxConnectionCount = 0, maxActiveCount = 0;
+    id changeListener = ^(CBLReplicatorChange * change) {
+        Listener* strongListener = weakListener;
+        if (change.status.activity == kCBLReplicatorBusy) {
+            maxConnectionCount = MAX(strongListener.status.connectionCount, maxConnectionCount);
+            maxActiveCount = MAX(strongListener.status.activeConnectionCount, maxActiveCount);
+        } else if (change.status.activity == kCBLReplicatorStopped) {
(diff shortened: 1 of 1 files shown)


Objective-C/Tests/URLEndpointListenerTest.m
```diff
@@ -437,4 +437,202 @@ - (void) testUnavailableNetworkInterface {
...
+             continuous: NO
+          authenticator: nil
+             serverCert: (__bridge SecCertificateRef) listener1.config.tlsIdentity.certs[0]
+              errorCode: 0
+            errorDomain: nil];
+    
+    [listener1 stop];
+    [listener2 stop];
+    
+    AssertEqual(self.otherDB.count, 1);
```
pasin, line 464: 1. Can you add allow keychain check? 
2. Can you clean up the identities of the two listeners?

Objective-C/Tests/URLEndpointListenerTest.m
```diff
@@ -437,4 +437,202 @@ - (void) testUnavailableNetworkInterface {
...
+    // cleanup
+    [_listener stop];
+    [repl1 removeChangeListenerWithToken: token1];
+    [repl2 removeChangeListenerWithToken: token2];
+    repl1 = nil;
+    repl2 = nil;
+    Assert([db2 delete: &err]);
+    AssertNil(err);
+    db2 = nil;
+    [self deleteDBNamed: @"db2" error: &err];
```
pasin, line 551: Same as line 548?

Objective-C/Tests/URLEndpointListenerTest.m
```diff
@@ -437,4 +437,202 @@ - (void) testUnavailableNetworkInterface {
...
+            errorDomain: nil];
+    
+    [listener1 stop];
+    [listener2 stop];
+    
+    AssertEqual(self.otherDB.count, 1);
+}
+
+- (void) testMultipleReplicatorsOnSameListener {
+    XCTestExpectation* exp1 = [self expectationWithDescription: @"replicator#1 stopped"];
```
pasin, line 468: Can you add allow keychain check?
Can you clean up the identity of the listener?

Objective-C/Tests/URLEndpointListenerTest.m
```diff
@@ -437,4 +437,202 @@ - (void) testUnavailableNetworkInterface {
...
+    [_listener stop];
+    [repl1 removeChangeListenerWithToken: token1];
+    [repl2 removeChangeListenerWithToken: token2];
+    repl1 = nil;
+    repl2 = nil;
+    Assert([db2 close: &err]);
+    AssertNil(err);
+    db2 = nil;
+
+    [self deleteDBNamed: @"db2" error: &err];
```
pasin, line 634: Same as line 630?

pasin: Can you add allow keychain check?
Can you clean up the identity of the listener?

Objective-C/Tests/URLEndpointListenerTest.m
```diff
@@ -437,4 +437,202 @@ - (void) testUnavailableNetworkInterface {
...
+                [exp2 fulfill];
+        }
+    };
+    id token1 = [repl1 addChangeListener: changeListener];
+    id token2 = [repl2 addChangeListener: changeListener];
+    
+    // start & wait for replication
+    [repl1 start];
+    [repl2 start];
+    [self waitForExpectations: @[exp1, exp2] timeout: 5];
```
pasin, line 531: If you plan to have several tests to test single lister and multiple replicators like this, can you add a method to facilitate this listener status check?

Objective-C/Tests/URLEndpointListenerTest.m
```diff
@@ -437,4 +437,202 @@ - (void) testUnavailableNetworkInterface {
...
+    [repl2 removeChangeListenerWithToken: token2];
+    repl1 = nil;
+    repl2 = nil;
+    Assert([db2 delete: &err]);
+    AssertNil(err);
+    db2 = nil;
+    [self deleteDBNamed: @"db2" error: &err];
+}
+
+- (void) testMultipleReplicatorsOnReadOnlyListener {
```
pasin, line 554: What is the goal of this test? If the goal to test pull and push-and-pull replicator, I personally like to make it into two separate this so that it's simpler and easier to understand. Up to you though.

Objective-C/Tests/URLEndpointListenerTest.m
```diff
@@ -437,4 +437,202 @@ - (void) testUnavailableNetworkInterface {
...
+    config2.pinnedServerCertificate = (__bridge SecCertificateRef) _listener.config.tlsIdentity.certs[0];
+    CBLReplicator* repl2 = [[CBLReplicator alloc] initWithConfig: config2];
+    
+    // get listener status
+    __block Listener* weakListener = _listener;
+    __block uint64_t maxConnectionCount = 0, maxActiveCount = 0;
+    id changeListener = ^(CBLReplicatorChange * change) {
+        Listener* listener = weakListener;
+        if (change.status.activity == kCBLReplicatorBusy) {
+            maxConnectionCount = MAX(listener.status.connectionCount, maxConnectionCount);
```
pasin, line 516: There could be a race condition when testing with multiple single shot replicators. For example, if one replicator stops very fast before the other one is busy, the maxConnectionCount could be just 1. Testing with continuous replicators will help on this.

Objective-C/Tests/URLEndpointListenerTest.m
```diff
@@ -437,4 +437,202 @@ - (void) testUnavailableNetworkInterface {
...
+            maxActiveCount = MAX(listener.status.activeConnectionCount, maxActiveCount);
```
pasin, line 517: Same race condition could happen. I don't have good ideas to avoid the race here. I don't know how to make sure that multiple replicators will be active at the same time.

pasin: Or you can check that maxActiveCount is b/w 1 and 2.

Objective-C/Tests/URLEndpointListenerTest.m
jayahariv: to avoid the race, we are passing in a blob. which should at least take some time to finish.

Objective-C/Tests/URLEndpointListenerTest.m
jayahariv: goal is to test multiple replicators simultaneously pull from read only listener.

Objective-C/Tests/URLEndpointListenerTest.m
jayahariv: listener cleanup should be done in the teardown method in this case right?

pasin: True.

Objective-C/Tests/URLEndpointListenerTest.m
jayahariv: only one listener, and expecting it will automatically gets cleaned up.

Objective-C/Tests/URLEndpointListenerTest.m
jayahariv: >=1 && <=2 ; won't be enough. I can make an image replicate, which can sure be taken few seconds. in turn make sure, both listeners active same time.

Objective-C/Tests/URLEndpointListenerTest.m
pasin: If blob is big enough, it would help reducing or avoid the race. If you decided to use blob, please add some comments to the code (in case the test intermittently fails in the future).

Wondering if the push/pull filter can be used as a workaround to sync b/w two replicators so that we can really  ensure that both of them alive at the same time.

Objective-C/Tests/URLEndpointListenerTest.m
```diff
@@ -460,11 +462,20 @@ - (void) testMultipleListenersOnSameDatabase {
     
     [listener1 stop];
     [listener2 stop];
+    [self ignoreException:^{
+        NSError* error = nil;
+        Assert([listener1.config.tlsIdentity deleteFromKeyChainWithError: &error],
+               @"Couldn't delete identity1: %@", error);
+        Assert([listener2.config.tlsIdentity deleteFromKeyChainWithError: &error],
+               @"Couldn't delete identity2: %@", error);
+    }];
```
pasin, line 471: This test creates two listeners and starts them manually. The TLSIdentities of the two listeners needs to be cleaned up at the end of the test.

Objective-C/Tests/URLEndpointListenerTest.m
pasin: If blob is big enough, it would help reducing or avoid the race. If you decided to use blob, please add some comments to the code (in case the test intermittently fails in the future).

Objective-C/Tests/URLEndpointListenerTest.m
pasin: ok

Objective-C/Tests/URLEndpointListenerTest.m
jayahariv: do you think anything else to be cleaned up?

pasin: listener1.config.tlsIdentity and listener2.config.tlsIdentity.

jayahariv: line 554 & 556 ?

pasin: Oh, I didn't see that. Thanks.

Objective-C/Tests/URLEndpointListenerTest.m
```diff
@@ -114,6 +114,93 @@ - (Listener*) listen: (Config*)config errorCode: (NSInteger)code errorDomain: (n
...
+                [exp2 fulfill];
+        }
+    };
+    id token1 = [repl1 addChangeListener: changeListener];
+    id token2 = [repl2 addChangeListener: changeListener];
+    
+    // start & wait for replication
+    [repl1 start];
+    [repl2 start];
+    [self waitForExpectations: @[exp1, exp2] timeout: 5];
```
pasin, line 180: Can we use the ReplicatorTest.timeout?

PR #2666, opened 2020-06-03
CBL-957: Listener and Replicator on same database

jayahariv:
* listener and replicator on same db.
* continuous replicator till data is finished replicating.


jayahariv:
do you think we should have separate test for validating readOnly listener here?

PR #2674, opened 2020-06-10
Fix missed handling the replicator in stopping state

pasin:
* Make sure to not set the internal state to others after the replicator is in stopping state.
* Make sure to not start the reachability in stopping state.
* Make sure to check whether the reachability is still mornitoring instead of checking if the instance is nil or not as now when we stop the reachability, we don’t set the reachability to nil.

#CBL-1053


pasin:
@jayahariv I'm merging this now as I want to test on Jenkins. Could you please review this when you have time? Thanks.

PR #2679, opened 2020-06-11
Refactor CBLMockConnection to fix intermittently failure in tests

pasin:
* Fixed threading and misordering of messages sent when closing the connection with error.
* Refactored to not use queue and ensure thread safe as necessary.
* Fixed code style formatting.


pasin:
Merging now for testing on Jenkins. Hopefully this will fix all weird issues in MessageEndpoint Peer-to-Peer tests.

PR #2684, opened 2020-06-16
enable empty network interface

jayahariv:
* skip link local, and validate the remaining addresses
* update lite core, so that the ipv6 changes are taken

### DIFF
--- a/Objective-C/Tests/URLEndpointListenerTest.m
+++ b/Objective-C/Tests/URLEndpointListenerTest.m
@@ -624,16 +624,20 @@ typedef CBLURLEndpointListener Listener;
     [self stopListener: listener];
 }
 
-- (void) _testEmptyNetworkInterface {
+- (void) testEmptyNetworkInterface {
     if (!self.keyChainAccessAllowed) return;
     
     [self listen];
-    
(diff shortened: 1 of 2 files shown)


Objective-C/Tests/URLEndpointListenerTest.m
```diff
@@ -624,16 +624,20 @@ - (void) testServerCertVerificationModeCACert {
...
-- (void) _testEmptyNetworkInterface {
+- (void) testEmptyNetworkInterface {
     if (!self.keyChainAccessAllowed) return;
     
     [self listen];
-    
     NSArray* urls = _listener.urls;
+    
+    /** Link local addresses cannot be assigned via network interface because they don't map to any given interface.  */
+    NSPredicate* p = [NSPredicate predicateWithFormat: @"NOT(SELF.host BEGINSWITH 'fe80::')"];
```
pasin, line 634: Wondering if LiteCore should filter this out?

jayahariv: https://github.com/couchbase/couchbase-lite-core/blob/master/REST/tests/SyncListenerTest.cc#L184

PR #2688, opened 2020-06-20
CBL-384: Update cookie logic

jayahariv:
* save and retrieve cookie from lite core via c4db api
* foundation helper class added for nsstring  and nsurl.
* cblcookie class added to extract cookie from a replicator instance
* fixes cookie split logic, since previous separation logic was based on comma separated. 

Fix: CBL-681


Objective-C/Internal/CBLCookie.h
```diff
@@ -0,0 +1,30 @@
...
+//
+
+#import <Foundation/Foundation.h>
+#import "CBLReplicator.h"
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface CBLCookie : NSObject
++ (NSString*) getCookiesForReplicator: (CBLReplicator*)r;
+
```
pasin, line 27: Extra line?

Objective-C/Internal/CBLCookie.h
```diff
@@ -0,0 +1,30 @@
...
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+//
+
+#import <Foundation/Foundation.h>
+#import "CBLReplicator.h"
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface CBLCookie : NSObject
```
pasin, line 25: Can you note that this is for test to get the cookies?

Objective-C/Internal/Foundation+CBL.h
```diff
@@ -0,0 +1,36 @@
...
+
+#import <Foundation/Foundation.h>
+#import "c4Base.h"
+#import "fleece/slice.hh"
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface NSString (CBL)
+- (C4Slice) c4slice;
+
```
pasin, line 28: extra empty line? What is the convention here?

Objective-C/Internal/Foundation+CBL.h
```diff
@@ -0,0 +1,36 @@
...
+//
+
+#import <Foundation/Foundation.h>
+#import "c4Base.h"
+#import "fleece/slice.hh"
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface NSString (CBL)
+- (C4Slice) c4slice;
```
pasin, line 27: We have been using CBLStringBytes for this. Can we use CBLStringBytes instead?

Objective-C/Internal/Foundation+CBL.mm
```diff
@@ -0,0 +1,37 @@
...
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+//
+
+#import "Foundation+CBL.h"
+
+@implementation NSString (CBL)
+- (C4Slice) c4slice {
+    return {self.UTF8String, self.length};
```
pasin, line 24: self.length is not correct. It's not the number of bytes.

Objective-C/Internal/Replicator/CBLWebSocket.mm
```diff
@@ -374,10 +380,23 @@ - (void) _sendWebSocketRequest {
...
+    if (sessionCookie.buf)
+        [_logic addValue: sessionCookie.asNSString()  forHTTPHeaderField: @"Cookie"];
+    
+    C4Error err = {};
+    C4Address addr = {};
+    [_remoteURL c4Address: &addr];
+    C4SliceResult cookies = c4db_getCookies(_c4db, addr, &err);
+    if (!cookies.buf) {
+        CBLWarnError(WebSocket, @"Error getting cookies %d/%d", err.domain, err.code);
+        // TODO: any more handling?
```
pasin, line 394: Warning is OK.

Objective-C/Internal/Replicator/CBLWebSocket.mm
```diff
@@ -374,10 +380,23 @@ - (void) _sendWebSocketRequest {
...
-
+    
+    slice sessionCookie = _options[kC4ReplicatorOptionCookies].asString();
+    if (sessionCookie.buf)
+        [_logic addValue: sessionCookie.asNSString()  forHTTPHeaderField: @"Cookie"];
+    
+    C4Error err = {};
+    C4Address addr = {};
+    [_remoteURL c4Address: &addr];
+    C4SliceResult cookies = c4db_getCookies(_c4db, addr, &err);
```
pasin, line 391: I don't have the info about its thread safe behavior. Can you check if we need to lock the db or not?

pasin: What do you think about doing is in CBLReplicator versus here?

Objective-C/Internal/Replicator/CBLWebSocket.mm
```diff
@@ -443,13 +462,37 @@ - (void) receivedProxyHTTPResponse: (CFHTTPMessageRef)httpResponse {
...
-        // CFHTTPMessage incorrectly merges multiple Set-Cookie headers. Undo that:
-        NSMutableDictionary* newHeaders = [headers mutableCopy];
-        newHeaders[@"Set-Cookie"] = [cookie componentsSeparatedByString: @", "];
-        headers = newHeaders;
+    if (cookie.length > 0) {
+        C4Error err = {};
+        if (!c4db_setCookie(_c4db, cookie.c4slice, _remoteURL.host.c4slice,
+                            _remoteURL.path.stringByDeletingLastPathComponent.c4slice, &err)) {
+            CBLWarnError(WebSocket, @"Cannot save cookie %d/%d", err.domain, err.code);
+            // TODO: any more handling?
```
pasin, line 473: Warning is fine.

Objective-C/Internal/Replicator/CBLWebSocket.mm
```diff
@@ -443,13 +462,37 @@ - (void) receivedProxyHTTPResponse: (CFHTTPMessageRef)httpResponse {
...
+        }
+        
+        // update the response headers! do we really need this now? since we are already saving to c4db?
+        // https://issues.couchbase.com/browse/CBL-681
+        NSError* error = NULL;
+        NSRegularExpression* r = [NSRegularExpression regularExpressionWithPattern: @"\\,\\s*\\w+="
```
pasin, line 479: The old logic is incorrect?

Objective-C/Internal/Replicator/CBLWebSocket.mm
```diff
@@ -443,13 +462,37 @@ - (void) receivedProxyHTTPResponse: (CFHTTPMessageRef)httpResponse {
...
+    // save to lite core
     NSString* cookie = headers[@"Set-Cookie"];
-    if ([cookie containsString: @", "]) {
-        // CFHTTPMessage incorrectly merges multiple Set-Cookie headers. Undo that:
-        NSMutableDictionary* newHeaders = [headers mutableCopy];
-        newHeaders[@"Set-Cookie"] = [cookie componentsSeparatedByString: @", "];
-        headers = newHeaders;
+    if (cookie.length > 0) {
+        C4Error err = {};
+        if (!c4db_setCookie(_c4db, cookie.c4slice, _remoteURL.host.c4slice,
```
pasin, line 470: It seems like c4db_setCookie handles only one cookie and CFHTTPMessage merged all into one. We need to break them first before set the cookie.

Objective-C/Internal/Replicator/CBLWebSocket.mm
```diff
@@ -443,13 +462,37 @@ - (void) receivedProxyHTTPResponse: (CFHTTPMessageRef)httpResponse {
...
+                            _remoteURL.path.stringByDeletingLastPathComponent.c4slice, &err)) {
+            CBLWarnError(WebSocket, @"Cannot save cookie %d/%d", err.domain, err.code);
+            // TODO: any more handling?
+        }
+        
+        // update the response headers! do we really need this now? since we are already saving to c4db?
```
pasin, line 476: We still need to break the merged set-cookies and I don't know if LiteCore still handles anything related the cookies. I think it's no harm to send the correct header to LiteCore.

Objective-C/Internal/Replicator/CBLWebSocket.mm
pasin: Look lie we save the cookie here so getting the cookies here seems to be inline.

Objective-C/Internal/Foundation+CBL.h
jayahariv: CBLStringBytes usage needs a declaration and usage.  Normally two lines, one variable declaration and then use it in c4 APIs. 
```
CBLStringBytes varSlice(platformString)
c4db_setCookies(..., varSlice, ...)
```

If we use this, its more easier to use
```
c4db_setCookies(..., platformString.c4Slice, ...)
```

Objective-C/Internal/Foundation+CBL.h
jayahariv: Each function will be separated with a single line, including last function. One line and then `@end`

Objective-C/Internal/Foundation+CBL.h
pasin: CBLStringBytes has also some optimization. I don't think we should have two ways of converting the NString <-> Slice if possible. Otherwise we will have a question to ourselves which one to use and why.

Objective-C/Internal/Foundation+CBL.mm
jayahariv: will update. ` sizeof(self.UTF8String)`

Objective-C/Internal/Replicator/CBLWebSocket.mm
jayahariv: https://issues.couchbase.com/browse/CBL-681

Yes we have an open issue/

Objective-C/Internal/Foundation+CBL.h
jayahariv: Ok. let me check, if i can consume CBLStringBytes, so we can achieve the optimization. I will anyways remove from this PR and create a separate PR

Objective-C/Internal/CBLCookie.h
jayahariv: We don't have any unit test which tests this. so will go ahead and remove this. since this doesn't add much value.

jayahariv: I used it to validate a unit test which used proxy app to inject cookie.

Objective-C/Internal/Replicator/CBLWebSocket.mm
```diff
@@ -483,7 +474,17 @@ - (void) receivedHTTPResponse: (CFHTTPMessageRef)httpResponse {
...
-                [cookies addObject: [cookie substringWithRange: [match range]]];
+                NSString* cookieStr = [cookie substringWithRange: [match range]];
+                [cookies addObject: cookieStr];
+                
+                // save to lite core
+                C4Error err = {};
+                CBLStringBytes cookieSlice(cookieStr);
+                CBLStringBytes host(_remoteURL.host);
+                CBLStringBytes path(_remoteURL.path.stringByDeletingLastPathComponent);
+                if (!c4db_setCookie(_c4db, cookieSlice, host, path, &err)) {
```
pasin, line 485: Can you confirm here too if the db needs to be locked when calling the function?

Objective-C/Internal/Replicator/CBLWebSocket.mm
```diff
@@ -374,10 +381,24 @@ - (void) _sendWebSocketRequest {
...
-    slice cookies = _options[kC4ReplicatorOptionCookies].asString();
-    if (cookies)
-        [_logic addValue: cookies.asNSString() forHTTPHeaderField: @"Cookie"];
-
+    
+    slice sessionCookie = _options[kC4ReplicatorOptionCookies].asString();
+    if (sessionCookie.buf)
+        [_logic addValue: sessionCookie.asNSString()  forHTTPHeaderField: @"Cookie"];
+    
+    CBL_LOCK(_db) {
```
pasin, line 389: Is this confirmed that we need to lock? If so, can you add an internal method to set cookie in the CBLDatabase class so that we don't try to lock the db outside the CBLDatabase?

Objective-C/Internal/Replicator/CBLWebSocket.mm
```diff
@@ -443,13 +464,40 @@ - (void) receivedProxyHTTPResponse: (CFHTTPMessageRef)httpResponse {
...
+                NSString* cookieStr = [cookie substringWithRange: [match range]];
+                [cookies addObject: cookieStr];
+                
+                // save to lite core
+                CBL_LOCK(_db) {
+                    C4Error err = {};
+                    CBLStringBytes cookieSlice(cookieStr);
+                    CBLStringBytes host(_remoteURL.host);
+                    CBLStringBytes path(_remoteURL.path.stringByDeletingLastPathComponent);
+                    if (!c4db_setCookie(_db.c4db, cookieSlice, host, path, &err)) {
```
pasin, line 488: Can you check if this requires lock?

Objective-C/Internal/Replicator/CBLWebSocket.mm
jayahariv: Yes. confirmed by Jens. Will move the logic

Objective-C/Internal/Replicator/CBLWebSocket.mm
jayahariv: both set and get are added to db lock. since both has db-access

PR #2689, opened 2020-06-24
CBD-3489

mhocouchbase:
allow test-only.  speed up the buld cycle by running test separately.
clean up of DerivedData will be handled on jenkins jobs.  it is
necessary so that test & build can be run in parallel.

-Ming Ho

### DIFF
--- a/Scripts/generate_release_zip.sh
+++ b/Scripts/generate_release_zip.sh
@@ -4,7 +4,7 @@ set -e
 
 function usage
 {
-  echo "Usage: ${0} -o <Output Directory> [-v <Version (<Version Number>[-<Build Number>])>] [--xcframework] [--EE] [--notest] [--nocov] [--pretty]"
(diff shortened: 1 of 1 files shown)


Scripts/generate_release_zip.sh
```diff
@@ -79,11 +82,6 @@ else
   XCPRETTY="xcpretty"
 fi
 
-# Clean DerivedData
```
jayahariv, line 82: why remove this? is this under the assumption, Jenkins clears it whenever we ran this script. if so, we might need to run this outside Jenkins too and no harm in clearing it again?

mhocouchbase: The idea to allow jenkins to run test-only & build-only in parallel.  having "rm DerivedData" will cause failure since either job might have written files in it.  I have a remove step in jenkins job to clean up DerivedData and /private/tmp after the build is done so that the next build will start with a clean env.

PR #2692, opened 2020-06-26
Use XCTSkip: Cleaner code

jayahariv:
* available from XCode 11.4; so it is compatible with latest XCode

### DIFF
--- a/Objective-C/Tests/CBLTestCase.h
+++ b/Objective-C/Tests/CBLTestCase.h
@@ -160,6 +160,7 @@ NS_ASSUME_NONNULL_BEGIN
 - (void) mayHaveException: (NSString*)name in: (void (^) (void))block;
 
 - (void) ignoreException: (void (^) (void))block;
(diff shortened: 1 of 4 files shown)


jayahariv:
Yes, its related to the PR. XCTSkip functions are throwing exception when it the condition is success(probably, uses exception throw to make it work?)

pasin:
Can we wrap ignore exception when calling XCTSkipUnless?

jayahariv:
that will not skip the test. instead continue and fail.

pasin:
If that is the case, using XCTSkip could be annoyed when running tests from XCode even though the code is cleaner.

pasin:
But it's puzzle to me that ignore exception doesn't work because it is just doing this:

```
+ (void) allowExceptionIn: (void (^)(void))block {
    ++gC4ExpectExceptions;
    block();
    --gC4ExpectExceptions;
}
```

jayahariv:
Included ignore exception only method, without @try@catch

PR #2693, opened 2020-06-26
-no-serialize-debugging-options

jayahariv:
* enable this to remove all absolute path from binary & dsym
* https://bugs.swift.org/browse/SR-12783?focusedCommentId=56548&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-56548

* But the issue(CBL-896, Couldn't IRGen expression) we encounter with XCode 11.x is not present in XCode 12 Beta. 
* i have tested only on my machine. probably needs to test this again, once the Jenkins job builds it.


jayahariv:
I couldn't find any doc or articles explaining when this is added, but a commit inside to Swift Package Manager. Which is recent, but we can replace this with `OTHER_SWIFT_FLAGS = -Xfrontend -no-serialize-debugging-options`, which seems to be existed earlier.

pasin:
What is Xfrontend?

jayahariv:
-Xfrontend is a 2 part option which passes along the option that follows to the front end swift compiler. In this case the option, `-no-serialize-debugging-options`

PR #2694, opened 2020-06-26
Synchronize sending message in CBLMockConnection

pasin:
* The message could be sent out-of-order and that causes the tests to failed intermittently due to the error such as "CouchbaseLite Network ERROR: {BLIPIO#5489} Caught exception handling incoming BLIP message: BLIP protocol error: Bad incoming REQ #2 (too high)".
* Synchronizing the sending message to prevent message out-of-order.

CBL-1092

### DIFF
--- a/Objective-C/Tests/Util/CBLMockConnection.m
+++ b/Objective-C/Tests/Util/CBLMockConnection.m
@@ -94,6 +94,7 @@
 }
 
 - (void) send: (CBLMessage*)message completion: (void (^)(BOOL success, CBLMessagingError* _Nullable))completion {
+    // Synchronize to prevent message getting sent out-of-order:
     @synchronized (self) {
         NSLog(@"%@: Sending message ...", self);
         CBLMessagingError* error;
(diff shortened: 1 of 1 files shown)


Objective-C/Tests/Util/CBLMockConnection.m
```diff
@@ -94,18 +94,20 @@ - (void) open: (id<CBLReplicatorConnection>)connection completion: (void (^)(BOO
...
-    NSLog(@"%@: Sending message ...", self);
-    CBLMessagingError* error;
-    if(self.isClient && [self.errorLogic shouldCloseAtLocation: kCBLMockConnectionSend]) {
-        error = [self.errorLogic createError];
-        NSLog(@"%@: Send message failed with error : %@", self, error);
-        [self connectionBroken: error];
-    } else {
-        [self performWrite: [message toData]];
-        NSLog(@"%@: Send message completed", self);
+    @synchronized (self) {
```
jayahariv, line 97: should we add a one liner comment describing same?

PR #2696, opened 2020-06-30
CBL-935: Fix Jenkins PR Validation

jayahariv:
* fix Jenkins PR validation
* also fix couple of travisCI failures. But not CBL Swift([CBL-1109](https://issues.couchbase.com/browse/CBL-1109)) is failing on travis, which needs to be taken care separately. 
* remove some dead file reference from project.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -13,7 +13,7 @@ pipeline {
                     mv !(tmp) tmp
 
 		    # clone and update submodules here
-                    git clone https://github.com/couchbaselabs/couchbase-lite-ios-ee.git --branch $CHANGE_TARGET
+                    git clone git@github.com:couchbaselabs/couchbase-lite-ios-ee.git --branch $CHANGE_TARGET
 		    # submodule update inside lite-ios
 		    pushd couchbase-lite-ios-ee
                     git submodule update --init --recursive
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -12,11 +12,16 @@ pipeline {
 		    mkdir tmp
                     mv !(tmp) tmp
 
-		    # clone and update submodules here
-                    git clone git@github.com:couchbaselabs/couchbase-lite-ios-ee.git --branch $CHANGE_TARGET
-		    # submodule update inside lite-ios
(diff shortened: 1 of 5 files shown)


Jenkinsfile
```diff
@@ -12,11 +12,16 @@ pipeline {
 		    mkdir tmp
                     mv !(tmp) tmp
 
-		    # clone and update submodules here
-                    git clone https://github.com/couchbaselabs/couchbase-lite-ios-ee.git --branch $CHANGE_TARGET
-		    # submodule update inside lite-ios
+		    # Sometimes a PR depends on a PR in the EE repo as well. This needs to be convention based, so if there is a branch with the name PR-###
```
jayahariv, line 15:
```suggestion
		    # Sometimes the PR depends on a PR in the EE repo as well. This needs to be convention based, so if there is a branch with the name PR-###
```

Jenkinsfile
```diff
@@ -12,11 +12,16 @@ pipeline {
...
+		    # Sometimes the PR depends on a PR in the EE repo as well. This needs to be convention based, so if there is a branch with the name PR-###
+                    # (with the GH PR number) in the EE repo then use that, otherwise use the name of the target branch (master, release/XXX etc) 
+		    # clone the EE-repo
+                    git clone git@github.com:couchbaselabs/couchbase-lite-ios-ee.git --branch $BRANCH_NAME || \
+		      git clone git@github.com:couchbaselabs/couchbase-lite-ios-ee.git --branch $CHANGE_TARGET
+
+		    # clone the core-EE
 		    pushd couchbase-lite-ios-ee
-                    git submodule update --init --recursive
+		    git clone git@github.com:couchbase/couchbase-lite-core-EE.git --branch $BRANCH_NAME || \
```
pasin, line 23: This is hard to achieve as CBL-iOS doesn't control lite-core-EE's branch. If we want to update lite-core-EE, can we update the commit to the ios-ee's branch?

jayahariv: not sure, how we can achieve that, even possible?. Since we need some intelligence in choosing it, other than some manual intervention. Here `lite-ios-EE` and `lite-core-EE` uses the same logic. If there is a branch called, `pr-xx`, it will pick up, else go with master.

PR #2699, opened 2020-06-30
CBL-1109: fix CBL_Swift target failure

jayahariv:
* Due to this, `CBL_Swift` target was failing. 
* It was failing locally when built CE separately. 
* Not sure, why the failure is not happening in Jenkins slaves. probably CE only source causes the failure?

### DIFF
--- a/.slather.yml
+++ b/.slather.yml
@@ -2,7 +2,7 @@
 
 coverage_service: coveralls
 xcodeproj: CouchbaseLite.xcodeproj
-scheme: "CBL ObjC"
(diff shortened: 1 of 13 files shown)


jayahariv:
Seeing a valid failure in TravisCI, but not on Jenkins. Even the local EE source project succeeds, but fails only on CE source. 
```
Testing failed:

	'id' is unavailable in Swift: 'id' is not available in Swift; use 'Any'

	Testing cancelled because the build failed.

** TEST FAILED **
```

PR #2700, opened 2020-06-30
Implement Replicator.serverCertificate property

pasin:
* Implement Replicator.serverCertificate property that has the value updated from CBLWebSocket.
* Reset the serverCertificate when starting the replicator.
* For Objective-C, add the note the API doc that the serverCertificate should be explictly retained / released as the CF Object doesn’t get ARC by default.

CBL-1095

### DIFF
--- a/Objective-C/Internal/Replicator/CBLWebSocket.mm
+++ b/Objective-C/Internal/Replicator/CBLWebSocket.mm
@@ -90,9 +90,11 @@ struct PendingWrite {
     BOOL _connectingToProxy;
     BOOL _connectedThruProxy;
     
-    NSArray* _clientIdentity;
+    CBLReplicator* _replicator;
     CBLDatabase* _db;
     NSURL* _remoteURL;
+    
+    NSArray* _clientIdentity;
     
 #ifdef COUCHBASE_ENTERPRISE
     BOOL _acceptOnlySelfSignedCert;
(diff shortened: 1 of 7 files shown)


Objective-C/Internal/Replicator/CBLWebSocket.mm
```diff
@@ -757,6 +775,10 @@ - (void)stream: (NSStream*)stream handleEvent: (NSStreamEvent)eventCode {
             break;
         case NSStreamEventErrorOccurred:
             CBLLogVerbose(WebSocket, @"%@: ErrorEncountered on %@", self, stream);
+            if (_checkSSLCert) {
+                SecTrustRef trust = [self copyTrustFromReadStream];
+                [self updateServerCertificateFromTrust: (SecTrustRef)CFAutorelease(trust)];
```
jayahariv, line 780: why not release after passing to replicator, instead of auto release?

Objective-C/Tests/URLEndpointListenerTest.m
```diff
@@ -893,4 +893,114 @@ - (void) testCloseWithActiveListener {
...
+    Assert(replicator.serverCertificate == NULL);
+    
+    [replicator start];
+    
+    [self waitForExpectations: @[x1] timeout: timeout];
+    Assert(replicator.serverCertificate != NULL);
+    
+    SecCertificateRef serverCert = (__bridge SecCertificateRef) listener.tlsIdentity.certs[0];
+    [self checkEqualForCert: serverCert andCert: replicator.serverCertificate];
+
```
jayahariv, line 990: make sure, same replicator can connect after the failure?

pasin: Done

PR #2702, opened 2020-07-02
Refact: clean code XCTSkip 

jayahariv:
* keeps forgetting, whether the test skips or really passing. 
* XCTSkip can be really helpful in detecting this case

### DIFF
--- a/Objective-C/Tests/URLEndpointListenerTest.m
+++ b/Objective-C/Tests/URLEndpointListenerTest.m
@@ -316,7 +316,9 @@ typedef CBLURLEndpointListener Listener;
 }
 
 - (void) testTLSIdentity {
-    if (!self.keyChainAccessAllowed) return;
+    [self ignoreExceptionBreakPointOnly: ^{
+        XCTSkipUnless(self.keyChainAccessAllowed, @"Skipping, keychain not accessible in this config");
+    }];
     
     // Disabled TLS:
     Config* config = [[Config alloc] initWithDatabase: self.otherDB];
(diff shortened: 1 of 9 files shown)


pasin:
This feature doesn't support by XCode 10 so we can't build the test to run on AWS device farm. Can this be hold off?

Swift/Tests/TLSIdentityTest.swift
```diff
@@ -149,7 +149,7 @@ class TLSIdentityTest: CBLTestCase {
     }
     
     func testCreateGetDeleteServerIdentity() throws {
-        if (!keyChainAccessAllowed) { return }
+        try XCTSkipUnless(keyChainAccessAllowed, "Skipping, keychain is not allowed in this config")
```
pasin, line 152: Don't we need to ignore exception here?

Objective-C/Tests/URLEndpointListenerTest.m
```diff
@@ -873,7 +901,9 @@ - (void) _testReadOnlyListener {
 }
 
 - (void) testCloseWithActiveListener {
-    if (!self.keyChainAccessAllowed) return;
+    [self ignoreExceptionBreakPointOnly: ^{
+        XCTSkipUnless(self.keyChainAccessAllowed, @"Skipping, keychain not accessible in this config");
+    }];
```
pasin, line 906: Can we extract this into a method to make this cleaner?

Swift/Tests/TLSIdentityTest.swift
jayahariv: Exception pointer is only hit on ObjC, and not in Swift.

pasin: I don't know. Can you wrap this in a function such as checkKeyChainAccess()?

jayahariv:
once device farm is ready, please merge.

PR #2705, opened 2020-07-03
stoppable protocol

jayahariv:
* listener, replicator, livequery can all implements stoppable protocol.
* so we can avoid duplication of code in case of more stoppable instances, like message-endpoint-listener(CBL-950).

### DIFF
--- a/Objective-C/CBLDatabase.mm
+++ b/Objective-C/CBLDatabase.mm
@@ -64,11 +64,7 @@ using namespace fleece;
     BOOL _shellMode;
     dispatch_source_t _docExpiryTimer;
     
-    NSMutableSet<CBLReplicator*>* _activeReplicators;
-    NSMutableSet<CBLLiveQuery*>* _activeLiveQueries;
-#ifdef COUCHBASE_ENTERPRISE
-    NSMutableSet<CBLURLEndpointListener*>* _activeListeners API_AVAILABLE(macos(10.12), ios(10.0));
-#endif
+    NSMutableSet<id<CBLStoppable>>* _activeStoppables;
     
     BOOL _isClosing;
     NSCondition* _closeCondition;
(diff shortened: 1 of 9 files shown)


Objective-C/Internal/Replicator/CBLStoppable.h
```diff
@@ -0,0 +1,31 @@
...
+//  limitations under the License.
+//
+
+#import <Foundation/Foundation.h>
+
+NS_ASSUME_NONNULL_BEGIN
+
+@protocol CBLStoppable <NSObject>
+
+- (void) _stop;
```
pasin, line 26: Is there a reason that this cannot be `stop`?

Objective-C/CBLDatabase.mm
```diff
@@ -1087,90 +1060,33 @@ - (BOOL) saveDocument: (CBLDocument*)document
...
-    CBL_LOCK(self) {
-        return _activeLiveQueries.count;
-    }
-}
-
-#pragma mark - REPLICATOR
-
-- (void) addActiveReplicator: (CBLReplicator*)replicator {
+#pragma mark - Stoppable
+- (void) addActiveStoppable: (id<CBLStoppable>)stoppable {
```
pasin, line 1064: Empty line between pragma and method.

Objective-C/Internal/Replicator/CBLStoppable.h
jayahariv: public interface declaration will be an extra declaration(only for user reference). Not sure, whether that cause any warnings/compile issue.

pasin: I think it shouldn't cause the compilation problem. If there is no logic problem to use stop, I think we can just use stop.

jayahariv: that will make code more cleaner, the overhead of calling `_stop` from `stop`. let me try that.

PR #2711, opened 2020-07-08
cbl-1094: update the server verfication mode

jayahariv:
* remove the enum and use a bool indicating whether only self-signed is accepted
* for every blip request, we look for cookies, if no cookies saved, we only print warning and not an error.

### DIFF
--- a/Objective-C/CBLReplicatorConfiguration.m
+++ b/Objective-C/CBLReplicatorConfiguration.m
@@ -44,7 +44,7 @@
 @synthesize conflictResolver=_conflictResolver;
 
 #ifdef COUCHBASE_ENTERPRISE
-@synthesize serverCertificateVerificationMode=_serverCertificateVerificationMode;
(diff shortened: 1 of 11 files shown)


Objective-C/CBLReplicatorConfiguration.m
```diff
@@ -205,6 +205,8 @@ - (NSDictionary*) effectiveOptions {
     NSString* uniqueID = $castIf(CBLMessageEndpoint, _target).uid;
     if (uniqueID)
         options[@kC4ReplicatorOptionRemoteDBUniqueID] = uniqueID;
+    
+    options[@kC4ReplicatorOptionOnlySelfSignedServerCert] = @(_acceptOnlySelfSignedServerCertificate);
```
jayahariv, line 209: pass the acceptOnly flag to the lite core

pasin: See my comment.

PR #2719, opened 2020-07-31
Fix client cert auth error

pasin:
* Fixed the error code for client cert auth error
* Added unit tests for testing failed cases for client cert auth.
* Shorten the test names
* Fixed crash check checking error in test.

CBL-1177


pasin:
PR validation failed is not related to PR.

PR #2732, opened 2020-08-27
CBL-1216: fix: check db close before creating the c4replicator

jayahariv:


Objective-C/CBLReplicator.mm
```diff
@@ -251,12 +251,17 @@ - (bool) _setupC4Replicator: (C4Error*)outErr {
...
     CBL_LOCK(_config.database) {
+        [_config.database mustBeOpenLocked];
+        
         if (remoteURL || !otherDB)
             _repl = c4repl_new(_config.database.c4db, addr, dbName, params, outErr);
         else  {
 #ifdef COUCHBASE_ENTERPRISE
-            if (otherDB)
+            if (otherDB) {
+                [otherDB mustBeOpenLocked];
```
pasin, line 261: This doesn't prevent the other thread to close the otherDB when creating the replicator. I think otherDB needs to be locked as well

Sandychuang8: question, isn't Jay already call the mustBeOpenLocked from otherDB?

jayahariv: after this line, it can get closed, before replicator is being started (in `c4repl_newLocal`)

PR #2739, opened 2020-09-10
skip local network in unit test

jayahariv:
* update test to ignore the local network, since the iphone simulator might pick up hostname instead of the localhost name


pasin (approved):

pasin:
Only ObjC?

PR #2741, opened 2020-09-11
CBL-1111 : Fix value type has no subscripts issue when building with XCode12 Beta

pasin:
* Cast to the Objective-C Protocol type that defines the subscript method.
* The fix is harmless and works with XCode 11.7.


pasin:
Note: PR is set to merge to release/hydrogen branch.

jayahariv (approved):

PR #2746, opened 2020-09-17
test fix: bonjour service name add to InfoPlist

jayahariv:
* include bonjour servcie name in info plist to grant permission  else will return NSNetServicesErrorCode  -72008
* only impacts the unit tests. 
* needs to cherry pick to Hydrogen branch too.


pasin (approved):

jayahariv:
probably i need to add the same to other iOS App targets, will update  the PR soon.

PR #2751, opened 2020-09-23
Update build script 

jayahariv:
* Update build script to generate package.swift with checksum in the output directory
* CBL-1313: Fix the issue when creating xcframework alone, it was tring to generate objc, swift documentation which end up failing.

### DIFF
--- a/Scripts/generate_package_manifest.sh
+++ b/Scripts/generate_package_manifest.sh
@@ -1,0 +1,77 @@
+#!/bin/sh
+
+set -e
+
+function usage
+{
+  echo "Usage: ${0} -zip-path <path to xcframework zip> -o <output file path> --EE"
+}
+
+while [[ $# -gt 0 ]]
+do
+  key=${1}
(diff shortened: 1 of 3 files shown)


Scripts/generate_package_manifest.sh
```diff
@@ -0,0 +1,77 @@
...
+      ;;
+  esac
+  shift
+done
+
+if [ -z "$EE" ]
+then
+  PRODUCT_NAME="CouchbaseLiteSwift"
+else
+  PRODUCT_NAME="CouchbaseLiteSwiftEE"
```
pasin, line 41: Is this will be referenced by users when they use Swift Package Manager? Does it need to be different b/w CE and EE? Does the binary file need to be CouchbaseLiteSwiftEE.xx?

jayahariv: Right. I think we can keep it as same.

pasin (approved):

PR #2752, opened 2020-09-30
test: add null listener authenticator tests

jayahariv:
* with unit tests, refactored the code to reuse the tlsIdentity creation and cleanup methods
* renamed the previous test : `testPasswordAuth` to `testNonTLSPasswordListenerAuthenticator`

### DIFF
--- a/Objective-C/Tests/URLEndpointListenerTest.m
+++ b/Objective-C/Tests/URLEndpointListenerTest.m
@@ -269,8 +269,8 @@ typedef CBLURLEndpointListener Listener;
     return identity;
 }
 
-- (BOOL) cleanupTLSIdentity: (BOOL)isServer {
-    if (!self.keyChainAccessAllowed) return nil;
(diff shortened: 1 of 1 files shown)


Objective-C/Tests/URLEndpointListenerTest.m
```diff
@@ -249,6 +249,34 @@ - (void) checkEqualForCert: (SecCertificateRef)cert1 andCert: (SecCertificateRef
...
+                                                            expiration: nil
+                                                                 label: label
+                                                                 error: &err];
+    AssertNotNil(identity);
+    AssertNil(err);
+    return identity;
+}
+
+- (BOOL) cleanupTLSIdentity: (BOOL)isServer {
+    if (!self.keyChainAccessAllowed) return nil;
```
pasin, line 273: Return false? or No need to return if the returned value is not used.

pasin (approved):

PR #2763, opened 2020-10-27
fix: stack memory usage after release

jayahariv:
* config directory setting inside the static was causing the already released directory string to be used when c4config is used second time

* more details of the error : CBL-1395

### DIFF
--- a/Objective-C/CBLDatabase.mm
+++ b/Objective-C/CBLDatabase.mm
@@ -493,6 +493,9 @@ static void dbObserverCallback(C4DatabaseObserver* obs, void* context) {
     slice fromPath(path.fileSystemRepresentation);
     CBLStringBytes destinationName(name);
     C4DatabaseConfig2 c4Config = c4DatabaseConfig2(config ?: [CBLDatabaseConfiguration new]);
+    CBLStringBytes dir(config.directory);
+    c4config.parentDirectory = dir;
+    
     if (!(c4db_copyNamed(fromPath, destinationName, &c4Config, &err) || err.code==0 || convertError(err, outError))) {
         NSString* toPathStr = databasePath(name, dir);
         NSError* removeError;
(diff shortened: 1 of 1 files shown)


jayahariv:
These failures are certificate related, which needs to be looked at after this.

Objective-C/CBLDatabase.mm
```diff
@@ -493,6 +493,9 @@ + (BOOL) copyFromPath: (NSString*)path
     slice fromPath(path.fileSystemRepresentation);
     CBLStringBytes destinationName(name);
     C4DatabaseConfig2 c4Config = c4DatabaseConfig2(config ?: [CBLDatabaseConfiguration new]);
+    CBLStringBytes d(config.directory);
+    c4Config.parentDirectory = d;
+
```
pasin, line 498: I'm confused with the behavior (I feel that maybe CBLStringBytes is the source of the issue?).  If later trying to get c4Config from the c4database, will the parentDirectory still **always** be valid?

pasin: I have just looked at CBLStringBytes class and the code here. I think the problem happens when the string is stored in the _local array (the other case shouldn't cause the issue as the live of the string's data is bound to the live of the config.directory string.  The problem here is that the struct will be copied when the struct is returned. The original struct will be deleted and the _local array will be freed as well. CBLStringBytes would need a copy constructor that copies the _local array.

pasin (changes requested):

Objective-C/CBLDatabase.mm
jayahariv: I can think of one scenario, where static function does reserve the c4Config.parentDirectory string when its first used. Later on when second c4Config with different directory is initialized it will try to use the first directory string.

pasin: I think the real issue is in `CBLStringBytes` when copying the CBLStringBytes that has the raw string value in _local[]. Implementing a copy constructor/assignment in `CBLStringBytes` to properly _local[] would fix the issue here.

pasin (changes requested):

Objective-C/CBLDatabase.mm
jayahariv: ok. let me try it.

pasin: I was wrong about suggesting to provide a copy constructor to copy the _local array. In general, the array should be copied by default. The real problem here that prevent _local to be copied because we have assign the CBLStringBytes to a slice variable which will cause the slice() operation to be used hence no the copying of CBLStringBytes.

I couldn't think a good way around this and I'm wondering how much performance would we get out of the getBytes() logic as I expect that CFStringGetCStringPtr would normally works.

pasin (approved):

PR #2768, opened 2020-11-10
fix: hydrogen - coverage data not found issue 

jayahariv:
* slather was using wrong scheme name
* also, without cleaning and testing wsa resulting in similar coverage data not found issue
* code cleanup(single line to multi line)

http://mobile.jenkins.couchbase.com/job/couchbase-lite-ios-edition-test/266/

### DIFF
--- a/Scripts/generate_release_zip.sh
+++ b/Scripts/generate_release_zip.sh
@@ -124,7 +124,7 @@ then
   echo "Run ObjC iOS tests ..."
   # iOS-App target runs Keychain-Accessing tests
   sh Scripts/xctest_crash_log.sh --delete-all
-  xcodebuild test \
+  xcodebuild clean test \
     -project CouchbaseLite.xcodeproj \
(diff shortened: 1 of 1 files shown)


Scripts/generate_release_zip.sh
```diff
@@ -124,7 +124,7 @@ then
   echo "Run ObjC iOS tests ..."
   # iOS-App target runs Keychain-Accessing tests
   sh Scripts/xctest_crash_log.sh --delete-all
-  xcodebuild test \
+  xcodebuild clean test \
```
pasin, line 127: Is the `clean` parameter needed for iOS but not macOS?

pasin (approved):

Scripts/generate_release_zip.sh
jayahariv: I thought, code coverage data is only taken for iOS. The last ran test results from xcodebuild.

PR #2790, opened 2020-12-19
make catalyst compatible

jayahariv:
* fix the deprecation warnings with catalyst
* make sdkroot=iphoneos

### DIFF
--- a/Objective-C/Internal/Replicator/CBLTrustCheck.mm
+++ b/Objective-C/Internal/Replicator/CBLTrustCheck.mm
@@ -115,7 +115,11 @@ static BOOL sOnlyTrustAnchorCerts;
 #if TARGET_OS_MACCATALYST
     if (@available(iOS 12.0, macos 10.14, *)) {
         CFErrorRef error;
-        Assert(SecTrustEvaluateWithError(_trust, &error));
+        BOOL trusted = SecTrustEvaluateWithError(_trust, &error);
+        if (!trusted)
+            CBLWarnError(Sync, @"Failed to force trust");
+        
+        return trusted;
     } else {
         CBLWarnError(Sync, @"Catalyst version not available: Not supported by macOS < 10.14 and iOS < 12.0");
     }
@@ -156,11 +160,7 @@ static BOOL sOnlyTrustAnchorCerts;
     err = SecTrustEvaluate(_trust, &result);
 #endif
     
-    if (err
-#if TARGET_OS_MACCATALYST
-        || !trusted
-#endif
-        ) {
+    if (err) {
         CBLWarn(Default, @"%@: SecTrustEvaluate failed with err %d", self, (int)err);
         MYReturnError(outError, err, NSOSStatusErrorDomain, @"Error evaluating certificate");
         return nil;
--- a/Objective-C/Internal/Replicator/CBLTrustCheck.mm
+++ b/Objective-C/Internal/Replicator/CBLTrustCheck.mm
@@ -147,10 +147,9 @@ static BOOL sOnlyTrustAnchorCerts;
     SecTrustResultType result;
     OSStatus err;
 #if TARGET_OS_MACCATALYST
-    BOOL trusted = NO;
-    CFErrorRef cfError = nullptr;
     if (@available(iOS 12.0, macos 10.14, *)) {
-        trusted = SecTrustEvaluateWithError(_trust, &cfError);
+        if (!SecTrustEvaluateWithError(_trust, nullptr))
+            CBLLogVerbose(Sync, @"SecTrustEvaluateWithError failed! Evaluating trust result...");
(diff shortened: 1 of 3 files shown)


Objective-C/Internal/Replicator/CBLTrustCheck.mm
```diff
@@ -110,10 +110,20 @@ - (BOOL) forceTrusted {
...
+        CFErrorRef error;
+        Assert(SecTrustEvaluateWithError(_trust, &error));
+    } else {
+        CBLWarnError(Sync, @"Catalyst version not available: Not supported by macOS < 10.14 and iOS < 12.0");
+    }
+#else
+    SecTrustResultType result;
     SecTrustEvaluate(_trust, &result);
+#endif
+
```
jayahariv, line 126:
```suggestion
```

Objective-C/Internal/Replicator/CBLTrustCheck.mm
```diff
@@ -110,10 +110,19 @@ - (BOOL) forceTrusted {
     CFDataRef exception = SecTrustCopyExceptions(_trust);
     if (!exception)
         return NO;
-    SecTrustResultType result;
     SecTrustSetExceptions(_trust, exception);
     CFRelease(exception);
+#if TARGET_OS_MACCATALYST
+    if (@available(iOS 12.0, macos 10.14, *)) {
+        CFErrorRef error;
+        Assert(SecTrustEvaluateWithError(_trust, &error));
```
pasin, line 118: I think it is better to return NO and do warning log if this is failed.

Objective-C/Internal/Replicator/CBLTrustCheck.mm
```diff
@@ -132,8 +141,26 @@ - (NSURLCredential*) checkTrust: (NSError**)outError {
...
+        CBLWarnError(Sync, @"SecTrustEvaluateWithError Catalyst version not available: Not supported by macOS < 10.14 and iOS < 12.0");
+        return nil;
+    }
+#else
+    err = SecTrustEvaluate(_trust, &result);
+#endif
+    
+    if (err
+#if TARGET_OS_MACCATALYST
+        || !trusted
```
pasin, line 161: Is it possible to clean this up? Also if trusted is false, will there be always an err?

xcconfigs/Project.xcconfig
```diff
@@ -61,6 +61,9 @@ SWIFT_VERSION                      = 5.0
 VERSION_INFO_PREFIX                = 
 VERSIONING_SYSTEM                  = apple-generic
 
+// Catalyst
+SDKROOT                            = iphoneos
```
pasin, line 65: I don't have info on this and will this be OK with the regular mac?

pasin:
You might need to do the catalyst fix for this as well :
https://github.com/couchbase/couchbase-lite-core/blob/master/Crypto/PublicKey%2BApple.mm

jayahariv:
LiteCore PR:
https://github.com/couchbase/couchbase-lite-core/pull/1097

Objective-C/Internal/Replicator/CBLTrustCheck.mm
jayahariv: As per the document, yes. Will clean it with that.

xcconfigs/Project.xcconfig
jayahariv: I have tested by creating a release package and used `.framework` inside a sample mac app.

Objective-C/Internal/Replicator/CBLTrustCheck.mm
```diff
@@ -132,7 +145,21 @@ - (NSURLCredential*) checkTrust: (NSError**)outError {
     
     // Evaluate trust:
     SecTrustResultType result;
-    OSStatus err = SecTrustEvaluate(_trust, &result);
+    OSStatus err;
+#if TARGET_OS_MACCATALYST
+    BOOL trusted = NO;
+    CFErrorRef cfError = nullptr;
+    if (@available(iOS 12.0, macos 10.14, *)) {
+        trusted = SecTrustEvaluateWithError(_trust, &cfError);
```
pasin, line 153: If we don't care about the result here, can we don't get the trusted boolean value and use NULL for the error?

pasin (changes requested):

Objective-C/Internal/Replicator/CBLTrustCheck.mm
jayahariv: ok

jayahariv: Return value('trusted') cannot be ignored, it will generate error . So adding a verbose log specifying the failure and evaluating the trust result. 
removed the cferror.

pasin (approved):

PR #2795, opened 2021-01-20
test: fix race

jayahariv:
* since the concurrent conflict queue is executing the blocks. 'order.count' can change in between which ?might skip the thread sleep


Swift/Tests/ReplicatorTest+CustomConflict.swift
```diff
@@ -632,10 +632,16 @@ class ReplicatorTest_CustomConflict: ReplicatorTest {
         let config = getConfig(.pull)
         var resolver: TestConflictResolver!
         var order = [String]()
+        let lock = NSLock()
         resolver = TestConflictResolver() { (conflict) -> Document? in
+
```
jayahariv, line 637:
```suggestion
```

pasin (approved):

PR #2799, opened 2021-01-27
CBL-1610: 

jayahariv:
* x86_64 + macosx
* x86_64 + catalyst
* arm64 + macosx (when ran on arm64 machine, it will test for the arm64 destination. )
* arm64 + catalyst (same ^)

* swift + objc combination for above


pasin (approved):
So many variants now.

PR #2811, opened 2021-03-02
feat: save / get blob from db

jayahariv:


jayahariv:
PR job failed with improper cleaning of the build. 
https://issues.couchbase.com/browse/CBL-1717

Objective-C/CBLDatabase.mm
```diff
@@ -327,6 +334,35 @@ - (BOOL) purgeDocumentWithID: (NSString*)documentID error: (NSError**)error {
...
+    
+    //TODO: If data is large, can get the file path & memory-map it
+    FLSliceResult res = c4blob_getContents(blobStore, key, nullptr);
+    NSData* content = sliceResult2data(res);
+    FLSliceResult_Release(res);
+    
+    NSMutableDictionary *tempProps = [dict mutableCopy];
+    tempProps[kBlobDataProperty] = content;
+    tempProps[kBlobLengthProperty] = @(content.length);
+    return [[CBLBlob alloc] initWithDatabase: self properties: tempProps];
```
pasin, line 363: What needs to be passed into the properties is the given dict from the method's parameter.

Objective-C/CBLDatabase.mm
```diff
@@ -327,6 +334,35 @@ - (BOOL) purgeDocumentWithID: (NSString*)documentID error: (NSError**)error {
     }
 }
 
+#pragma mark - Blob Save/Get
+
+- (BOOL) saveBlob: (CBLBlob*)blob error: (NSError**)error {
+    return [blob installInDatabase: self error: error];
+}
+
+- (CBLBlob*) getBlob: (NSDictionary*)dict {
```
pasin, line 343: Discuss: Should this be getBlobWithProperties for Objective-C? What is your opinion?

jayahariv: `getBlob: (NSDictionary*)properties ` should be fine, argument placeholder makes it clear right?

Objective-C/CBLDatabase.mm
jayahariv: ok. this means, 
- here we will only pass the db instance and the props to Blob object. 
- inside the blob class constructor, we will get the content (move the logic to blob class)
- populates the content, content-length, content-type etc from blob constructor

Objective-C/CBLDatabase.mm
pasin: ok

Objective-C/CBLDatabase.mm
pasin: You can also make the init to the assertion to check the required properties (not sure if this will affect any existing functionalities or not).

jayahariv: assertion check will fail other tests, since the same function without the digest information also uses the same constructor. So not going to add the assertion check.

jayahariv:
will add Blake back, once everything is approved and ready

Objective-C/CBLBlob.mm
```diff
@@ -184,6 +189,11 @@ - (NSData*) content {
         FLSliceResult_Release(res);
         if (content && content.length <= kMaxCachedContentLength)
             _content = content;
+        
+        // update the content and length while loading it
+        [self updateProperties: kLengthMetaProperty value: @(_content.length)];
+        [self updateProperties: kDataMetaProperty value: _content];
```
pasin, line 195: We don't need to put _content into the dictionary. As we don't need to keep _properties, we will just update _length directly (make sure to synchronize).

Objective-C/CBLBlob.mm
```diff
@@ -121,9 +121,7 @@ - (instancetype) initWithDatabase: (CBLDatabase*)db
     self = [super init];
     if(self) {
         _db = db;
-        NSMutableDictionary *tempProps = [properties mutableCopy];
-        tempProps[kTypeMetaProperty] = nil;
-        _properties = tempProps;
+        _properties = properties;
```
pasin, line 124: It doesn't seem like it's worth to keep the _properties anymore as all of the properties have been read and stored in  variables.

Objective-C/CBLDatabase.mm
```diff
@@ -53,6 +55,12 @@
 // How long to wait after a database opens before expiring docs
 #define kHousekeepingDelayAfterOpening 3.0
 
+static NSString* kBlobTypeProperty = @kC4ObjectTypeProperty;
+static NSString* kBlobDigestProperty = @kC4BlobDigestProperty;
+static NSString* kBlobDataProperty = @kC4BlobDataProperty;
+static NSString* kBlobLengthProperty = @"length";
+static NSString* kBlobContentTypeProperty = @"content_type";
```
pasin, line 62: We should make these constants beside data public. I will update the API spec.

Objective-C/CBLBlob.mm
```diff
@@ -147,9 +151,10 @@ - (NSDictionary*) properties {
         return _properties;
     } else {
         // New blob:
-        return $dict({@"digest", _digest},
-                     {@"length", (_length ? @(_length) : nil)},
-                     {@"content_type", _contentType});
+        return $dict({kTypeMetaProperty, kBlobType},
```
pasin, line 154: To simplify, we don't keep the _properties and we always return a new dict.

pasin: This method is not thread safe as _length could be changed. Can you also synchronize this method?

Objective-C/CBLBlob.mm
```diff
@@ -206,6 +216,11 @@ - (NSData*) content {
         _initialContentStream = nil;
         _content = result;
         _length = _content.length;
+        
+        // update the content and length while loading it
+        [self updateProperties: kLengthMetaProperty value: @(_content.length)];
+        [self updateProperties: kDataMetaProperty value: _content];
+
```
pasin, line 223: Same as my comment above. We will need to make sure that updating _content and _length are thread-safe.

Objective-C/CBLDatabase.h
```diff
@@ -218,6 +219,37 @@ typedef NS_ENUM(uint32_t, CBLMaintenanceType) {
...
+ when compacting the database.
+ 
+ @param blob The blob to save.
+ @param error On return, the error if any.
+ @return /True on success, false on failure.
+ */
+- (BOOL) saveBlob: (CBLBlob*)blob error: (NSError**)error;
+
+/**
+ Get a blob object using a blob’s digest.
```
pasin, line 237: Change to `using blob's metadata`.

Objective-C/CBLDatabase.h
```diff
@@ -218,6 +219,37 @@ typedef NS_ENUM(uint32_t, CBLMaintenanceType) {
...
+ If the blob of the specified digest doesn’t exist, the nil value will be returned.
```
pasin, line 238: Change to `If the blob of the specified metadata doesn’t exist`,

Objective-C/CBLDatabase.mm
```diff
@@ -327,6 +335,28 @@ - (BOOL) purgeDocumentWithID: (NSString*)documentID error: (NSError**)error {
...
+    if (!properties[kBlobDigestProperty] ||
+        ![properties[kBlobDigestProperty] isKindOfClass: [NSString class]] ||
+        !properties[kBlobTypeProperty] || ![properties[kBlobTypeProperty] isEqual: @"blob"] ||
+        (properties[kBlobContentTypeProperty] &&
+         ![properties[kBlobContentTypeProperty] isKindOfClass: [NSString class]]) ||
+        (properties[kBlobLengthProperty] &&
+         ![properties[kBlobLengthProperty] isKindOfClass: [NSNumber class]])) {
+        
+        [NSException raise: NSInvalidArgumentException
+                    format: @"Property dictionary is missing the digest or @type"];
```
pasin, line 354: Could be "The given blob's metadata might be missing the digest / @type key or containing invalid values".

pasin (changes requested):

Objective-C/CBLBlob.mm
```diff
@@ -42,19 +42,20 @@
...
 
-static NSString* const kTypeMetaProperty = @kC4ObjectTypeProperty;
-static NSString* const kDigestMetaProperty = @kC4BlobDigestProperty;
-static NSString* const kDataMetaProperty = @kC4BlobDataProperty;
-static NSString* const kLengthMetaProperty = @"length";
-static NSString* const kContentTypeMetaProperty = @"content_type";
-static NSString* const kBlobType = @kC4ObjectType_Blob;
+NSString* const kCBLBlobType = @kC4ObjectType_Blob;
+NSString* const kCBLBlobTypeMetaProperty = @kC4ObjectTypeProperty;
+NSString* const kCBLBlobDigestMetaProperty = @kC4BlobDigestProperty;
```
pasin, line 47: Let's do to match CBL-C.

**Blob:**
kCBLBlobType
kCBLBlobDigestProperty
kCBLBlobContentTypeProperty
kCBLBlobLengthProperty

**Type**
kCBLTypeProperty

pasin (changes requested):
There are still some threading issues that I found but let's fix that in a separate issue (we can discuss).

Objective-C/Tests/DocumentTest.m
```diff
@@ -1588,6 +1598,64 @@ - (void)testBlob {
...
+    CBLBlob* retrivedBlob = [self.db getBlob: dict];
+    AssertEqual(retrivedBlob.properties.count, dict.count);
+    AssertEqualObjects(retrivedBlob.properties, dict);
+    
+    // access the content and see the content length gets populated
+    AssertEqualObjects(retrivedBlob.content, content);
+    AssertEqual([retrivedBlob.properties[kCBLBlobLengthProperty] unsignedIntValue], content.length);
+}
+
+- (void) testGteBlobUsingInvalidJSON {
```
jayahariv, line 1623:
```suggestion
- (void) testGetBlobUsingInvalidJSON {
```

PR #2812, opened 2021-03-03
feat: add toJSON function1

jayahariv:


jayahariv:
Related PR 
https://github.com/couchbaselabs/fleece/pull/86

Objective-C/CBLArray.mm
```diff
@@ -99,6 +100,15 @@ - (void) fl_encodeToFLEncoder: (FLEncoder)enc {
     abort();
 }
 
+#pragma mark - toJSON
+
+- (NSString*) toJSON {
+    JSONEncoder enc;
+    _array.encodeTo(enc);
+    auto data = enc.finish();
+    return slice2string(data);
```
pasin, line 109: In CBLDictionary's toJSON, the code is under _sharedLock. Can you check if the lock is needed?

pasin (commented):
There is also CBLNewDictionary that needs to implement toJSON() as well.

Objective-C/CBLBlob.mm
```diff
@@ -367,7 +367,7 @@ - (void) fl_encodeToFLEncoder: (FLEncoder)encoder {
         }
     }
     
-    NSDictionary* dict = self.jsonRepresentation;
+    NSDictionary* dict = FLEncoder_isFleece(encoder) ? self.jsonRepresentation : self.properties;
```
jayahariv, line 370: https://github.com/couchbaselabs/fleece/pull/87

pasin: I think we should take more direct approach here as it's impossible to understand when reading this code only why FleeceEncoder uses data from json representation and why JSONEncoder uses properties. I suggested that we create a new FLEncoderContext that can be used to put into the FLEncoder's extra info. FLEncoderContext could have two properties:
* Nullable document.
* boolean encodeQueryParameter (so that we really know what is going on here).

pasin: I think we should rename jsonRepresentation to something like

- (NSDictionary*) blobProperties: (BOOL)mayIncludeContent;

pasin (changes requested):

Objective-C/CBLBlob.mm
jayahariv: Updated the code with FLEncoderContext.

Objective-C/Tests/CBLTestCase.m
```diff
@@ -371,4 +371,47 @@ - (BOOL) isProfiling {
...
+    [doc setValue: [NSNull null] forKey: @"nullKey"];
+    
+    NSData* content = [@"i'm blob" dataUsingEncoding: NSUTF8StringEncoding];
+    CBLBlob* blob = [[CBLBlob alloc] initWithContentType:@"text/plain" data: content];
+    [doc setBlob: blob forKey: @"blob"];
+    
+    CBLMutableArray* array = [[CBLMutableArray alloc] init];
+    [array addValue: @1];
+    [array addValue: @2];
+    [doc setValue: array forKey: @"array"];
```
pasin, line 391: Can you add a Blob in here as well?

Objective-C/Tests/CBLTestCase.m
```diff
@@ -371,4 +371,47 @@ - (BOOL) isProfiling {
...
+    
+    CBLMutableDictionary* dict = [[CBLMutableDictionary alloc] init];
+    [dict setValue: @"CA" forKey: @"state"];
+    [doc setValue: dict forKey: @"dict"];
```
pasin, line 395: Can you add a Blob in here as well?

Objective-C/Tests/CBLTestCase.m
```diff
@@ -371,4 +371,47 @@ - (BOOL) isProfiling {
...
+    
+    return doc;
+}
+
+- (NSDictionary*) populatedMDocToJSONObj {
+    return @{@"array": @[@1,@2],
```
pasin, line 401: Can you add a Blob in here as well?

Objective-C/Tests/CBLTestCase.m
```diff
@@ -371,4 +371,47 @@ - (BOOL) isProfiling {
...
+             @"boolVal": @YES,
+             @"dateKey": @"1970-01-01T00:00:10.000Z",
+             @"dict": @{@"state": @"CA"},
```
pasin, line 404: Can you add a Blob in here as well?

pasin (approved):
Small comments for adding blob in a nested structure in test.

pasin (approved):

PR #2814, opened 2021-03-12
Add init/setJSON functions 

jayahariv:
* Convert the nullable constructor/setters for data to non-nullable. In case of nil, use empty dictionary. (swift and objc)

Only for objc
* Add `initWithJSON:error` and `setJSON:error` functions to MutableDocument, MutableArray, MutableDictionary.
* Enhanced the `cbl_toCBLObject` dictionary category to handle in case of blob metadata in it.
* Add some unit tests. (Not optimized, since other PR might include some data to reuse)

Blob
* Handle Blob.content access on any Mutable Objects(JSON constructed)
* Skip install when there is digest associated with the metadata.


pasin (approved):

jayahariv:
Oh my! did I added you as reviewer, it's too early to review, only in draft mode. Sorry! will add you back when its ready for review.

jayahariv:
Failure on `testConflictResolverReturningBlobFromDifferentDB` might be related to the this change.

pasin:
It this ready to re-review?

jayahariv:
Yes, read to review

Objective-C/CBLMutableArray.mm
```diff
@@ -228,14 +241,47 @@ - (void) insertDictionary: (nullable CBLDictionary*)value atIndex: (NSUInteger)i
...
+    FLSliceResult result = FLEncoder_Finish(enc, &flEerror);
+    if (!result.buf) {
+        CBLWarnError(Database, @"%@: Error decoding JSON: %d", self, flEerror);
+        convertError(flEerror, error);
+        return NO;
+    }
+    
+    FLArray array = FLValue_AsArray(FLValue_FromData(C4Slice(result), kFLTrusted));
+    CBL_LOCK(self.sharedLock) {
+        _array.clear();
```
pasin, line 275: Can we just replace _array with a new MArray? There is `MArray(MValue *mv, MCollection *parent)` that can be used to create a new Array with `array` and the parent from `_array.parent`.

Objective-C/CBLMutableDictionary.mm
```diff
@@ -143,6 +153,39 @@ - (void) setData: (nullable NSDictionary<NSString*,id>*)data {
...
+    flEerror = {};
+    FLSliceResult result = FLEncoder_Finish(enc, &flEerror);
+    if (!result.buf) {
+        CBLWarnError(Database, @"%@: Error decoding JSON: %d", self, flEerror);
+        convertError(flEerror, error);
+        return NO;
+    }
+    
+    FLDict dict = FLValue_AsDict(FLValue_FromData(C4Slice(result), kFLTrusted));
+    _dict.clear();
```
pasin, line 176: Can we just replace _array with a new MDict? There is `MDict(MValue *mv, MCollection *parent)` that can be used to create a new Array with `dict` and the parent from `_array.parent`.

pasin (changes requested):

Objective-C/CBLMutableArray.mm
jayahariv: https://issues.couchbase.com/browse/CBL-1760

Objective-C/CBLMutableDictionary.mm
jayahariv: https://issues.couchbase.com/browse/CBL-1760

pasin (approved):

PR #2821, opened 2021-04-06
CBL-1455: Fix flaky test

jayahariv:
* when both replicators ran concurrently and false expectation to finish in similar time.
* now its running separately, and test validation for replicator and listener on same database also done.

### DIFF
--- a/Objective-C/Tests/URLEndpointListenerTest.m
+++ b/Objective-C/Tests/URLEndpointListenerTest.m
@@ -1108,6 +1108,7 @@ typedef CBLURLEndpointListener Listener;
 - (void) testReplicatorAndListenerOnSameDatabase {
     if (!self.keyChainAccessAllowed) return;
     
+    timeout = 5.0;
     XCTestExpectation* exp1 = [self expectationWithDescription: @"replicator#1 stopped"];
     XCTestExpectation* exp2 = [self expectationWithDescription: @"replicator#2 stopped"];
     
@@ -1123,8 +1124,8 @@ typedef CBLURLEndpointListener Listener;
     Assert([self.db saveDocument: doc1 error: &err], @"Fail to save db1 %@", err);
     
     CBLDatabaseEndpoint* target = [[CBLDatabaseEndpoint alloc] initWithDatabase: self.db];
-    CBLReplicator* repl1 = [self replicator: self.otherDB continous: YES target: target
-                                 serverCert: nil];
+    __block CBLReplicator* repl1 = [self replicator: self.otherDB continous: YES target: target
+                                         serverCert: nil];
     
     // Replicator#2 (DB#2 -> Listener(otherDB))
     Assert([self deleteDBNamed: @"db2" error: &err], @"Failed to delete db2 %@", err);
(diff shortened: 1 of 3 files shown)


Swift/Tests/ReplicatorTest.swift
```diff
@@ -1144,6 +1144,7 @@ class ReplicatorTest_Main: ReplicatorTest {
     }
     
     func testMaxRetryWaitTimeOfReplicator() {
+        timeout = 12
```
jayahariv, line 1147:
```suggestion
        timeout = 12 // already it takes 8 secs of retry, hence 12secs timeout. 
```

Objective-C/Tests/URLEndpointListenerTest.m
```diff
@@ -1108,6 +1108,7 @@ - (void) testMultipleReplicatorsOnReadOnlyListener {
 - (void) testReplicatorAndListenerOnSameDatabase {
     if (!self.keyChainAccessAllowed) return;
     
+    timeout = 5.0;
```
pasin, line 1111: Is this your intention to set the timeout to 5 second? Will this affect the other tests?

Swift/Tests/ReplicatorTest.swift
```diff
@@ -1144,6 +1144,7 @@ class ReplicatorTest_Main: ReplicatorTest {
     }
     
     func testMaxRetryWaitTimeOfReplicator() {
+        timeout = 12 // already it takes 8 secs of retry, hence 12secs timeout.
```
pasin, line 1147: I don't remember. Is this affect the other tests?

Objective-C/Tests/URLEndpointListenerTest.m
jayahariv: https://github.com/couchbase/couchbase-lite-ios/blob/master/Objective-C/Tests/ReplicatorTest.m#L82

Its reset inside setup() method.

jayahariv: this will be reset inside the Replicator>setup() method 
https://github.com/couchbase/couchbase-lite-ios/blob/master/Objective-C/Tests/ReplicatorTest.m#L82

Swift/Tests/ReplicatorTest.swift
jayahariv: I was under the assumption, it will be reset in setup method, but probably not. Updated the code to set the 10secs in setup method.

Swift/Tests/ReplicatorTest.swift
```diff
@@ -23,7 +23,7 @@ import CouchbaseLiteSwift
 class ReplicatorTest: CBLTestCase {
     var oDB: Database!
     var repl: Replicator!
-    var timeout: TimeInterval = 10  // At least 10 to cover single-shot replicator's retry logic
+    var timeout: TimeInterval = 10
```
pasin, line 26: var timeout: TimeInterval! without = 10 as the value will be set in setup.

pasin (changes requested):

pasin (approved):

PR #2829, opened 2021-04-23
feat: enhanced config API

jayahariv:
* change the DBConfig and ReplicatorConfig to struct
* remove the readOnly flag
* add couple of tests to validate the update
* tests chnges  to make config a 'var', so we can update the properties.


pasin (approved):
Much cleaner.

PR #2832, opened 2021-04-27
CBL-1879: Add unit tests to validate the Config API

jayahariv:
## DatabaseConfiguration
1. Ensure that all of the properties will be initialized correctly when using the new Configuration API.
    - testSetDatabaseConfigurationProperties
2. Ensure that all non required properties are initialized with their default value correctly when they are not set.
    -  testDefaultDatabaseConfiguration
3. Ensure that the configuration object specified to the target object’s constructor is copied and subsequent changes to the original configuration object will not affect the target object’s configuration.
    -  testSetDatabaseConfigurationProperties
4. Ensure that the configuration object specified to the configuration’s constructor is copied and subsequent changes to the original configuration object will not affect the new configuration’s object.
    - testCopyingDatabaseConfiguration**

## ReplicatorConfiguration
1. Ensure that all of the properties will be initialized correctly when using the new Configuration API.
    - testReplicationConfigSetterMethods
2. Ensure that all non required properties are initialized with their default value correctly when they are not set.
    - testDefaultReplicatorConfiguration
3. Ensure that the configuration object specified to the target object’s constructor is copied and subsequent changes to the original configuration object will not affect the target object’s configuration.
    - testReplicationConfigSetterMethods
4. Ensure that the configuration object specified to the configuration’s constructor is copied and subsequent changes to the original configuration object will not affect the new configuration’s object.
    - testReplicationConfigCopying

## ListenerConfiguration
1. Ensure that all of the properties will be initialized correctly when using the new Configuration API.
    - testSetListenerConfigurationProperties
2. Ensure that all non required properties are initialized with their default value correctly when they are not set.
    - testDefaultListenerConfiguration
3. Ensure that the configuration object specified to the target object’s constructor is copied and subsequent changes to the original configuration object will not affect the target object’s configuration.
    - testSetListenerConfigurationProperties
4. Ensure that the configuration object specified to the configuration’s constructor is copied and subsequent changes to the original configuration object will not affect the new configuration’s object.
    - testCopyingListenerConfiguration

### DIFF
--- a/Swift/Tests/ReplicatorTest.swift
+++ b/Swift/Tests/ReplicatorTest.swift
@@ -1117,26 +1117,26 @@ class ReplicatorTest_Main: ReplicatorTest {
     
     // MARK: ReplicatorConfig
     
-    func testReplicationConfigCopying() throws {
+    func testCopyingReplicatorConfiguration() throws {
         let basic = BasicAuthenticator(username: "abcd", password: "1234")
         let target = URLEndpoint(url: URL(string: "ws://foo.couchbase.com/db")!)
-        var temp = ReplicatorConfiguration(database: oDB, target: target)
+        var config1 = ReplicatorConfiguration(database: oDB, target: target)
         #if COUCHBASE_ENTERPRISE
-        temp.acceptOnlySelfSignedServerCertificate = true
+        config1.acceptOnlySelfSignedServerCertificate = true
         #endif
-        temp.continuous = true
-        temp.authenticator = basic
-        temp.channels = ["c1", "c2"]
-        temp.documentIDs = ["d1", "d2"]
-        temp.headers = ["a": "aa", "b": "bb"]
-        temp.replicatorType = .pull
-        temp.heartbeat = 211
-        temp.maxRetries = 223
-        temp.maxRetryWaitTime = 227
+        config1.continuous = true
+        config1.authenticator = basic
+        config1.channels = ["c1", "c2"]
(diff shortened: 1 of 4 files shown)


Swift/Tests/ReplicatorTest.swift
```diff
@@ -1133,4 +1114,166 @@ class ReplicatorTest_Main: ReplicatorTest {
         repl.removeChangeListener(withToken: token1)
         repl.removeChangeListener(withToken: token3)
     }
+    
+    // MARK: ReplicatorConfig
+    
+    func testReplicationConfigCopying() throws {
```
pasin, line 1120: Should be testCopyingReplicatorConfiguration to match the others?

Swift/Tests/ReplicatorTest.swift
```diff
@@ -1133,4 +1114,166 @@ class ReplicatorTest_Main: ReplicatorTest {
...
+        let basic = BasicAuthenticator(username: "abcd", password: "1234")
+        let target = URLEndpoint(url: URL(string: "ws://foo.couchbase.com/db")!)
+        var temp = ReplicatorConfiguration(database: oDB, target: target)
```
pasin, line 1123: maybe use config1 instead of temp?

pasin (changes requested):
Can you also add tests to test getting a copy of config from database, replicator, and so on?

The test should ensure that changing properties of the copied of config returned from those objects are isolated from the original config.

pasin (approved):

PR #2839, opened 2021-05-19
fix: build error

jayahariv:
` unused variable 'err'` 
* its an unnecessary check, or assert will fail.


Objective-C/CBLReplicator.mm
```diff
@@ -389,6 +389,8 @@ - (void) setProgressLevel: (CBLReplicatorProgressLevel)level {
     if (_repl) {
         C4Error err = {};
         assert(c4repl_setProgressLevel(_repl, (C4ReplicatorProgressLevel)level, &err));
+        if (err.code > 0)
```
pasin, line 392: If err is not used, can we pass null to c4repl_setProgressLevel() instead?

pasin (changes requested):

PR #2841, opened 2021-05-21
fix: cookie being appended with comma

jayahariv:
* this should ideally add new cookies with semi-colons.
* https://developer.apple.com/documentation/foundation/nsmutableurlrequest/1408793-setvalue

### DIFF
--- a/Objective-C/Internal/Replicator/CBLWebSocket.mm
+++ b/Objective-C/Internal/Replicator/CBLWebSocket.mm
@@ -375,13 +375,16 @@ static void doDispose(C4Socket* s) {
         _logic[slice2string(header.keyString())] = slice2string(header.value().asString());
     
     slice sessionCookie = _options[kC4ReplicatorOptionCookies].asString();
+    NSMutableString* cookies = [NSMutableString string];
     if (sessionCookie.buf)
-        [_logic setValue: sessionCookie.asNSString()  forHTTPHeaderField: @"Cookie"];
-    
+        [cookies appendFormat: @"%@;", sessionCookie.asNSString()];
     
     NSString* cookie = [_db getCookies: _remoteURL];
     if (cookie.length > 0)
-        [_logic setValue: cookie  forHTTPHeaderField: @"Cookie"];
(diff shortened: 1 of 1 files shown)


Objective-C/Internal/Replicator/CBLWebSocket.mm
```diff
@@ -376,12 +376,12 @@ - (void) _sendWebSocketRequest {
...
     slice sessionCookie = _options[kC4ReplicatorOptionCookies].asString();
     if (sessionCookie.buf)
-        [_logic addValue: sessionCookie.asNSString()  forHTTPHeaderField: @"Cookie"];
+        [_logic setValue: sessionCookie.asNSString()  forHTTPHeaderField: @"Cookie"];
     
     
     NSString* cookie = [_db getCookies: _remoteURL];
     if (cookie.length > 0)
-        [_logic addValue: cookie  forHTTPHeaderField: @"Cookie"];
+        [_logic setValue: cookie  forHTTPHeaderField: @"Cookie"];
```
pasin, line 384: How does this change work with multiple Cookies (one from kC4ReplicatorOptionCookies and the other from the cookie store)?

jayahariv: it should ideally comes like cookieOne=value; cookieTwo=value2

pasin: From the doc that you have linked:
```
The new value for the header field. Any existing value for the field is replaced by the new value.
```
Do you mean Cookie key is a special key that setting multiple values will be combined with semi-colons?

jayahariv: if same cookie-key is used, it will override the previous set cookie.

jayahariv: I get what you mean

pasin: OK

pasin (approved):

PR #2842, opened 2021-05-26
CBL-1760: Optimize SetJSON 

jayahariv:


Objective-C/Internal/CBLFleece.hh
```diff
@@ -56,15 +56,16 @@ namespace cbl {
     class DocContext : public fleece::MContext {
     public:
         DocContext(CBLDatabase *db, CBLC4Document* __nullable doc);
+        DocContext(); // setJSON uses this context, without db or c4doc
         
-        CBLDatabase* database() const   {return _db;}
+        CBLDatabase* __nullable database() const   {return _db;}
```
pasin, line 61: DocContext requires a database object for locking. I don't remember all of the detail as it has been a long time. Can you dig this up to see if it's OK not to have the database?

jayahariv: It was required for creating the blob(1).  But we are already creating Blob without DB(2), since the documents created from SetJSON lacks DB. 

(1) https://github.com/couchbase/couchbase-lite-ios/blob/7d9724464d99e5b5cbb3203a58287f29c4f8ef0b/Objective-C/Internal/CBLFleece.mm#L66

(2) https://github.com/couchbase/couchbase-lite-ios/blob/7d9724464d99e5b5cbb3203a58287f29c4f8ef0b/Objective-C/Internal/CBLData.mm#L57

Objective-C/Internal/CBLFleece.mm
```diff
@@ -63,8 +67,12 @@ static bool isOldAttachment(Dict properties) {
...
         slice type = properties.get(C4STR(kC4ObjectTypeProperty)).asString();
         if ((type && type == C4STR(kC4ObjectType_Blob)) || isOldAttachment(properties)) {
-            return [[CBLBlob alloc] initWithDatabase: context->database()
-                                          properties: context->toObject(properties)];
+            if (context->database())
+                return [[CBLBlob alloc] initWithDatabase: context->database()
+                                              properties: context->toObject(properties)];
+            else
+                // when setJSON has a blob, which lacks database. 
+                return [[CBLBlob alloc] initWithProperties: context->toObject(properties)];
```
jayahariv, line 75: ^ this will check for whether there is a database or not. if not, use the `initWithProperties` constructor

Objective-C/CBLMutableDictionary.mm
```diff
@@ -166,18 +166,12 @@ - (BOOL) setJSON: (NSString*)json error: (NSError**)error {
...
-    FLDictIterator_Begin(dict, &iter);
-    FLValue value;
-    while (NULL != (value = FLDictIterator_GetValue(&iter))) {
-        id val = FLValue_GetNSObject(value, nil);
-        _dict.set(FLDictIterator_GetKeyString(&iter), [val cbl_toCBLObject]);
-        FLDictIterator_Next(&iter);
+
+    CBL_LOCK(self.sharedLock) {
+        MDict<id> mDict(new cbl::DocContext(), result, true);
+        _dict.initAsCopyOf(mDict, true);
```
pasin, line 172: I haven't been away from this code for long so I'm not sure if this is correct. A few things that need to be checked and confirmed:

1. cbl::parseJSON() returns FLValue which is a pointer to fleece data from parsing the JSON). I have just noticed that the fleece data (FLSliceResult) is not in this method so it could be leaked inside the parseJSON() itself.
2. To replace the Dict inside the _dict, the fleeceData from 1 needs to be retained as a root similarly to what has been done in CBLDocument.
3. One of the purposes of the DocContext is to hold the C4Doc which owns the fleece data of the doc. To hold a Non-Doc fleece data, it's better to create a new type.

PR #2845, opened 2021-06-02
CBL-1567: Module initializer

jayahariv:
* add a module initializer
* called the c4log_fatalException from it.
* ref: CBL-1852, CBL-1567

### DIFF
--- a/Objective-C/Internal/main.m
+++ b/Objective-C/Internal/main.m
@@ -1,0 +1,23 @@
+//
+//  main.m
+//  CouchbaseLite
+//
+//  Created by Jayahari Vavachan on 6/1/21.
+//  Copyright © 2021 Couchbase. All rights reserved.
+//
+
+#import <Foundation/Foundation.h>
+#import "c4.h"
+
+__attribute__((constructor))
+static void initialize()
+{
+    static int initialized = 0;
+    if (!initialized)
+    {
+        initialized = 1;
+        
+        NSLog(@">>>>>>>>>>>>>>>>>>>>>>>>>>> main initializer!!");
+        c4log_enableFatalExceptionBacktrace();
+    }
+}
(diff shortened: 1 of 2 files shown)


Objective-C/Internal/main.m
```diff
@@ -0,0 +1,22 @@
+//
+//  main.m
+//  CouchbaseLite
+//
+//  Created by Jayahari Vavachan on 6/1/21.
+//  Copyright © 2021 Couchbase. All rights reserved.
```
pasin, line 6: Can you update the license?

pasin (commented):
1. When we do CBLCouchbaseLite.logger, the module initializer can be moved to CBLCouchbaseLite class.
2. Can you confirm if this has been tested on both Objective-C and Swift and on both iOS and macOS.

jayahariv:
1. We cannot move it to CBLCouchbaseLite class, since 'the symbols for any module initializer functions must not be exported by your framework.'

    We can keep the `main.m` and create a separate one for CouchbaseLite which can be exported by framework?

2. Yes, tested on all combination
- CBL_EE_Swift + iPhone simulator
- CBL_EE_Swift + Mac
- CBL_EE_Swift_Tests_iOS_App + device 
- CBL_EE_ObjC + iPhone simulator
- CBL_EE_ObjC + Mac 
- CBL_EE_ObjC_Tests_iOS_App + device

pasin:
> We can keep the main.m and create a separate one for CouchbaseLite which can be exported by framework?

Yes

pasin:
We could move what in here (https://github.com/couchbase/couchbase-lite-ios/blob/master/Objective-C/CBLDatabase.mm#L101-L108) to the Module Initializer as well.

pasin (approved):

jayahariv:
Will create a separate PR for the CBLDatabase initialize() function content

PR #2849, opened 2021-06-05
CBL-1888: FROM clause is mandatory

jayahariv:
* Litecore decided that the "FROM" clause would be required in the query. Hence manually, '*' transformed to default database-name. 
* CBL-1923, CBL-1888
* More details comments in the CBL-1888

### DIFF
--- a/Objective-C/CBLQueryDataSource.m
+++ b/Objective-C/CBLQueryDataSource.m
@@ -32,7 +32,8 @@
 }
 
 - (id) asJSON {
-    return _alias ?  @{@"AS": _alias} : @{ };
+    NSString* columnName = [self columnName];
+    return columnName ? @{ @"AS" : columnName } : @{ };
 }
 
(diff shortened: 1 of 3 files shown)


Objective-C/CBLQueryDataSource.m
```diff
@@ -32,7 +32,8 @@ - (instancetype) initWithDataSource: (id)source as: (nullable NSString*)alias {
 }
 
 - (id) asJSON {
-    return _alias ?  @{@"AS": _alias} : @{ };
+    NSString* columnName = [self columnName];
+    return columnName ? @{ @"AS" : columnName } : @{ };
```
pasin, line 36: Instead of allowing @ { } which is invalid now, we can assert(columnName) instead.

pasin (changes requested):

pasin (approved):

PR #2857, opened 2021-06-23
Query optimize enum exposed

jayahariv:


jayahariv:
not sure, how we can effectively unit test this

pasin (approved):

PR #2862, opened 2021-06-29
CBL-2036: Add unit test to validate 

jayahariv:
* Submodule update with the LiteCore fix
* Unit test to validate the failure.

### DIFF
--- a/Swift/Tests/QueryTest.swift
+++ b/Swift/Tests/QueryTest.swift
@@ -265,6 +265,32 @@ class QueryTest: CBLTestCase {
         XCTAssertEqual(firstNames.count, 5);
     }
     
+    // https://issues.couchbase.com/browse/CBL-2036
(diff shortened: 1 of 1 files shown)


Swift/Tests/QueryTest.swift
```diff
@@ -265,6 +265,32 @@ class QueryTest: CBLTestCase {
         XCTAssertEqual(firstNames.count, 5);
     }
     
+    // https://issues.couchbase.com/browse/CBL-2036
+    func testFTSQueryOnDBNameWithHyphen() throws {
```
pasin, line 269: Based on the test name, does the issue happen only with FTS?

jayahariv: Yea, this happens with FTSQuery only, 
https://github.com/couchbase/couchbase-lite-core/blob/master/LiteCore/Query/QueryParser.cc#L317

pasin (approved):

PR #2863, opened 2021-07-13
fix: issue with n1ql query instance

jayahariv:
* query instance is not save with database.
* add a unit test with validating the n1ql query crash
* description updated with n1ql as well as json string

### DIFF
--- a/Swift/Query.swift
+++ b/Swift/Query.swift
@@ -112,7 +112,7 @@ public class Query {
         })
         
         if tokens.count == 0 {
-            database!.addQuery(self)
+            database.addQuery(self)
         }
         
         let listenerToken = ListenerToken(token)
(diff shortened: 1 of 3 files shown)


Swift/Query.swift
```diff
@@ -156,6 +157,7 @@ public class Query {
     ///     - database  The database to query.
     ///     - expressions  String representing the query expression.
     init(database: Database, expressions: String) {
+        self.database = database
```
pasin, line 160: Seem like self.database will not be nil. Can we change `var database: Database?` to `var database: Database`?

pasin (changes requested):

Swift/Query.swift
jayahariv: made it Database! , since init() {} constructor is used by other classes like Where.swift, etc..

pasin (approved):

PR #2867, opened 2021-07-20
CBL-1985: fix: assert vs Assert

jayahariv:
* looks like assert is a DEBUG only function, which is get ignored when release build is created.

https://github.com/couchbaselabs/fleece/blob/5c0eb9b9c726b65793dd4fa3f352ac9db2139244/Fleece/Support/betterassert.hh#L73

### DIFF
--- a/Objective-C/CBLReplicator.mm
+++ b/Objective-C/CBLReplicator.mm
@@ -388,7 +388,7 @@ static C4ReplicatorValidationFunction filter(CBLReplicationFilter filter, bool i
 - (void) setProgressLevel: (CBLReplicatorProgressLevel)level {
     _progressLevel = level;
     if (_repl)
-        assert(c4repl_setProgressLevel(_repl, (C4ReplicatorProgressLevel)level, nullptr));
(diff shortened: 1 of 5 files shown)


Objective-C/CBLReplicator.mm
```diff
@@ -387,8 +387,11 @@ - (void) removeChangeListenerWithToken: (id<CBLListenerToken>)token {
 
 - (void) setProgressLevel: (CBLReplicatorProgressLevel)level {
     _progressLevel = level;
-    if (_repl)
-        assert(c4repl_setProgressLevel(_repl, (C4ReplicatorProgressLevel)level, nullptr));
+    if (_repl) {
+        BOOL success = c4repl_setProgressLevel(_repl, (C4ReplicatorProgressLevel)level, nullptr);
```
pasin, line 391: Can you add an error to the last parameter as well so the error can be logged? Also, in what circumstance, this can be failed?

pasin (approved):

Objective-C/CBLReplicator.mm
jayahariv: only two cases when replicator is null OR level is wrong, which are already taken care or show build error in platform side.

PR #2869, opened 2021-07-29
fix: kHasAttachments flag may be lost in Conflict Resolver

jayahariv:
* It did successfully push the document, even without this PR change. But local DB doc was missing the kRevHasAttachments flag. 
* CBL-2192

### DIFF
--- a/Objective-C/CBLBlob.mm
+++ b/Objective-C/CBLBlob.mm
@@ -382,8 +382,9 @@ static NSString* const kCBLBlobDataProperty = @kC4BlobDataProperty;
 - (void) fl_encodeToFLEncoder: (FLEncoder)encoder {
     // Note: If CBLDictionary can be encoded independently of CBLDocument,
     // so there could be no extra info:
+    bool hasAttachment = true;
     FLEncoderContext* encContext = (FLEncoderContext*)FLEncoder_GetExtraInfo(encoder);
-    encContext->hasAttachment = true;
+    encContext->outHasAttachment = &hasAttachment;
     if (encContext->document) {
         CBLDatabase* database = encContext->document.database;
         [self checkBlobFromSameDatabase: database];
--- a/Objective-C/CBLBlob.mm
+++ b/Objective-C/CBLBlob.mm
@@ -382,9 +382,13 @@ static NSString* const kCBLBlobDataProperty = @kC4BlobDataProperty;
 - (void) fl_encodeToFLEncoder: (FLEncoder)encoder {
     // Note: If CBLDictionary can be encoded independently of CBLDocument,
     // so there could be no extra info:
-    bool hasAttachment = true;
     FLEncoderContext* encContext = (FLEncoderContext*)FLEncoder_GetExtraInfo(encoder);
-    encContext->outHasAttachment = &hasAttachment;
+    
+    // mark this document includs an attachment
+    bool* outHasAttachment = encContext->outHasAttachment;
+    if (outHasAttachment)
+        *outHasAttachment = true;
+    
     if (encContext->document) {
         CBLDatabase* database = encContext->document.database;
         [self checkBlobFromSameDatabase: database];
--- a/Objective-C/CBLBlob.mm
+++ b/Objective-C/CBLBlob.mm
@@ -383,6 +383,7 @@ static NSString* const kCBLBlobDataProperty = @kC4BlobDataProperty;
     // Note: If CBLDictionary can be encoded independently of CBLDocument,
     // so there could be no extra info:
     FLEncoderContext* encContext = (FLEncoderContext*)FLEncoder_GetExtraInfo(encoder);
+    encContext->hasAttachment = true;
     if (encContext->document) {
         CBLDatabase* database = encContext->document.database;
         [self checkBlobFromSameDatabase: database];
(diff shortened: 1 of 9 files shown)


pasin (changes requested):
The current hasBlob() implementation requires deep iteration as blobs might be in nested structure. Alternative solutions would be

1.  You could do as the saveDocument() method is doing as follows:

```
FLDoc doc = FLDoc_FromResultData(body, kFLTrusted, self.sharedKeys, nullslice);
if (c4doc_dictContainsBlobs((FLDict)FLDoc_GetRoot(doc)))
            revFlags |= kRevHasAttachments;
FLDoc_Release(doc);
```

2. You could also enhance the CBLDocument's `- (FLSliceResult) encode: (NSError**)outError` to something like

```- (FLSliceResult) encode: (NSError**)outError revFlags: (C4RevisionFlags &)outRevFlags;```

Then add the (C4RevisionFlags &)outRevFlags to the FLEncoderContext struct. When encoding CBLBlob, CBLBlob can set the `kRevHasAttachments` flag to the outRevFlags. 

The second approach would have better performance. But as we have already done the first approach in save() method, I think doing the first approach is fine.

File: Objective-C/Internal/CBLFleece.hh [hunk 1]
```diff
@@ -33,6 +33,7 @@ NS_ASSUME_NONNULL_BEGIN
 typedef struct FLEncoderContext {
     CBLDocument* __nullable document = nullptr;
     bool encodeQueryParameter = false;
+    bool hasAttachment = false;
```
pasin, line 36: 1. I think we need comment on each properties in this FLEncoderContext.
2. As hasAttachment is for setting, I think hasAttachment should be a bool pointer e.g. bool* outHasAttachment.

Objective-C/CBLDocument.mm
```diff
@@ -210,7 +210,7 @@ - (void) replaceC4Doc: (CBLC4Document*)c4doc {
 
 #pragma mark - Fleece Encoding
 
-- (FLSliceResult) encode: (NSError**)outError {
+- (FLSliceResult) encode: (C4RevisionFlags*)outRevFlags error:(NSError**)outError {
```
pasin, line 213: Should we name the method like `(FLSliceResult) encodeWithRevFlags: (C4RevisionFlags*)outRevFlags error:(NSError**)outError`?

Objective-C/CBLDocument.mm
```diff
@@ -228,6 +228,10 @@ - (FLSliceResult) encode: (NSError**)outError {
     FLSliceResult body = FLEncoder_Finish(encoder, &flErr);
     if (!body.buf)
         createError(flErr, [NSString stringWithUTF8String: errMessage], outError);
+    
+    // if it has attachments, add the flag
+    *outRevFlags |= ctx.hasAttachment ? kRevHasAttachments : 0;
```
pasin, line 233: In theory, outRevFlags could be null. We need `if (!outRevFlags)` here.

pasin (changes requested):

Objective-C/CBLDocument.mm
pasin: Also can you add some description about what the encode method is doing with the outRevFlags?

pasin (changes requested):

Objective-C/CBLBlob.mm
```diff
@@ -382,7 +382,9 @@ - (id) cbl_toCBLObject {
 - (void) fl_encodeToFLEncoder: (FLEncoder)encoder {
     // Note: If CBLDictionary can be encoded independently of CBLDocument,
     // so there could be no extra info:
+    bool hasAttachment = true;
     FLEncoderContext* encContext = (FLEncoderContext*)FLEncoder_GetExtraInfo(encoder);
+    encContext->outHasAttachment = &hasAttachment;
```
pasin, line 387: The pointer of hasAttachment will be invalid right after this method return.

I think it should be like:
```
bool* outHasAttachment = encContext->outHasAttachment;
if (outHasAttachment) {
    *outHasAttachment = true;
}
```

Objective-C/CBLDocument.mm
```diff
@@ -228,6 +228,11 @@ - (FLSliceResult) encode: (NSError**)outError {
     FLSliceResult body = FLEncoder_Finish(encoder, &flErr);
     if (!body.buf)
         createError(flErr, [NSString stringWithUTF8String: errMessage], outError);
+
```
pasin, line 231: Two things in this method:

1. The FLEncoderContext is missing a pointer of outHasAttachment. I think it could be like:

bool hasAttachment = false;
FLEncoderContext ctx = { ... , outHasAttachment = &hasAttachment }
```
2. Why is encodeQueryParameter true here? I might have misunderstood something.

pasin (changes requested):

Objective-C/CBLDocument.mm
jayahariv: `encodeQueryParameter` probably a mistake in setting in CBLDocument. I have removed and tried running unit tests.

PR doesn't prove its required. 
https://github.com/couchbase/couchbase-lite-ios/pull/2812

pasin (approved):

PR #2872, opened 2021-08-04
CBL-2168: Add flag for app extension safe APIs

jayahariv:
* background monitor methods will be skipped in case of extension
https://github.com/snej/MYUtilities/blob/master/MYBackgroundMonitor.m#L47
https://github.com/snej/MYUtilities/blob/master/MYBackgroundMonitor.m#L67
https://github.com/snej/MYUtilities/blob/master/MYBackgroundMonitor.m#L81
https://github.com/snej/MYUtilities/blob/master/MYBackgroundMonitor.m#L95

* even though we access the UIApplication sharedApplication, its dynamically getting it via performSelector call.
https://github.com/snej/MYUtilities/blob/master/MYBackgroundMonitor.m#L33

### DIFF
--- a/xcconfigs/Project.xcconfig
+++ b/xcconfigs/Project.xcconfig
@@ -63,6 +63,9 @@ VERSIONING_SYSTEM                  = apple-generic
 
 SUPPORTS_MACCATALYST               = YES
 
+// background monitor feature will be disabled if its extension project.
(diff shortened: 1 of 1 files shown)


pasin (approved):

pasin:
I forgot that I made those changes. Thanks for checking.

PR #2873, opened 2021-08-07
CBL-1920: Thread safety for LiteCore APIs

jayahariv:
* Protect the `c4queryenum_release` from multiple threads accessing it
* comments about calling the thread-unsafe calls from locks.

### DIFF
--- a/Objective-C/CBLDatabase.mm
+++ b/Objective-C/CBLDatabase.mm
@@ -823,9 +823,7 @@ static void dbObserverCallback(C4DatabaseObserver* obs, void* context) {
     if (!_c4db)
         return convertError(err, outError);
     
-    CBL_LOCK(self) {
-        _sharedKeys = c4db_getFLSharedKeys(_c4db);
-    }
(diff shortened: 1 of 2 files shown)


Objective-C/CBLDatabase.mm
```diff
@@ -823,7 +823,9 @@ - (BOOL) open: (NSError**)outError {
     if (!_c4db)
         return convertError(err, outError);
     
-    _sharedKeys = c4db_getFLSharedKeys(_c4db);
+    CBL_LOCK(self) {
+        _sharedKeys = c4db_getFLSharedKeys(_c4db);
+    }
```
pasin, line 828: I think we don't need to lock this as the open is called when CBLDatabase object is created so there shouldn't be anything concurrent yet.

jayahariv: `initWithName: config: error` is a public API right, which can be called concurrently

jayahariv: nvm, each call will end up in separate db instances.

pasin (approved):

PR #2874, opened 2021-08-07
Add UseLock method to NSObject

jayahariv:
* For objects which are locked outside its own class, it will be called via useLock method in NSObject


jayahariv:
Closing for now! Will take a look at it Post-Lithium

PR #2875, opened 2021-08-07
lock query instead of database in case of c4query_columnCount

jayahariv:


Objective-C/CBLQueryResult.mm
```diff
@@ -55,7 +55,7 @@ - (instancetype) initWithResultSet: (CBLQueryResultSet*)rs
 #pragma mark - CBLArray
 
 - (NSUInteger) count {
-    CBL_LOCK(_rs.database) {
+    CBL_LOCK(_rs.query) {
```
pasin, line 58: Instead of calling c4query_columnCount() here, I think it's better to move this to CBLQuery class itself so CBLQueryResult doesn't need to use c4Query at all.

pasin (changes requested):

Objective-C/CBLQuery.mm
```diff
@@ -275,6 +275,12 @@ - (instancetype) copyWithZone: (NSZone*)zone {
     }
 }
 
+- (NSUInteger) count {
```
pasin, line 278: I think 'count' is ambiguous when it is at the CBLQuery class. Can you rename to columnCount?

Objective-C/Internal/CBLQueryResultSet+Internal.h
```diff
@@ -31,6 +31,7 @@ NS_ASSUME_NONNULL_BEGIN
                    columnNames: (NSDictionary*)columnNames;
 
 @property (nonatomic, readonly) CBLDatabase* database;
+@property (nonatomic, readonly) CBLQuery* query;
```
pasin, line 34: Do we need c4Query property anymore? Can you have a quick check?

pasin (changes requested):

pasin (approved):

PR #2881, opened 2021-09-14
clean the listeners after use

jayahariv:


pasin (commented):
Is there an issue that requires to remove to token?

jayahariv:
There is an intermittent Jenkins failure [CBL-2338](https://issues.couchbase.com/browse/CBL-2338). One of them(fail_testCloseWithActiveLiveQueriesAndReplicators) is due to multiple calls to the expectations. So while going through the test, I saw the change listeners are not removed, which made me to correct it.

pasin (commented):
There is no need to remove the listener after the replicator is stopped.

PR #2883, opened 2021-09-23
CBL-2392: Add a small delay to make the unit test pass

jayahariv:
Add a small delay to make the unit test pass

### DIFF
--- a/Objective-C/Tests/DocumentExpirationTest.m
+++ b/Objective-C/Tests/DocumentExpirationTest.m
@@ -381,9 +381,14 @@
     Assert([self.db setDocumentExpirationWithID: doc.id expiration: expiryDate error: &err]);
     AssertNil(err);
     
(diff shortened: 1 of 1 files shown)


pasin:
Can you explain why putting 0.1 delay make the unit test pass?

pasin (commented):
This seems to be a LiteCore issue that doesn't update the expiry time but I don't understand why adding 0.1 delay makes the test pass. Can you explain? Also I think it's better to disable the test until the underling issue is fix instead of adding a workaround.

pasin:
Also there seems to be an issue with PR validation.

jayahariv:
PR validation is failing on other unit tests. 

My understanding was 0.1 is giving enough time to save the new value. Which seems to be passing tests 5/5 times. I haven't seen the same test fail after this delay. But sure, we could wait for the LiteCore fix, and then enable, which would be the ideal fix.

PR #2900, opened 2021-10-09
CBL-1410: LiveQuery to use C4QueryObserver

jayahariv:
* remove CBL handling
* use LC API directly
* disable the expiry doc unit test


Objective-C/Internal/CBLQueryChangeNotifier.h
```diff
@@ -0,0 +1,29 @@
...
+@interface CBLQueryChangeNotifier : CBLChangeNotifier
+
+/** Starts an observer and listener on the queue. */
+- (CBLChangeListenerToken*) addChangeListenerWithQueue: (nullable dispatch_queue_t)queue
+                                              listener: (void (^)(CBLQueryChange*))listener
+                                                 queue: (CBLQuery*)query
+                                           columnNames: (NSDictionary *)columnNames;
+
+/** Removes the observer and listener*/
+- (NSUInteger) removeChangeListenerWithToken: (id<CBLListenerToken>)token;
```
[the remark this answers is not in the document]
pasin: It doesn't seem to get used so I think it is better to just do void.

pasin (changes requested):
Please make sure the following cases are covered:

1. When adding a second listener after the first listener is notified, the second listener should get the change (current result).

2. When having more than one listeners, each listener should have independent result set. This means that both listeners should be able to iterate through the result correct when getting the same change notified.

3. Change the parameters should result to a new query and a new change notified.

4. Test that the changes are coalescing. This test could be that after the first notification, create 2 docs. The change could be notified only once, not two.

jayahariv:
Regarding the tests, will keep them in a separate PR CBL-2582

Objective-C/CBLListenerToken.h
```diff
@@ -23,7 +23,7 @@ NS_ASSUME_NONNULL_BEGIN
  Listener token returned when adding a change listener. The token is used
  for removing the added change listener.
  */
-@protocol CBLListenerToken <NSObject>
+@protocol CBLListenerToken <NSObject, NSCopying>
```
pasin, line 26: Ugh, I don't think we should NSCopying as this means that our API allows the token to be copied :(. So reverting back to use UUID  as a key is better. Sorry for this.

Objective-C/CBLQuery.mm
```diff
@@ -63,6 +68,12 @@ - (instancetype) initWithDatabase: (CBLDatabase*)database
         _database = database;
         _json = json;
         _language = kC4JSONQuery;
+        
+        // Return error if the query is not compiled;
+        NSError* error = nil;
+        if (![self compile: &error])
+            [NSException raise: NSInvalidArgumentException
```
pasin, line 75: Can you also log the error for the case that somebody experiences a crash due to this so they have some info in the log?

Objective-C/CBLQuery.mm
```diff
@@ -200,18 +209,24 @@ - (CBLQueryParameters*) parameters {
 
 - (void) setParameters: (CBLQueryParameters*)parameters {
     CBL_LOCK(self) {
-        if (parameters)
+        if (parameters) {
+            NSError* error = nil;
+            NSData* params = [parameters encode: &error];
+            if (!params)
+                [NSException raise: NSInvalidArgumentException
```
pasin, line 216: Can you also log the error for the case that somebody experiences a crash due to this so they have some info in the log?

pasin (changes requested):

Objective-C/Internal/CBLChangeListenerToken.m
```diff
@@ -45,4 +45,8 @@ - (void) postChange: (id)change {
     });
 }
 
+- (nonnull id)copyWithZone:(nullable NSZone *)zone {
+    return [[[self class] alloc] initWithListener: _listener queue: _queue];
+}
+
```
jayahariv, line 51:
```suggestion
```

pasin (approved):

PR #2901, opened 2021-10-16
Change Database’s createQuery() to return error

pasin:
* Changed the signature to return an error or throw exception when the given query string is invalid.
* Changed Swift to call CBLDatabase’s -createQuery:error instead of using internal constructor.
* Renamed CBLQuery's -(BOOL)check: method to -(BOOL)compile:
* Renamed CBLQuery+N.h tp CBLQuery+N1QL.h
* Remove CBLQuery+N.h from Swift’s modulemap file as it’s not used by Swift code (Swift will not call the methods in CBLQuery+N1QL.h directly).

CBL-2483


pasin:
Tests failed are not related to this PR.

PR #2903, opened 2021-10-20
Update build script

jayahariv:
* remove combined/framework build setup, since we will only create XCFramework from now on.
* keep --nodocs flag for API docs free binary creation.
* update the instrucments command to use xctrace
* ~~we need this to be cherry-picked to Lithium Beta03 as well~~.


pasin:
What we don't have (supported platform wise) if we don't merge this PR in Lithium?

jayahariv:
If we don't merge, it will create extra '.frameworks' with XCFramework. 

IIRC Ming does avoid storing the '.framework' to `latestbuilds.service.couchbase.com`. Even then the build job will take unnecessary time to create them.

jayahariv:
Useful link 
https://help.apple.com/xcode/mac/11.4/#/dev544efab96

PR #2905, opened 2021-11-02
[lithium] bump deployment target to iOS 10.3

jayahariv:


pasin (commented):
10.3 could be right but I don't remember why it is not 10.0?

jayahariv:
Since 10.0 and 10.3 are similar but with some extra features without any API breaking changes. Which makes sense for any users to adopt it seamlessly.

pasin:
If I have an app that set the target deployment to 10.0, can the app's project use CBL with 10.3 as the target deployment?

pasin:
IMO, Setting it to 10.0 doesn't require to verify and doesn't cause any problems.

jayahariv:
> If I have an app that set the target deployment to 10.0, can the app's project use CBL with 10.3 as the target deployment?
No. they need to upgrade that to 10.3.

xcconfigs/Project.xcconfig
```diff
@@ -22,9 +22,9 @@ CBL_BUILD_NUMBER                   = 0
 CBL_EXPORTED_SYMBOLS_FILE          = Objective-C/CouchbaseLite.exp
 CBL_SWIFT_MODULEMAP_FILE           = Swift/CouchbaseLiteSwift.modulemap
 
-IPHONEOS_DEPLOYMENT_TARGET         = 9.0
+IPHONEOS_DEPLOYMENT_TARGET         = 10.3
```
jayahariv, line 25:
```suggestion
IPHONEOS_DEPLOYMENT_TARGET         = 10.0
```

xcconfigs/Project.xcconfig
```diff
@@ -22,9 +22,9 @@ CBL_BUILD_NUMBER                   = 0
...
 MACOSX_DEPLOYMENT_TARGET           = 10.12
-TVOS_DEPLOYMENT_TARGET             = 9.0
+TVOS_DEPLOYMENT_TARGET             = 10.3
```
jayahariv, line 27:
```suggestion
TVOS_DEPLOYMENT_TARGET             = 10.0
```

pasin (approved):

PR #2907, opened 2021-11-09
CBL-2491: Update info plist - Copyright & CBLEdition

jayahariv:
* make sure, copyright always uses the current year when building new binary
* also pass CBLEdition property with the info plist to distinguish between EE and CE


Objective-C/Info.plist
```diff
@@ -19,8 +19,10 @@
 	<key>CFBundleVersion</key>
 	<string>$(CBL_BUILD_NUMBER)</string>
 	<key>NSHumanReadableCopyright</key>
-	<string>Copyright © 2017 Couchbase. All rights reserved.</string>
+	<string>$(CBL_COPYRIGHT)</string>
```
pasin, line 22: I'm a bit reluctant to this change and it seems like it's safe to have the copy string in the Info.plist incase somebody grab the project and build by themselves without using the script.

pasin: Can we do two things?
1. < string >Copyright © $(CBL_COPYRIGHT_YEAR) Couchbase. All rights reserved.< / string>
2. Add a build phase script to assign a default value to CBL_COPYRIGHT_YEAR variable if it doesn't have  the value? (NEED to try it out)

pasin (changes requested):

xcconfigs/Project.xcconfig
```diff
@@ -21,6 +21,7 @@ CBL_VERSION_STRING                 = 3.1.0
 CBL_BUILD_NUMBER                   = 0
 CBL_EXPORTED_SYMBOLS_FILE          = Objective-C/CouchbaseLite.exp
 CBL_SWIFT_MODULEMAP_FILE           = Swift/CouchbaseLiteSwift.modulemap
+CBL_COPYRIGHT_YEAR                 = 2020
```
pasin, line 24: Should it be 2021?

pasin (approved):
One comment about the year. The rest looks good to me.

xcconfigs/Project.xcconfig
jayahariv: I kept that for testing, forget to revert it.
```suggestion
CBL_COPYRIGHT_YEAR                 = 2021
```

PR #2911, opened 2021-11-16
CBL-2546: Error description in swift & objc

jayahariv:
* Error description in swift & objc

### DIFF
--- a/Swift/Errors.swift
+++ b/Swift/Errors.swift
@@ -1,0 +1,238 @@
+//
+//  Errors.swift
+//  CBL ObjC
+//
+//  Copyright (c) 2021 Couchbase  Inc All rights reserved.
+//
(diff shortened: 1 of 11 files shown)


pasin (approved):

Swift/Errors.swift
```diff
@@ -0,0 +1,237 @@
...
+//
+
+
+import Foundation
+
+public struct CBLError {
+    public static let domain = CBLErrorDomain
+    
+    /// Internal assertion failure
+    public static let assertionFailed       = 1
```
jayahariv, line 27: For easy reference for the users, declaring the values here seems to be better. 

Else users might want to go and see the objc documentation together with swift in order to understand the error code.

jayahariv:
wrong commit message b78fb4a6921bd64b44a5c311004bdc17be55279a

PR #2915, opened 2021-11-18
bump ios and macos for package swift

jayahariv:


jayahariv:
unit tests failure is not related! even though they are disabled in later commit.

pasin (approved):

PR #2919, opened 2021-11-24
Update Deprecated Sec APIs

jayahariv:


Objective-C/Internal/Replicator/CBLTrustCheck.mm
```diff
@@ -173,8 +173,19 @@ - (NSURLCredential*) checkTrust: (NSError**)outError {
 
     // If using cert-pinning, accept cert iff it matches the pin:
     if (_pinnedCertData) {
-        SecCertificateRef cert = SecTrustGetCertificateAtIndex(_trust, 0);
-        if ([_pinnedCertData isEqual: CFBridgingRelease(SecCertificateCopyData(cert))]) {
+        SecCertificateRef cert = NULL;
+#if __MAC_OS_X_VERSION_MIN_REQUIRED >= 120000 || __IPHONE_OS_VERSION_MIN_REQUIRED >= 150000
+        if (@available(macOS 12.0, iOS 15.0, *)) {
```
pasin, line 178: Do we still need `if (@available(macOS 12.0, iOS 15.0, *)) {`?

Objective-C/Internal/Replicator/CBLTrustCheck.mm
```diff
@@ -112,7 +112,7 @@ - (BOOL) forceTrusted {
         return NO;
     SecTrustSetExceptions(_trust, exception);
     CFRelease(exception);
-#if TARGET_OS_MACCATALYST
+#if __MAC_OS_X_VERSION_MIN_REQUIRED >= 101400 || __IPHONE_OS_VERSION_MIN_REQUIRED >= 120000
     if (@available(iOS 12.0, macos 10.14, *)) {
```
pasin, line 116: Do we need `if (@available(iOS 12.0, macos 10.14, *))`?

Objective-C/Internal/Replicator/CBLTrustCheck.mm
```diff
@@ -146,7 +146,7 @@ - (NSURLCredential*) checkTrust: (NSError**)outError {
     // Evaluate trust:
     SecTrustResultType result;
     OSStatus err;
-#if TARGET_OS_MACCATALYST
+#if __MAC_OS_X_VERSION_MIN_REQUIRED >= 101400 || __IPHONE_OS_VERSION_MIN_REQUIRED >= 120000
```
pasin, line 149: I'm trying to understand the problem. Doesn't this have the affect when we build the framework not when the user uses the framework?

Objective-C/Internal/Replicator/CBLTrustCheck.mm
pasin: When I looked at this again, the fix will have side effect because we will left out anything in #else completely when we built the binary. #if __IPHONE_OS_VERSION_MIN_REQUIRED is for compile time checking of the min target SDK version.

PR #2922, opened 2021-12-10
[hydrogen] hotfix-11072

jayahariv:
* keep a state for the web socket to determine closing/closed
* if there is failure in getCookies will return the error, which will check for the state.

### DIFF
--- a/Objective-C/Internal/Replicator/CBLWebSocket.mm
+++ b/Objective-C/Internal/Replicator/CBLWebSocket.mm
@@ -602,7 +602,6 @@ static BOOL checkHeader(NSDictionary* headers, NSString* header, NSString* expec
 - (void) closeSocket {
     CBLLogInfo(WebSocket, @"%@ CBLWebSocket closeSocket requested", self);
     CBL_LOCK(self) {
-        // ignore the second closing or closed
         if (_state <= kCBLWebsocketStateClosing) {
             CBLWarn(WebSocket, @"%@ Ignoring the closeSocket request, since it already closing or closed (state=%d)",
                     self, _state);
--- a/Objective-C/Internal/Replicator/CBLWebSocket.mm
+++ b/Objective-C/Internal/Replicator/CBLWebSocket.mm
@@ -613,8 +613,10 @@ static BOOL checkHeader(NSDictionary* headers, NSString* header, NSString* expec
 // Closes the connection and passes the NSError (if any) to LiteCore.
 - (void) closeWithError: (NSError*)error {
     CBL_LOCK(self) {
-        if (_closing)
+        if (_closing) {
+            CBLWarn(Sync, @"%@ Websocket is already closing. ignoring the close now", self);
             return;
+        }
         
         _closing = YES;
     }
--- a/Objective-C/Internal/Replicator/CBLWebSocket.mm
+++ b/Objective-C/Internal/Replicator/CBLWebSocket.mm
@@ -69,13 +69,6 @@ struct PendingWrite {
 @interface CBLWebSocket () <NSStreamDelegate>
 @end
 
-// For controlling async closing, closed states. 
-typedef enum {
-    kCBLWebsocketStateClosed = 0,       ///< The websocket was closed.
-    kCBLWebsocketStateClosing,          ///< The websocket was asked to close but in progress.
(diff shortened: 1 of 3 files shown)


Objective-C/Internal/Replicator/CBLWebSocket.mm
```diff
@@ -576,11 +600,26 @@ - (void) completedReceive: (size_t)byteCount {
 
 // callback from C4Socket
 - (void) closeSocket {
-    CBLLogInfo(WebSocket, @"CBLWebSocket closeSocket requested");
+    CBLLogInfo(WebSocket, @"%@ CBLWebSocket closeSocket requested", self);
+    CBL_LOCK(self) {
+        // ignore the second closing or closed
```
jayahariv, line 605:
```suggestion
```

Objective-C/Internal/Replicator/CBLWebSocket.mm
```diff
@@ -69,6 +69,13 @@
 @interface CBLWebSocket () <NSStreamDelegate>
 @end
 
+// For controlling async closing, closed states. 
+typedef enum {
+    kCBLWebsocketStateClosed = 0,       ///< The websocket was closed.
+    kCBLWebsocketStateClosing,          ///< The websocket was asked to close but in progress.
+    kCBLWebsocketStateConnecting,       ///< The websocket is created and requested to connect.
+} CBLWebsocketState;
```
pasin, line 77: It feels uncompleted as it jumps from Connecting to Closing. Either we add Connected for the sake of completeness OR we change it to just boolean (e.g. closeRequested).

jayahariv: I added connected, created, etc. but there is nothing much of a use though. 

Keeping boolean, _closing i will try and update the PR soon

pasin: After thought, we may not need a flag / state at all. What we need are:

1. Handle the cookie error and call closeWithError:
2. In closeWithError, if already disconnected, just return so that we will not acknowledge closed twice.

Objective-C/Internal/Replicator/CBLWebSocket.mm
```diff
@@ -603,6 +612,15 @@ - (void) closeWithCode: (C4WebSocketCloseCode)code reason: (NSString*)reason {
 
 // Closes the connection and passes the NSError (if any) to LiteCore.
 - (void) closeWithError: (NSError*)error {
+    CBL_LOCK(self) {
+        if (_closing) {
+            CBLWarn(Sync, @"%@ Websocket is already closing. ignoring the close now", self);
```
pasin, line 617: I think this is not a warning message. Can you change it to VERBOSE or INFO?

Objective-C/Internal/Replicator/CBLWebSocket.mm
```diff
@@ -379,7 +381,14 @@ - (void) _sendWebSocketRequest {
     if (sessionCookie.buf)
         [cookies appendFormat: @"%@;", sessionCookie.asNSString()];
     
-    NSString* cookie = [_db getCookies: _remoteURL];
+    NSError* error = nil;
+    NSString* cookie = [_db getCookies: _remoteURL error: &error];
+    if (error) {
+        // in case database is not open: CBL-2657
```
pasin, line 387: Can you add a warning log with error here?

pasin (approved):

Objective-C/Internal/Replicator/CBLWebSocket.mm
```diff
@@ -603,6 +613,15 @@ - (void) closeWithCode: (C4WebSocketCloseCode)code reason: (NSString*)reason {
 
 // Closes the connection and passes the NSError (if any) to LiteCore.
 - (void) closeWithError: (NSError*)error {
+    CBL_LOCK(self) {
```
pasin, line 616: Sorry I just noticed this. I think there is no need to lock because closeWithError: is always called inside a queue.

jayahariv: I can see many flows, some of which are not through queue. Not sure whether they go further back starting from queue.

jayahariv: yea, I got this. will update now

PR #2941, opened 2022-02-17
CBL-2825: Fix: [lithium] Attachment missing when doc updated

jayahariv:
* get the attachment present flag check with this call as well c4doc_dictContainsBlobs()
* add a SG unit test, to validate the fix. Its not useful for lithium


pasin (approved):
I have never tried to reproduce the issue with db-to-db replication. It could be good to try that to see the behavior.

jayahariv:
It is not reproducible with db-to-db replication. Even though the flag is not send, the document will be pulled successfully, also blob is present too. That's why we keep it as SG test.

pasin:
Please tag 3.0.0 and merge the PR.

PR #2967, opened 2022-06-14
Collection Indexable functions

jayahariv:
* Add collection index functions & unit tests to validate. 
* Add collection management functions & unit tests to validate.
* CBLCollection internal constructor updated. 
* Kept database property in Scope & Collection. 
* Code cleanup - header include ordered

### DIFF
--- a/Objective-C/CBLCollection.mm
+++ b/Objective-C/CBLCollection.mm
@@ -21,44 +21,138 @@
 #import "CBLIndexable.h"
 #import "CBLChangeListenerToken.h"
 #import "CBLCollectionChangeObservable.h"
+#import "CBLDatabase+Internal.h"
+#import "CBLStringBytes.h"
+#import "CBLScope.h"
+#import "CBLIndexConfiguration+Internal.h"
+#import "CBLStatus.h"
+
+using namespace fleece;
 
 NSString* const kCBLDefaultCollectionName = @"_default";
 
-@implementation CBLCollection
+@implementation CBLCollection {
+    __weak CBLDatabase* _db;
+    C4Collection* _c4col;
+}
 
 @synthesize count=_count, name=_name, scope=_scope;
 
-- (instancetype) initWithName: (NSString*)name
-                        scope: (nullable CBLScope*)scope
-                        error: (NSError**)error {
-    CBLAssertNotNil(name);
+- (instancetype) initWithDB: (CBLDatabase*)db
+             collectionName: (NSString*)collectionName
+                  scopeName: (NSString*)scopeName
+                      error: (NSError**)error {
+    CBLAssertNotNil(db);
+    CBLAssertNotNil(collectionName);
     
     self = [super init];
     if (self) {
-        _name = name;
-        if (scope)
-            _scope = scope;
+        _db = db;
+        _name = collectionName;
+        
+        C4CollectionSpec spec = { .scope = kFLSliceNull };
+        CBLStringBytes cName(collectionName);
+        spec.name = cName;
+        if (scopeName.length > 0) {
+            CBLStringBytes sName(scopeName);
+            spec.scope = sName;
+        }
+        
+        __block C4Collection* c;
+        __block C4Error err = {};
+        [db safeBlock:^{
+            c = c4db_getCollection(db.c4db, spec);
+            if (!c) {
+                // collection doesn't exists, create one
+                c = c4db_createCollection(db.c4db, spec, &err);
+            }
+        }];
+        
+        if (err.code != 0) {
+            // TODO: Add to CBLErrorMessage : CBL-3296
+            CBLWarn(Database, @"%@ Failed to create new collection (%d/%d)",
+                    self, err.code, err.domain);
+            convertError(err, error);
+            return nil;
+        }
+        
+        _c4col = c;
     }
     return self;
 }
 
+- (void) dealloc {
+    _db = nil;
+    _c4col = nil;
+}
+
+- (NSString*) description {
+    return [NSString stringWithFormat: @"%@[%@.%@]%@", self.class, _scope.name, _name, _db];
+}
+
+#pragma mark - Indexable
+
 - (BOOL) createIndexWithName: (NSString*)name
                       config: (CBLIndexConfiguration *)config
                        error: (NSError**)error {
-    // TODO: add implementation
-    return NO;
+    CBLAssertNotNil(name);
+    CBLAssertNotNil(config);
+    
+    CBLDatabase* db = _db;
+    __block BOOL success = NO;
+    __block C4Error c4err;
+    [db safeBlock: ^{
+        CBLStringBytes iName(name);
+        CBLStringBytes c4IndexSpec(config.getIndexSpecs);
+        C4IndexOptions options = config.indexOptions;
+        
+        success = c4coll_createIndex(_c4col, iName, c4IndexSpec, config.queryLanguage,
+                                          config.indexType, &options, &c4err);
+    }];
+    
+    return success || convertError(c4err, error);
 }
 
 - (BOOL) deleteIndexWithName: (NSString*)name
                        error: (NSError**)error {
-    // TODO: add implementation
-    return NO;
+    CBLAssertNotNil(name);
+    
+    CBLDatabase* db = _db;
+    __block BOOL success = NO;
+    __block C4Error c4err;
+    [db safeBlock: ^{
+        CBLStringBytes iName(name);
+        success = c4coll_deleteIndex(_c4col, iName, &c4err);
+    }];
+    
+    return success || convertError(c4err, error);
 }
 
 - (nullable NSArray*) indexes: (NSError**)error {
-    // TODO: add implementation
-    return [NSArray array];
+    CBLDatabase* db = _db;
+    __block NSMutableArray* ins = nil;
+    [db safeBlock: ^{
+        [self collectionIsValid];
+        
+        C4Error err = {};
+        C4SliceResult res = c4coll_getIndexesInfo(_c4col, &err);
+        FLDoc doc = FLDoc_FromResultData(res, kFLTrusted, nullptr, nullslice);
+        FLSliceResult_Release(res);
+        
+        NSArray* indexes = FLValue_GetNSObject(FLDoc_GetRoot(doc), nullptr);
+        FLDoc_Release(doc);
+        
+        // extract only names
+        ins = [NSMutableArray arrayWithCapacity: indexes.count];
+        for (NSDictionary* dict in indexes) {
+            [ins addObject: dict[@"name"]];
+        }
+    }];
+    
+    return [NSArray arrayWithArray: ins];
 }
+
+#pragma mark - Document Change Listeners
 
 - (id<CBLListenerToken>) addDocumentChangeListenerWithID: (NSString*)listenerID
                                                 listener: (void (^)(CBLDocumentChange*))listener {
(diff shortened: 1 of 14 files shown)


Objective-C/CBLCollection.mm
```diff
@@ -21,45 +21,139 @@
...
+#import "CBLStatus.h"
+#import "CBLScope+Internal.h"
+
+using namespace fleece;
 
 NSString* const kCBLDefaultCollectionName = @"_default";
 
-@implementation CBLCollection
+@implementation CBLCollection {
+    __weak CBLDatabase* _db;
```
pasin, line 36: This doesn't need to fix in this PR. The question is that how do you handle the case the CBLDatabase* pointer is dangle from the case that the database is released but the user still keep the collection object? When that happen, anything that uses _db in CBLCollection class will cause the crash including [db safeBlock:]. This is something that needs to be handled.

Objective-C/CBLCollection.mm
```diff
@@ -21,45 +21,139 @@
...
+        CBLStringBytes sName(scopeName);
+        C4CollectionSpec spec = { .name = cName, .scope = sName };
+        
+        __block C4Collection* c;
+        __block C4Error err = {};
+        [db safeBlock:^{
+            c = c4db_getCollection(db.c4db, spec);
+            if (!c) {
+                // collection doesn't exists, create one
+                c = c4db_createCollection(db.c4db, spec, &err);
```
pasin, line 65: 1. This logic is not correct for the case that you want to get the collection but don't want to create.
2. For the case of create, calling c4db_createCollection() is enough.

I would suggest that changing the signature to something as below, and let the CBLDatabase gives the c4collection.

`(instancetype) initWithDB: (CBLDatabase*)db collection: (C4Collection*)c4col error: (NSError**)error`.

Objective-C/CBLCollection.mm
```diff
@@ -21,45 +21,139 @@
...
                        error: (NSError**)error {
-    // TODO: add implementation
-    return NO;
+    CBLAssertNotNil(name);
+    CBLAssertNotNil(config);
+    
+    CBLDatabase* db = _db;
+    __block BOOL success = NO;
+    __block C4Error c4err;
+    [db safeBlock: ^{
```
pasin, line 104: You would need to take care two things:
1. When the db pointer is not valid anymore.
2. When database is closed. So check Open is still needed.

pasin: Probably checking if the collection is valid might be enough? I think there is a PR in LiteCore that tries to handle this but not at the current master.

Objective-C/CBLCollection.mm
pasin: Noted that if CBLDatabase doesn't cache any CBLCollections, you could retain db in CBLCollection.

Objective-C/CBLScope.m
```diff
@@ -19,25 +19,31 @@
 
 #import "CBLScope+Internal.h"
 #import "CBLCollection+Internal.h"
+#import "CBLDatabase.h"
 
 NSString* const kCBLDefaultScopeName = @"_default";
 
-@implementation CBLScope
+@implementation CBLScope {
+    __weak CBLDatabase* _db;
```
pasin, line 27: Same comment as CBLCollection.

pasin (changes requested):

Objective-C/CBLCollection.mm
jayahariv: CBLDatabase is not the only class which uses the [CBLCollection alloc] code, but CBLScope as well. So I guess, passing the collectionName and ScopeName can make it common NSString -> Slice conversion and everything inside CBLCollection itself.

Objective-C/CBLDatabase.mm
```diff
@@ -722,45 +722,92 @@ - (nullable CBLScope*) scopeWithName: (NSString*)name error: (NSError**)error {
...
+    return [NSArray arrayWithArray: collections];
 }
 
-- (nullable CBLCollection*) createCollectionWithName: (NSString*)name
-                                      scope: (nullable NSString*)scope
-                                      error: (NSError**)error {
+- (nullable CBLCollection*) createCollectionWithName: (NSString*)collectionName
+                                               scope: (nullable NSString*)scopeName
+                                               error: (NSError**)error {
+    scopeName = scopeName.length > 0 ? scopeName : kCBLDefaultScopeName;
```
pasin, line 755: Just check nil scopeName for the default scope. If using nil, LiteCore will return an error when creating the collection.

Objective-C/CBLDatabase.mm
```diff
@@ -722,45 +722,92 @@ - (nullable CBLScope*) scopeWithName: (NSString*)name error: (NSError**)error {
...
-    // TODO: add implementation
+    C4Collection* c4collection;
+    CBL_LOCK(self) {
+        [self mustBeOpen];
+        
+        C4Error c4err = {};
+        c4collection = c4db_createCollection(_c4db, spec, &c4err);
+        if (!c4collection) {
+            // TODO: handle error
+            return nil;
```
pasin, line 768: Let handle the error to complete the implementation.

Objective-C/CBLCollection.mm
```diff
@@ -17,49 +17,122 @@
...
+@synthesize count=_count, name=_name, scope=_scope, db=_db;
 
-- (instancetype) initWithName: (NSString*)name
-                        scope: (nullable CBLScope*)scope
-                        error: (NSError**)error {
-    CBLAssertNotNil(name);
-    
+- (instancetype) initWithDB: (CBLDatabase*)db
+               c4collection: (C4Collection*)c4collection
+                      error: (NSError**)error {
```
pasin, line 44: The error is not needed here. I have the same comment for the CBLScope's init as it is only for passing to the CBLScope's init.

Objective-C/CBLDatabase.mm
```diff
@@ -722,45 +722,92 @@ - (nullable CBLScope*) scopeWithName: (NSString*)name error: (NSError**)error {
...
                             scope: (nullable NSString*)scope
                             error: (NSError**)error {
-    // TODO: add implementation
-    return NO;
+    NSString* scopeName = scope ?: kCBLDefaultScopeName;
+    CBLStringBytes cName(name);
+    CBLStringBytes sName(scopeName);
+    C4CollectionSpec spec = { .name = cName, .scope = sName };
+    CBL_LOCK(self) {
+        C4Error c4err = {};
```
pasin, line 808: Need to check if the collection is valid before doing anything as well.

Objective-C/CBLCollection.mm
jayahariv: collectionIsValid check done!

Objective-C/CBLDatabase.mm
jayahariv: does this mean, we have to get the c4collection from collectionSpec and then check validity and then delete? which is not very efficient right? instead LC can do the check, whether it exists, if not throw error? 

Or can we guarantee, getting a collection means, it exists and is valid. No need to further check for isValid?

jayahariv: I see `c4db_hasCollection` API

jayahariv:
Changed the `[db safeBlock:...]` code to directly use the `CBL_LOCK(db) { ... }`. Like you said, it started becoming messy, too many variables just collect the value to-from block.

Objective-C/Tests/DatabaseTest.m
```diff
@@ -2421,22 +2421,159 @@ - (void) testDeleteIndex {
...
+// TODO: remove when implementation and unit tests are done
+// This is generic test to make sure, Collection APIs are working fine for QE.
 - (void) testCollection {
     [self createDocs: 10];
     CBLDocument* doc1 = [self.db documentWithID: $sprintf(@"doc_%03d", 1)];
     
     NSError* error = nil;
-    CBLDatabase* db = [[CBLDatabase alloc] initWithName: @"dbName" error: &error];
-    CBLCollection* c = [db createCollectionWithName: @"collection1" scope: nil error: &error];
+    CBLCollection* c = [self.db createCollectionWithName: @"collection1" scope: nil error: &error];
```
pasin, line 2566: What is the purpose of createDocs and get doc1 above?

Objective-C/CBLCollection.mm
```diff
@@ -17,49 +17,128 @@
...
 - (BOOL) createIndexWithName: (NSString*)name
                       config: (CBLIndexConfiguration *)config
                        error: (NSError**)error {
-    // TODO: add implementation
-    return NO;
+    CBLAssertNotNil(name);
+    CBLAssertNotNil(config);
+    
+    CBLDatabase* db = _db;
+    CBL_LOCK(db) {
```
pasin, line 79: I think it is probably time to change the way we do the lock but let's finish CBLCollection and CBLScope first.

I think we could switch to use NSLock / NSRecursiveLock. Another option is to use LiteCore's AccessLock.

pasin (changes requested):

Objective-C/Tests/DatabaseTest.m
```diff
@@ -2421,22 +2421,159 @@ - (void) testDeleteIndex {
...
+    AssertEqual(collections.count, 1);
+    AssertEqualObjects(collections[0].name, @"collection2");
+}
+
+- (void) testEmptyCollection {
+    AssertNil([self.db collectionWithName: @"dummy" scope: nil error: nil]);
+    AssertNil([self.db collectionWithName: @"dummy" scope: kCBLDefaultScopeName error: nil]);
+    AssertNil([self.db collectionWithName: @"dummy" scope: @"scope1" error: nil]);
+    
+    // TODO: When error is handled, check error here
```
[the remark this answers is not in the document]
jayahariv: will update, since we are returning error.

Objective-C/Tests/DatabaseTest.m
jayahariv: Sorry, this doc1 is used in below `[collection deleteDocument..]` functions we can remove when we are done with that.

Objective-C/CBLCollection.mm
jayahariv: https://issues.couchbase.com/browse/CBL-3308

Objective-C/CBLDatabase.mm
```diff
@@ -722,45 +726,131 @@ - (nullable CBLScope*) scopeWithName: (NSString*)name error: (NSError**)error {
...
+        if (![self mustBeOpen: error])
+            return nil;
+        
+        C4Error c4err = {};
+        FLMutableArray list = c4db_collectionNames(_c4db, sName, &c4err);
+        if (c4err.code != 0) {
+            CBLWarn(Database, @"%@ Failed to get collection names: %@ (%d/%d)",
+                    self, scopeName, c4err.domain, c4err.code);
+            convertError(c4err, error);
+            return nil;
```
pasin, line 756: Just notice this. The `list` needs to be released.

Objective-C/CBLDatabase.mm
```diff
@@ -722,45 +726,131 @@ - (nullable CBLScope*) scopeWithName: (NSString*)name error: (NSError**)error {
...
+            NSString* name = FLValue_GetNSObject(FLArray_Get(list, i), nullptr);
+            
+            NSError* err = nil;
+            CBLCollection* c = [self collectionWithName: name scope: scopeName error: &err];
+            if (!c) {
+                CBLWarn(Database, @"%@ Failed to get collection: %@.%@ error: %@",
+                        self, scopeName, name, err);
+                if (error)
+                    *error = err;
+                return nil;
```
pasin, line 772: Just notice this. The `list` needs to be released.

Objective-C/CBLDatabase.mm
```diff
@@ -722,45 +726,131 @@ - (nullable CBLScope*) scopeWithName: (NSString*)name error: (NSError**)error {
...
+            return nil;
+        }
+        
+        if (!c)
+            return nil;
+    }
     
-    CBLScope* s = [[CBLScope alloc] initWithName: scope ?: kCBLDefaultScopeName error: nil];
-    return [[CBLCollection alloc] initWithName: name scope: s error: nil];
+    return [[CBLCollection alloc] initWithDB: self c4collection: c];
```
pasin, line 836: Just notice this, should be returned inside the lock to be able to retain `c` before the other thread could cause `c` to be freed on the other thread (remember that `c` is cached in LiteCore and could be released anytime).

pasin (changes requested):
Notice a few more things. After fixing that, feel free to merge.

PR #2968, opened 2022-06-16
Document manage

jayahariv:


Objective-C/CBLCollection.mm
```diff
@@ -169,26 +170,101 @@ - (BOOL) deleteDocument: (CBLDocument*)document
     return NO;
 }
 
-- (CBLDocument*) documentWithID: (NSString*)docID error: (NSError**)error {
-    // TODO: add implementation
-    return nil;
+- (CBLDocument*) documentWithID: (NSString*)documentID error: (NSError**)error {
+    CBL_LOCK(self) {
```
pasin, line 174: lock database not self.

Objective-C/CBLCollection.mm
```diff
@@ -169,26 +170,101 @@ - (BOOL) deleteDocument: (CBLDocument*)document
...
+    }
 }
 
-- (NSDate*) getDocumentExpirationWithID: (NSString*)docID error: (NSError**)error {
-    // TODO: add implementation
-    return nil;
+- (NSDate*) getDocumentExpirationWithID: (NSString*)documentID error: (NSError**)error {
+    CBLAssertNotNil(documentID);
+    
+    CBL_LOCK(self) {
```
pasin, line 197: lock database

Objective-C/CBLCollection.mm
```diff
@@ -169,26 +170,101 @@ - (BOOL) deleteDocument: (CBLDocument*)document
...
+    }
 }
 
 - (BOOL) purgeDocument: (CBLDocument*)document
                  error: (NSError**)error {
-    // TODO: add implementation
-    return NO;
+    CBLAssertNotNil(document);
+    
+    CBL_LOCK(self) {
```
pasin, line 220: lock database

Objective-C/CBLCollection.mm
```diff
@@ -169,26 +170,101 @@ - (BOOL) deleteDocument: (CBLDocument*)document
     return NO;
 }
 
-- (CBLDocument*) documentWithID: (NSString*)docID error: (NSError**)error {
-    // TODO: add implementation
-    return nil;
+- (CBLDocument*) documentWithID: (NSString*)documentID error: (NSError**)error {
```
pasin, line 173: assert documentID not nil

Objective-C/CBLCollection.mm
```diff
@@ -169,26 +170,101 @@ - (BOOL) deleteDocument: (CBLDocument*)document
...
 }
 
 - (BOOL) purgeDocumentWithID: (NSString*)documentID
                        error: (NSError**)error {
-    // TODO: add implementation
-    return NO;
+    CBLAssertNotNil(documentID);
+    
+    CBL_LOCK(_db) {
+        if ([_db isClosedLocked]) {
```
pasin, line 248: As db is locked, use isClosed.

Objective-C/CBLDocument.mm
```diff
@@ -127,6 +128,16 @@ - (instancetype) initWithDatabase: (CBLDatabase*)database
     return self;
 }
 
+- (instancetype) initWithCollection: (CBLCollection*)collection
+                         documentID: (NSString*)documentID
+                     includeDeleted: (BOOL)includeDeleted
+                              error: (NSError**)outError {
+    return [self initWithDatabase: collection.db
```
pasin, line 135: The collection need to be retained in the document and the database used in the CBLDocument will need to get from the collection. This means that any constructors that has database as a parameter should be changed to constructor.

Objective-C/Internal/CBLDatabase+Internal.h
```diff
@@ -76,6 +78,9 @@ NS_ASSUME_NONNULL_BEGIN
 - (nullable NSString*) getCookies: (NSURL*)url error: (NSError**)error;
 - (BOOL) saveCookie: (NSString*)cookie url: (NSURL*)url;
 
+// Gets called from CBLCollection as well. Must be called inside a lock
+- (BOOL) prepareDocument: (CBLDocument*)document error: (NSError**)error;
```
pasin, line 82: Can you double check if this method is used only from CBLCollection? If so, please move it to CBLCollection. My impression is that CBLDatabase class will not use this anymore as it will call into CBLCollection via the default collection.

Objective-C/Tests/DatabaseTest.m
```diff
@@ -2640,6 +2640,176 @@ - (void) testCollection {
     
 }
 
+#pragma mark - GET Document
```
pasin, line 2643: As database is not the data store anymore, I think it is better to have all of the documents test else where such as in DocumentTest or a new CollectionDocumentTest.

pasin (changes requested):

Objective-C/CBLDocument.mm
pasin: > should be changed to constructor.

should be changed to collection.

Objective-C/CBLCollection.mm
jayahariv: will keep it simple, move to DocumentTest for now.

Objective-C/CBLCollection.mm
jayahariv: I guess, we make the change in separate PR. If we move all calls from Database to use the Collection API, which might make this a huge PR. I will add a TODO to do same.

Objective-C/Internal/CBLDatabase+Internal.h
jayahariv: I thought I added a comment.
I have added a comment to remove this function, once we redirect the calls to Collection.

jayahariv:
my bad, I have squashed the master merge as well, now if you check the commit alone, it include the  listener change as well.

jayahariv:
Sorry this PR is out of hands, let me refactor before reviewing.

jayahariv:
I am going to close this PR, since its too big. Refactored and made two separate ones #2973 #2974

PR #2972, opened 2022-06-24
CBL-3309: Collection change listener objc 

jayahariv:
* Add delegate to listener token class and CBLDatabase/CBLCollection/CBLReplicator/CBLQuery classes now implements it.
* now removeToken function does calls the removeChangeListener callback.

### DIFF
--- a/Objective-C/CBLCollection.mm
+++ b/Objective-C/CBLCollection.mm
@@ -18,7 +18,9 @@
 //
 
 #import "CBLChangeListenerToken.h"
+#import "CBLChangeNotifier.h"
 #import "CBLCollection+Internal.h"
+#import "CBLCollectionChange.h"
 #import "CBLCollectionChangeObservable.h"
 #import "CBLCoreBridge.h"
 #import "CBLDatabase+Internal.h"
@@ -35,9 +37,13 @@ NSString* const kCBLDefaultCollectionName = @"_default";
 
 @implementation CBLCollection {
     C4Collection* _c4col;
+    
+    C4DatabaseObserver* _colObs;
+    CBLChangeNotifier<CBLCollectionChange*>* _colChangeNotifier;
 }
 
 @synthesize count=_count, name=_name, scope=_scope, db=_db;
+@synthesize dispatchQueue=_dispatchQueue;
 
 - (instancetype) initWithDB: (CBLDatabase*)db
                c4collection: (C4Collection*)c4collection {
@@ -51,6 +57,9 @@ NSString* const kCBLDefaultCollectionName = @"_default";
         _c4col = c4coll_retain(c4collection);
         _name = slice2string(spec.name);
         _scope = [[CBLScope alloc] initWithDB: db name: slice2string(spec.scope)];
+        
+        NSString* qName = $sprintf(@"Collection <%@>", _name);
+        _dispatchQueue = dispatch_queue_create(qName.UTF8String, DISPATCH_QUEUE_SERIAL);
         
         CBLLogVerbose(Database, @"%@ Creating collection:%@ db=%@ scope=%@",
                       self, _name, db, _scope);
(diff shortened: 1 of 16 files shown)


Objective-C/CBLCollection.mm
```diff
@@ -244,4 +263,69 @@ - (BOOL) collectionIsValid: (NSError**)error {
...
+                [docIDs addObject: docID];
+            }
+            c4dbobs_releaseChanges(changes, obs.numChanges);
+        } while(obs.numChanges > 0);
+    }
+}
+
+- (void) removeToken: (id)token {
+    CBL_LOCK(self) {
+        if (![self collectionIsValid: nil])
```
pasin, line 318: I think this check might not be needed. Will it cause any problems if remove the check?

Swift/Collection.swift
```diff
@@ -244,7 +244,7 @@ public final class Collection : CollectionChangeObservable, Indexable {
     /// To remove the listener, call remove() function on the returned listener token
     ///.
     /// If the collection is deleted or the database is closed, a warning message will be logged.
-    public func addChangeListener(listener: @escaping (CollectionChange) -> Void) -> ListenerToken {
+    public func addChangeListener(listener: @escaping (CollectionChange) -> Void) -> ListenerToken? {
```
pasin, line 247: The returned ListenerToken is non-null.

Swift/Collection.swift
```diff
@@ -254,11 +254,16 @@ public final class Collection : CollectionChangeObservable, Indexable {
     ///
     /// If the collection is deleted or the database is closed, a warning message will be logged.
     public func addChangeListener(queue: DispatchQueue?,
-                                  listener: @escaping (CollectionChange) -> Void) -> ListenerToken {
+                                  listener: @escaping (CollectionChange) -> Void) -> ListenerToken? {
```
pasin, line 257: The returned ListenerToken is non-null.

Swift/CollectionChangeObservable.swift
```diff
@@ -24,13 +24,15 @@ import Foundation
 public protocol CollectionChangeObservable {
     /// Add a change listener to listen to change events occurring to any documents in the collection.
     /// To remove the listener, call remove() function on the returned listener token.
-    /// Throw Illegal State Exception or equivalent if the default collection doesn’t exist.
-    func addChangeListener(listener: @escaping (CollectionChange) -> Void) -> ListenerToken
+    ///
+    /// If the collection is deleted or the database is closed, a warning message will be logged.
+    func addChangeListener(listener: @escaping (CollectionChange) -> Void) -> ListenerToken?
```
pasin, line 29: The returned ListenerToken is non-null.

Swift/CollectionChangeObservable.swift
```diff
@@ -24,13 +24,15 @@ import Foundation
...
     
     /// Add a change listener to listen to change events occurring to any documents in the collection.
     /// If a dispatch queue is given, the events will be posted on the dispatch queue.
     /// To remove the listener, call remove() function on the returned listener token.
-    /// Throw Illegal State Exception or equivalent if the default collection doesn’t exist.
+    ///
+    /// If the collection is deleted or the database is closed, a warning message will be logged.
     func addChangeListener(queue: DispatchQueue?,
-                           listener: @escaping (CollectionChange) -> Void) -> ListenerToken
+                           listener: @escaping (CollectionChange) -> Void) -> ListenerToken?
```
pasin, line 37: The returned ListenerToken is non-null.

pasin (changes requested):

Objective-C/CBLCollection.mm
```diff
@@ -52,6 +58,9 @@ - (instancetype) initWithDB: (CBLDatabase*)db
         _name = slice2string(spec.name);
         _scope = [[CBLScope alloc] initWithDB: db name: slice2string(spec.scope)];
         
+        NSString* qName = $sprintf(@"Collection <%@>", _name);
```
pasin, line 61: Will it be OK if I have two instances of the same collection?

pasin (changes requested):

Objective-C/CBLCollection.mm
jayahariv: yes. I will update to include the self address as well. 

I copied this code from database, which means the same issue(non unique GCD queue can be created) can happen in Database as well .

Swift/Collection.swift
pasin: It still returns the ListenerToken but there will be no notifications. It's the same behavior when holding the ListenerToken but the database is closed later.

Objective-C/CBLCollection.mm
jayahariv: right, there shouldn't be. we don't work with c4collection here.

pasin (approved):

PR #2973, opened 2022-07-01
Collection cached + Doc Expiry

jayahariv:
* Default collection cached
* Document expiry functions added
* Redirect the Database -> Collections - indexes, doc expiry

### DIFF
--- a/Objective-C/CBLDatabase.mm
+++ b/Objective-C/CBLDatabase.mm
@@ -91,7 +91,6 @@ typedef enum {
     CBLDatabaseState _state;
     
     CBLCollection* _defaultCollection;
-    BOOL _defaultCollectionIsDeleted;
 }
 
 @synthesize name=_name;
@@ -146,7 +145,13 @@ static void dbObserverCallback(C4DatabaseObserver* obs, void* context) {
         
         _state = kCBLDatabaseStateOpened;
         
-        _defaultCollectionIsDeleted = NO;
+        NSError* error = nil;
+        _defaultCollection = [self collectionWithName: kCBLDefaultCollectionName
+                                                scope: kCBLDefaultScopeName
+                                                error: &error];
+        if (error) {
+            CBLWarn(Database, @"%@ : Error getting the default collection: %@", self, error);
+        }
     }
     return self;
 }
@@ -611,7 +616,10 @@ static void dbObserverCallback(C4DatabaseObserver* obs, void* context) {
 #pragma mark - Index:
 
 - (NSArray<NSString*>*) indexes {
-    return [[self mustDefaultCollection: nil] indexes: nil];
+    NSError *error = nil;
+    NSArray *res = [[self defaultCollectionOrThrow] indexes: &error];
+    throwIfNotOpenError(error);
+    return res;
 }
 
 - (BOOL) createIndex: (CBLIndex*)index withName: (NSString*)name error: (NSError**)error {
@@ -624,14 +632,23 @@ static void dbObserverCallback(C4DatabaseObserver* obs, void* context) {
 }
 
 - (BOOL) createIndex: (NSString*)name withConfig: (id<CBLIndexSpec>)config error: (NSError**)error {
-    return [[self mustDefaultCollection: error] createIndexWithName: name
-                                                             config: config
-                                                              error: error];
+    NSError *e = nil;
+    BOOL res = [[self defaultCollectionOrThrow] createIndexWithName: name config: config error: &e];
+    if (error)
+        *error = e;
+    
+    throwIfNotOpenError(e);
+    return res;
 }
 
 - (BOOL) deleteIndexForName: (NSString*)name error: (NSError**)outError {
-    return [[self mustDefaultCollection: outError] deleteIndexWithName: name
-                                                                 error: outError];
+    NSError *e = nil;
+    BOOL res = [[self defaultCollectionOrThrow] deleteIndexWithName: name error: &e];
+    if (outError)
+        *outError = e;
+    
+    throwIfNotOpenError(e);
+    return res;
 }
 
 #pragma mark - DOCUMENT EXPIRATION
@@ -639,14 +656,23 @@ static void dbObserverCallback(C4DatabaseObserver* obs, void* context) {
 - (BOOL) setDocumentExpirationWithID: (NSString*)documentID
                           expiration: (nullable NSDate*)date
                                error: (NSError**)error {
-    return [[self mustDefaultCollection: error] setDocumentExpirationWithID: documentID
+    NSError *e = nil;
+    BOOL res = [[self defaultCollectionOrThrow] setDocumentExpirationWithID: documentID
                                                                  expiration: date
-                                                                      error: error];
+                                                                      error: &e];
+    if (error)
+        *error = e;
+    
+    throwIfNotOpenError(e);
+    return res;
 }
 
 - (nullable NSDate*) getDocumentExpirationWithID: (NSString*)documentID {
-    return [[self mustDefaultCollection: nil] getDocumentExpirationWithID: documentID
-                                                                    error: nil];
+    NSError *error = nil;
+    NSDate* res = [[self defaultCollectionOrThrow] getDocumentExpirationWithID: documentID
+                                                                         error: &error];
+    throwIfNotOpenError(error);
+    return res;
 }
 
 #pragma mark - Query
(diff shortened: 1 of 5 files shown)


Objective-C/Internal/CBLDatabase+Internal.h
```diff
@@ -37,6 +37,8 @@
 #import "CBLURLEndpointListener.h"
 #endif
 
+#define msec 1000.0
```
pasin, line 40: Is this required?

Objective-C/CBLDatabase.mm
```diff
@@ -638,65 +624,29 @@ - (BOOL) createIndexWithConfig: (CBLIndexConfiguration*)config
...
-    CBLAssertNotNil(documentID);
-    
-    CBL_LOCK(self) {
-        UInt64 timestamp = date ? (UInt64)(date.timeIntervalSince1970*msec) : 0;
-        C4Error err;
-        CBLStringBytes docID(documentID);
-        return c4doc_setExpiration(_c4db, docID, timestamp, &err) || convertError(err, error);
-    }
+                               error: (NSError**)error {
+    return [[self mustDefaultCollection: error] setDocumentExpirationWithID: documentID
```
pasin, line 642: It also needs to check the error if the error is NotOpen or not. If it's NotOpen, an exception needs to be thrown.

Objective-C/CBLDatabase.mm
```diff
@@ -638,65 +624,29 @@ - (BOOL) createIndexWithConfig: (CBLIndexConfiguration*)config
...
-    
-    CBL_LOCK(self) {
-        CBLStringBytes docID(documentID);
-        UInt64 timestamp = c4doc_getExpiration(_c4db, docID, nullptr);
-        if (timestamp == 0) {
-            return nil;
-        }
-        return [NSDate dateWithTimeIntervalSince1970: (timestamp/msec)];
-    }
+    return [[self mustDefaultCollection: nil] getDocumentExpirationWithID: documentID
```
pasin, line 648: It also needs to check the error if the error is NotOpen or not. If it's NotOpen, an exception needs to be thrown.

Objective-C/CBLDatabase.mm
```diff
@@ -1214,6 +1176,18 @@ - (void) safeBlock:(void (^)())block {
...
+    CBLCollection* c = [self defaultCollection: &e];
+    if (!c) {
+        [NSException raise: NSInternalInconsistencyException
+                    format: @"Attempt to perform an operation on removed collection %@", e];
+        if (outError)
+            *outError = e;
+    }
+    return c;
+}
+
```
pasin, line 1190: How about this?

```
- (instancetype) initWithName: (NSString*)name
                       config: (nullable CBLDatabaseConfiguration*)config
                        error: (NSError**)outError {
    ...
    if (self) {
        ....
        NSError* error;
        _defaultCollection = [self collectionWithName: kCBLDefaultCollectionName scope: kCBLDefaultScopeName error: &error];
        if (error) { 
           CBLWarn(Database, @"%@ : Error getting the default collection: %@", self, error); 
       }
    }
    return self;
}

- (nullable CBLCollection*) defaultCollection: (NSError**)error {
   CBL_LOCK(self) {
       if (![self mustBeOpen: error]) {
           return nil;
       }
       return _defaultCollection.isValid ? _defaultCollection : nil;
   }
}

- (CBLCollection*) defaultCollectionOrThrow {
    if (!_defaultCollection )
        throwNotOpen();
    return _defaultCollection; 
}

- (nullable NSDate*) getDocumentExpirationWithID: (NSString*)documentID {
    NSError * error;
    NSDate* exp = [[self defaultCollectionOrThrow] getDocumentExpirationWithID: documentID error: error]];
    throwIfNotOpenError(error);
    return exp;
}

static void throwNotOpen() {
    [NSException raise: NSInternalInconsistencyException format: @"The database was closed, or the default collection was deleted."];
}

static void throwIfNotOpenError(NSError* error) {
    if (error && error.domain == CBLErrorDomain && error.code == CBLErrorNotOpen) {
        throwNotOpen();
    }
}
```

pasin (changes requested):

Objective-C/Internal/CBLDatabase+Internal.h
jayahariv: I guess, no. we can move to CBLCollection.m, since we have already redirected the calls.

Objective-C/CBLDatabase.mm
```diff
@@ -1206,6 +1166,8 @@ - (void) freeC4DB {
     _c4db = nil;
     
     _state = kCBLDatabaseStateClosed;
+    
+    _defaultCollection = nil;
```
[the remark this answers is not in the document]
jayahariv: https://issues.couchbase.com/browse/CBL-3367

Objective-C/CBLDatabase.mm
pasin: Note : defaultCollectionOrThrow() doesn't check if the collection is valid because it will be checked when calling any [CBLCollection XXX] method, and a big plus, the method doesn't require locking.

Objective-C/CBLDatabase.mm
```diff
@@ -624,29 +632,47 @@ - (BOOL) createIndexWithConfig: (CBLIndexConfiguration*)config
 }
 
 - (BOOL) createIndex: (NSString*)name withConfig: (id<CBLIndexSpec>)config error: (NSError**)error {
-    return [[self mustDefaultCollection: error] createIndexWithName: name
-                                                             config: config
-                                                              error: error];
+    NSError *e = nil;
+    BOOL res = [[self defaultCollectionOrThrow] createIndexWithName: name config: config error: &e];
+    if (error)
```
pasin, line 637: Better to do this after `throwIfNotOpenError(e)` so if the exception is thrown, the error will not be set.

Objective-C/CBLDatabase.mm
```diff
@@ -624,29 +632,47 @@ - (BOOL) createIndexWithConfig: (CBLIndexConfiguration*)config
...
                           expiration: (nullable NSDate*)date
                                error: (NSError**)error {
-    return [[self mustDefaultCollection: error] setDocumentExpirationWithID: documentID
+    NSError *e = nil;
+    BOOL res = [[self defaultCollectionOrThrow] setDocumentExpirationWithID: documentID
                                                                  expiration: date
-                                                                      error: error];
+                                                                      error: &e];
+    if (error)
+        *error = e;
```
pasin, line 664: Same as above, do this after throwIfNotOpenError(e)

pasin (changes requested):

Objective-C/Internal/CBLDatabase+Internal.h
```diff
@@ -77,6 +77,8 @@ NS_ASSUME_NONNULL_BEGIN
 - (nullable NSString*) getCookies: (NSURL*)url error: (NSError**)error;
 - (BOOL) saveCookie: (NSString*)cookie url: (NSURL*)url;
 
+- (CBLCollection*) defaultCollectionOrThrow;
```
pasin, line 80: There is no need to declare it here unless there are somewhere else beside the database is using.

Objective-C/CBLDatabase.mm
```diff
@@ -144,6 +144,15 @@ - (instancetype) initWithName: (NSString*)name
...
         
         _state = kCBLDatabaseStateOpened;
+        
+        C4Error c4error = {};
+        C4Collection* c4col = c4db_getDefaultCollection(_c4db, &c4error);
+        if (c4error.code != 0) {
+            CBLWarn(Database, @"%@ : Error getting the default collection: %d/%d",
+                    self, c4error.domain, c4error.code);
+        }
+
```
pasin, line 154: Please remove extra line so the code can be group together.

Objective-C/CBLDatabase.mm
```diff
@@ -144,6 +144,15 @@ - (instancetype) initWithName: (NSString*)name
...
         _state = kCBLDatabaseStateOpened;
+        
+        C4Error c4error = {};
+        C4Collection* c4col = c4db_getDefaultCollection(_c4db, &c4error);
+        if (c4error.code != 0) {
+            CBLWarn(Database, @"%@ : Error getting the default collection: %d/%d",
+                    self, c4error.domain, c4error.code);
+        }
+        
+        _defaultCollection = [[CBLCollection alloc] initWithDB: self c4collection: c4col];
```
pasin, line 155: The reason that this solves the problem because it doesn't put the returned collection into the main auto releasepool of the test which will be drained after the teardown. I don't have a good clean solution for this. If you want to use the original code, you can do it this way. I'm fine either way.

```
- (void) initDefaultCollection {
    // Note: initialize _defaultCollection under an auto releasepool to prevent the test failure
    // when checking the c4object leaks in teardown. Depending on where the database object
    // is initialized, the returned collection could be put into the main auto releasepool of
    // the test, which is drained after the teardown.
    @autoreleasepool {
        NSError* error = nil;
        _defaultCollection = [self collectionWithName: kCBLDefaultCollectionName
                                                scope: kCBLDefaultScopeName
                                                error: &error];
        if (error) {
            CBLWarn(Database, @"%@ : Error when getting the default collection: %@", self, error);
        }
    }
}
```

pasin (changes requested):

PR #2976, opened 2022-07-06
swift collection unit tests

jayahariv:
* made changes to make collections API public and useful in unit tests
* add unit tests for collection management, index, change event.


Swift/DatabaseConfiguration.swift
```diff
@@ -57,4 +57,10 @@ public struct DatabaseConfiguration {
         return config
     }
     
+    init(_ impl: CBLDatabaseConfiguration) {
+        self.directory = impl.directory
+        #if COUCHBASE_ENTERPRISE
+        // TODO: handle encryption-key
```
jayahariv, line 63:
```suggestion
        // TODO: handle encryption-key: https://issues.couchbase.com/browse/CBL-3372
```

Swift/Database.swift
```diff
@@ -547,6 +544,11 @@ public final class Database {
 
     let _impl: CBLDatabase
     
+    init(_ impl: CBLDatabase) {
```
pasin, line 547: If Database is passed when creating a collection, I think we don't need this.

Swift/DatabaseConfiguration.swift
```diff
@@ -57,4 +57,10 @@ public struct DatabaseConfiguration {
         return config
     }
     
+    init(_ impl: CBLDatabaseConfiguration) {
```
pasin, line 60: Same as above, if database is passed, when creating a collection, I think this is not needed.

Swift/Document.swift
```diff
@@ -47,7 +47,7 @@ public class Document : DictionaryProtocol, Equatable, Hashable, Sequence {
             return nil
         }
         
-        return Collection(impl: c)
+        return Collection(c)
```
pasin, line 50: Can the collection be passed in when creating a Document?

Swift/Collection.swift
```diff
@@ -58,21 +58,21 @@ public final class Collection : CollectionChangeObservable, Indexable {
     public static let defaultCollectionName: String = "_default"
     
     /// Collection name.
-    var name: String { _impl.name }
+    public var name: String { _impl.name }
     
     /// The scope of the collection.
-    var scope: Scope { Scope(_impl.scope) }
+    public var scope: Scope { Scope(_impl.scope) }
```
pasin, line 64: You can keep the scope inside the Collection so that it will not create a new Scope every time.

pasin (changes requested):

Swift/Collection.swift
```diff
@@ -283,11 +282,15 @@ public final class Collection : CollectionChangeObservable, Indexable {
     
     // MARK: Internal
     
-    init(impl: CBLCollection) {
+    var db: Database {
+        return _db
+    }
+    
+    init(_ impl: CBLCollection) {
```
pasin, line 289: Can the database be passed so that the database will be the same instance where the collection is created or is gotten?

pasin (changes requested):

jayahariv:
Rebased here.

pasin (changes requested):
For Database, I don't see the benefit of using the default collection as it will be just a simple wrap (no need to handle exception as well). I think it's much simple to call to CBLDatabase like the original code that you did.

Swift/Database.swift
```diff
@@ -165,16 +152,20 @@ public final class Database {
         _ document: MutableDocument, conflictHandler: @escaping (MutableDocument, Document?) -> Bool
     ) throws -> Bool {
         do {
-            return try defaultCollectionOrThrowEx().save(document: document,
-                                                         conflictHandler: conflictHandler)
+            try _impl.save(
+                document._impl as! CBLMutableDocument,
+                conflictHandler: { (cur: CBLMutableDocument, old: CBLDocument?) -> Bool in
+                    let col = try? self.defaultCollection()
```
pasin, line 158: Same as above, need to throw NotOpen error when default collection is nil.

Swift/ReplicatorConfiguration.swift
```diff
@@ -322,7 +322,7 @@ public struct ReplicatorConfiguration {
         
         if let resolver = self.conflictResolver {
             c.setConflictResolverUsing { (conflict) -> CBLDocument? in
-                return resolver.resolve(conflict: Conflict(impl: conflict))?._impl
+                return resolver.resolve(conflict: Conflict(impl: conflict, db: self.database))?._impl
```
pasin, line 325: The idea is to get the collection from the replicator's config and pass to the Conflict here.

Swift/Scope.swift
```diff
@@ -59,5 +59,5 @@ public final class Scope {
     }
     
     let _impl: CBLScope
-    weak var _db: Database?
+    let _db: Database?
```
pasin, line 62: Is there the case the database is nil?

Swift/Database.swift
```diff
@@ -515,14 +470,8 @@ public final class Database {
...
-        do {
-            let _ = try _impl.defaultCollection()
-        } catch let err as NSError {
-            _defaultCollection = nil
-            throw err
-        }
-        
-        return _defaultCollection
+        let c = try _impl.defaultCollection()
+        return Collection(c, db: self)
```
pasin, line 474: It would be nice if we can cache the default collection but that will cause the circular reference that will require to bring back the weak database in Collection and Scope. Let's do this at least for now.

pasin (changes requested):

Swift/Collection.swift
```diff
@@ -231,11 +231,17 @@ public final class Collection : CollectionChangeObservable, Indexable {
...
     ///
     /// If the collection is deleted or the database is closed, a warning message will be logged.
-    func addDocumentChangeListener(id: String, queue: DispatchQueue?,
+    public func addDocumentChangeListener(id: String, queue: DispatchQueue?,
                                    listener: @escaping (DocumentChange) -> Void) -> ListenerToken {
         let token = _impl.addDocumentChangeListener(withID: id, queue: queue) {
             [unowned self] (change) in
-            listener(DocumentChange(database: _db, documentID: change.documentID, collection: self))
+            if let db = _db {
+                listener(DocumentChange(database: db,
```
pasin, line 239: DocumentChange also needs collection.

pasin (changes requested):

jayahariv:
rebased to master here

Swift/Collection.swift
```diff
@@ -299,6 +299,6 @@ public final class Collection : CollectionChangeObservable, Indexable {
     }
     
     let _impl: CBLCollection
-    weak var _db: Database?
+    let _db: Database?
```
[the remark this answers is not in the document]
jayahariv: its made to optional, because of the weak type

Swift/Database.swift
```diff
@@ -85,12 +82,8 @@ public final class Database {
     /// Gets a Document object with the given ID.
     @available(*, deprecated, message: "Use database.defaultCollection().document(withID:) instead.")
     public func document(withID id: String) -> Document? {
-        do {
-            return try defaultCollectionOrThrowEx().document(id: id)
-        } catch let err as NSError {
-            if err.code == CBLError.notOpen && err.domain == CBLErrorDomain {
-                Database.throwNotOpenEx()
-            }
+        if let implDoc = _impl.document(withID: id), let c = try? _impl.defaultCollection() {
```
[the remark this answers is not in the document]
jayahariv: We cannot throw the error, but we can do the NSException. Since the document(withID:...) is not throwable.

Swift/Database.swift
jayahariv: will throw exception

Swift/Scope.swift
jayahariv: no need, it was from the previous weak reference.

Swift/Conflict.swift
```diff
@@ -32,24 +32,28 @@ public struct Conflict {
         guard let doc = toImpl().localDocument else {
             return nil
         }
-        return Document(doc)
+        
+        return Document(doc,collection: col)
```
pasin, line 36: [Format] space after comma.

pasin (approved):

PR #2981, opened 2022-07-15
feat: Replicator Config Collection Change

jayahariv:
* Implement the collection support changes
* Add all unit tests.


Objective-C/CBLReplicatorConfiguration.m
```diff
@@ -89,8 +99,7 @@ - (instancetype) initWithTarget:(id<CBLEndpoint>)target {
     self = [super init];
     if (self) {
         _target = target;
-        
-        // TODO: Add implementation
+        _collectionConfigs = [NSMutableDictionary dictionary];
```
pasin, line 102: Need to initialize the same properties as initWithDatabase:target does. Try to extract them out in a method.

Objective-C/CBLReplicatorConfiguration.m
```diff
@@ -139,16 +148,29 @@ - (void) setNetworkInterface: (NSString*)networkInterface {
 - (void) setDocumentIDs: (NSArray<NSString *>*)documentIDs {
     [self checkReadonly];
     _documentIDs = documentIDs;
+    _defaultCollectionConfig.documentIDs = documentIDs;
```
pasin, line 151: The documentID needs to link with the one in the _collectionConfigs. And if there is no default config in the dictionary, an exception will be thrown.

```
- (void) setDocumentIDs: (NSArray<NSString *>*)documentIDs {
    auto config = _collectionConfigs[_defaultCollection]; // Need to make sure that the collection can be key (e.g. implement hash and etc).
    if (!config) {
        // Throw exception
    }
    config.documentIDs = documentIDs
}

- (NSArray<NSString *>*) getDocumentIDs {
    auto config = _collectionConfigs[_defaultCollection]; // Need to make sure that the collection can be key (e.g. 
    return config.documentIDs;
}
```

Objective-C/CBLReplicatorConfiguration.m
```diff
@@ -139,16 +148,29 @@ - (void) setNetworkInterface: (NSString*)networkInterface {
...
 }
 
 - (void) setChannels: (NSArray<NSString *>*)channels {
     [self checkReadonly];
     _channels = channels;
+    _defaultCollectionConfig.channels = channels;
```
pasin, line 157: Same as documenIDs

Objective-C/CBLReplicatorConfiguration.m
```diff
@@ -139,16 +148,29 @@ - (void) setNetworkInterface: (NSString*)networkInterface {
...
 }
 
 - (void) setConflictResolver: (id<CBLConflictResolver>)conflictResolver {
     [self checkReadonly];
     _conflictResolver = conflictResolver;
+    _defaultCollectionConfig.conflictResolver = conflictResolver;
```
pasin, line 163: Same as documenIDs

Objective-C/CBLReplicatorConfiguration.m
```diff
@@ -189,31 +211,95 @@ - (void) setEnableAutoPurge: (BOOL)enableAutoPurge {
...
-    // TODO: Add implementation
-
+    if (_collections.count > 0) {
+        if (_collections[0].db != collection.db) {
+            [NSException raise: NSInvalidArgumentException
+                        format: @"Attempt to add collection from different databases"];
+        }
+    }
+    
+    CBLCollection* defaultCollection = [_database defaultCollectionOrThrow];
```
pasin, line 235: It shouldn't throw if default collection doesn't exist.

Objective-C/CBLReplicatorConfiguration.m
```diff
@@ -189,31 +211,95 @@ - (void) setEnableAutoPurge: (BOOL)enableAutoPurge {
...
+    
+    CBLCollectionConfiguration* colConfig = nil;
+    if (config)
+        colConfig = [[CBLCollectionConfiguration alloc] initWithConfig: config];
+    else
+        colConfig = [[CBLCollectionConfiguration alloc] init];
+    
+    [_collectionConfigs setObject: colConfig forKey: collection];
+    
+    _collections = _collectionConfigs.allKeys;
```
pasin, line 247: This shouldn't be set every time. Create a getter method for this.

Objective-C/CBLReplicatorConfiguration.m
```diff
@@ -189,31 +211,95 @@ - (void) setEnableAutoPurge: (BOOL)enableAutoPurge {
...
+        
+        if (db != col.db) {
+            [NSException raise: NSInvalidArgumentException
+                        format: @"Attempt to add collections from different databases"];
+        }
+    }
+    
+    CBLCollection* defaultCollection = [_database defaultCollectionOrThrow];
+    for (CBLCollection* col in collections) {
+        if (col == defaultCollection)
```
pasin, line 269: Call addCollection instead of duplicating the code here.

Objective-C/CBLReplicatorConfiguration.m
```diff
@@ -189,31 +211,95 @@ - (void) setEnableAutoPurge: (BOOL)enableAutoPurge {
...
+    if (config.pullFilter != _pullFilter)
+        _pullFilter = config.pullFilter;
 }
 
 - (void) removeCollection:(CBLCollection *)collection {
-    
-    // TODO: Add implementation
-    
+    [_collectionConfigs removeObjectForKey: collection];
+    _collections = _collectionConfigs.allKeys;
```
pasin, line 298: Same as above, add a getter for _collections instead of updating it every operation.

Objective-C/CBLReplicatorConfiguration.m
```diff
@@ -189,31 +211,95 @@ - (void) setEnableAutoPurge: (BOOL)enableAutoPurge {
...
+            colConfig = [[CBLCollectionConfiguration alloc] initWithConfig: config];
+        else
+            colConfig = [[CBLCollectionConfiguration alloc] init];
+        
+        [_collectionConfigs setObject: colConfig forKey: col];
+    }
+    _collections = _collectionConfigs.allKeys;
+}
+
+- (void) checkAndUpdateConfig: (CBLCollectionConfiguration*)config {
```
pasin, line 283: This method is not needed.

Objective-C/CBLReplicatorConfiguration.m
```diff
@@ -189,31 +211,95 @@ - (void) setEnableAutoPurge: (BOOL)enableAutoPurge {
...
+    if (_collections.count > 0) {
+        if (_collections[0].db != collection.db) {
+            [NSException raise: NSInvalidArgumentException
+                        format: @"Attempt to add collection from different databases"];
+        }
+    }
+    
+    CBLCollection* defaultCollection = [_database defaultCollectionOrThrow];
+    if (collection == defaultCollection)
+        [self checkAndUpdateConfig: config];
```
pasin, line 237: There is no need to do this.

Objective-C/CBLReplicatorConfiguration.m
pasin: Then, there is no need to have _documentIDs as well.

Objective-C/CBLReplicatorConfiguration.m
```diff
@@ -189,31 +211,95 @@ - (void) setEnableAutoPurge: (BOOL)enableAutoPurge {
...
+    
+    _collections = _collectionConfigs.allKeys;
 }
 
-- (void) addCollections: (NSArray*)collections
+- (void) addCollections: (NSArray<CBLCollection*>*)collections
                  config: (nullable CBLCollectionConfiguration*)config {
-
-    // TODO: Add implementation
+    Assert(collections.count > 0);
```
pasin, line 252: Should throw invalid argument exception instead.

Objective-C/CBLReplicatorConfiguration.m
```diff
@@ -189,31 +211,95 @@ - (void) setEnableAutoPurge: (BOOL)enableAutoPurge {
     _enableAutoPurge = enableAutoPurge;
 }
 
+- (CBLDatabase*) database {
+    if (!_database)
+        [NSException raise: NSInternalInconsistencyException
+                    format: @"Attempt to access database property, with initWithTarget:"];
```
pasin, line 217: This will only happen when all collections are removed from the collection. You can change the message to something like "Attempt to access database property but no collections added."

Objective-C/CBLReplicatorConfiguration.m
```diff
@@ -189,31 +211,95 @@ - (void) setEnableAutoPurge: (BOOL)enableAutoPurge {
...
+    for (CBLCollection* col in collections) {
+        if (!col.isValid) {
+            [NSException raise: NSInvalidArgumentException
+                        format: @"Attempt to add an invalid collection"];
+        }
+        
+        if (db != col.db) {
+            [NSException raise: NSInvalidArgumentException
+                        format: @"Attempt to add collections from different databases"];
+        }
```
pasin, line 264: Need to update _database if it's nil as well.

Objective-C/CBLReplicatorConfiguration.m
pasin: Need to reset the _database if no collections.

pasin (changes requested):

Objective-C/CBLReplicatorConfiguration.m
```diff
@@ -145,32 +148,61 @@ - (void) setNetworkInterface: (NSString*)networkInterface {
...
+- (CBLCollectionConfiguration*) defaultCollectionConfig {
+    CBLCollection* defaultCollection = [_database defaultCollectionOrThrow];
+    return _collectionConfigs[defaultCollection];
+}
+
+- (CBLCollectionConfiguration*) defaultCollectionConfigOrThrow {
+    CBLCollectionConfiguration* config = [self defaultCollectionConfig];
+    if (!config) {
+        [NSException raise: NSInternalInconsistencyException
+                    format: @"Attempt to perform action with empty default collection config"];
```
pasin, line 160: It is not "empty default collection config". Maybe "No default collection added to the configuration."

Objective-C/CBLReplicatorConfiguration.m
```diff
@@ -232,9 +264,8 @@ - (void) addCollection: (CBLCollection*)collection
         }
     }
     
-    CBLCollection* defaultCollection = [_database defaultCollectionOrThrow];
-    if (collection == defaultCollection)
-        [self checkAndUpdateConfig: config];
+    if (!_database)
+        _database = collection.db;
```
pasin, line 268: I think it would be more safe to also check if _database is nil or not. If it's not nil, throw an exception as well.

Objective-C/CBLReplicatorConfiguration.m
```diff
@@ -323,12 +328,14 @@ - (instancetype) initWithConfig: (CBLReplicatorConfiguration*)config
...
-        _pullFilter = config.pullFilter;
-        _conflictResolver = config.conflictResolver;
+        [self initCollectionConfigs];
+        
+        CBLCollectionConfiguration* colConfig = [self defaultCollectionConfigOrThrow];
+        colConfig.documentIDs = config.documentIDs;
+        colConfig.channels = config.channels;
+        colConfig.pushFilter = config.pushFilter;
+        colConfig.pullFilter = config.pullFilter;
+        colConfig.conflictResolver = config.conflictResolver;
```
pasin, line 338: colConfig is not used and I'm not sure what is the purpose of doing this.

Objective-C/CBLReplicatorConfiguration.m
```diff
@@ -323,12 +328,14 @@ - (instancetype) initWithConfig: (CBLReplicatorConfiguration*)config
 // TODO: Remove https://issues.couchbase.com/browse/CBL-3206
 #pragma clang diagnostic push
 #pragma clang diagnostic ignored "-Wdeprecated-declarations"
-        _database = config.database;
-        _documentIDs = config.documentIDs;
-        _channels = config.channels;
-        _pushFilter = config.pushFilter;
-        _pullFilter = config.pullFilter;
-        _conflictResolver = config.conflictResolver;
+        [self initCollectionConfigs];
```
pasin, line 331: I'm not sure if this is required. I think that it needs is to copy _collectionConfigs from the other config and verify if the collections are valid && database is not nil.

pasin (changes requested):

jayahariv:
rebased here

Objective-C/CBLReplicatorConfiguration.m
jayahariv: but we are not setting the _database  property in case of initWithTarget:, we link it when user calls the addCollection:config

Objective-C/CBLReplicatorConfiguration.m
jayahariv: yes, right. But why do we need to check database. I guess, I can verify each collection is valid, which will internally check the db open or not for that collection.

Objective-C/CBLReplicatorConfiguration.m
pasin: The assumption is that if there is a collection added, the database cannot be nil. As collection.db is weak, there is no guarantee that it will not be null even after calling collection.isValid() (technically). 

Also, I just noticed that _collections is only populated when collection property is called. I think we actually do not need _collections at all.

Also, the below logic is not correct. I should check against _database if it is not nil. If it's nil, it will just set _database.

```
if (_collections.count > 0) {
        if (_collections[0].db != collection.db) {
            [NSException raise: NSInvalidArgumentException
                        format: @"Attempt to add collection from different databases"];
        }
    }
    
    if (!_database)
        _database = collection.db;
```

pasin (changes requested):
I replied to your comment.

Objective-C/CBLReplicatorConfiguration.m
jayahariv: simplified

Objective-C/CBLReplicatorConfiguration.m
```diff
@@ -256,17 +255,16 @@ - (void) addCollection: (CBLCollection*)collection
         [NSException raise: NSInvalidArgumentException
                     format: @"Attempt to add an invalid collection"];
     }
-    
-    if (_collections.count > 0) {
-        if (_collections[0].db != collection.db) {
+    CBLDatabase* colDB = collection.db;
+    if (_database) {
```
pasin, line 259: Technically, colDB could be nil at this point as well.

```
CBLDatabase* colDB = collection.db;
if (!colDB) {
    [NSException raise: NSInvalidArgumentException
                    format: @"Attempt to add an invalid collection"];
}
```

Objective-C/CBLReplicatorConfiguration.m
```diff
@@ -289,15 +287,14 @@ - (void) addCollections: (NSArray<CBLCollection*>*)collections
...
-    return _collections;
+    return _collectionConfigs.allKeys;
 }
 
 - (void) removeCollection:(CBLCollection *)collection {
     [_collectionConfigs removeObjectForKey: collection];
     
     // reset the database, when all collections are removed
-    if (_collections.count == 0) {
+    if (self.collections.count == 0) {
```
pasin, line 297: It's simpler by checking _collectionConfigs.count.

```
if (_collectionConfigs.count == 0)
```

pasin (approved):

Objective-C/CBLReplicatorConfiguration.m
jayahariv: ok

Objective-C/CBLReplicatorConfiguration.m
jayahariv:
```
CBLDatabase* colDB = collection.db;
    if (!collection.isValid || !colDB) {
        [NSException raise: NSInvalidArgumentException
                    format: @"Attempt to add an invalid collection"];
    }
    ```

PR #2983, opened 2022-07-19
feat: pending document ids with collection

jayahariv:
* Add implementation to include the pending doc ids implementation with collection
* Update unit tests to validate both methods

### DIFF
--- a/Objective-C/CBLReplicator.mm
+++ b/Objective-C/CBLReplicator.mm
@@ -39,6 +39,7 @@
 #import "CBLStringBytes.h"
 #import "CBLErrorMessage.h"
 #import "CBLLockable.h"
+#import "CBLScope.h"
 
 #import "c4Replicator.h"
 #import "c4Socket.h"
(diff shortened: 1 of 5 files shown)


Objective-C/CBLReplicator.mm
```diff
@@ -471,14 +479,26 @@ - (void) setProgressLevel: (CBLReplicatorProgressLevel)level {
 - (BOOL) isDocumentPending: (NSString *)documentID
                 collection: (CBLCollection *)collection
                      error: (NSError**)error {
+    CBLAssertNotNil(collection);
     
-    // TODO: Add implementation 
+    CBLStringBytes name(collection.name);
+    CBLStringBytes scopeName(collection.scope.name);
+    C4CollectionSpec c4spec = { .name = name, .scope = scopeName };
```
pasin, line 486: Optionally, you could made an internal properly to get the collection spec from the collection object (e.g. collection.spec).

Objective-C/CBLReplicator.mm
```diff
@@ -471,14 +479,26 @@ - (void) setProgressLevel: (CBLReplicatorProgressLevel)level {
...
 }
 
 - (BOOL) isDocumentPending: (NSString*)documentID error: (NSError**)error {
     CBLAssertNotNil(documentID);
+    
+    return [self isDocumentPending: documentID c4Spec: kC4DefaultCollectionSpec error: error];
+}
+
+// internal
+- (BOOL) isDocumentPending:(NSString *)documentID
```
pasin, line 498: Why doesn't this get implemented in `- (BOOL) isDocumentPending: (NSString*)documentID collection: error: (NSError**)error` itself?

Objective-C/CBLReplicator.mm
```diff
@@ -423,13 +424,21 @@ - (void) setProgressLevel: (CBLReplicatorProgressLevel)level {
...
-    return nil;
+    return [self pendingDocumentIDsForCollectionSpec: c4spec error: error];
 }
 
 - (NSSet<NSString*>*) pendingDocumentIDs: (NSError**)error {
+    return [self pendingDocumentIDsForCollectionSpec: kC4DefaultCollectionSpec error: error];
+}
+
+// internal
+- (NSSet<NSString*>*) pendingDocumentIDsForCollectionSpec: (C4CollectionSpec)spec error: (NSError**)error {
```
pasin, line 441: Why doesn't this get implemented in `- (NSSet<NSString*>*) pendingDocumentIDsForCollection: (CBLCollection *)collection error: (NSError**)error` itself? Then the old pendingDocumentIDs will be called into the new one.

pasin (changes requested):

Objective-C/CBLReplicator.mm
jayahariv: it was to avoid getting the default collection but use the default collection spec.

pasin (approved):

PR #2987, opened 2022-07-25
[feat] [objc] replicator collection LC API update

jayahariv:
* Collection configs are passed to ReplicatorParams
* Filter & ConflictResolver, get the collection if LC returns collection-spec, or use defaultCollection.

### DIFF
--- a/Objective-C/CBLReplicator.mm
+++ b/Objective-C/CBLReplicator.mm
@@ -18,6 +18,7 @@
 //
 
 #import "CBLReplicator+Backgrounding.h"
+#import "CBLCollectionConfiguration+Internal.h"
 #import "CBLCollection+Internal.h"
 #import "CBLDocumentReplication+Internal.h"
 #import "CBLReplicator+Internal.h"
@@ -190,7 +191,7 @@ typedef enum {
 - (bool) _setupC4Replicator: (C4Error*)outErr {
     if (_repl) {
         // If _repl exists, just update the options dict, in case -resetCheckpoint has been called:
-        c4repl_setOptions(_repl, self._encodedOptions);
+        c4repl_setOptions(_repl, [self encodedOptions: _config.effectiveOptions]);
         return true;
     }
 
@@ -242,7 +243,7 @@ typedef enum {
     socketFactory.context = (__bridge void*)self;
 
     // Parameters:
-    alloc_slice optionsFleece = self._encodedOptions;
+    alloc_slice optionsFleece = [self encodedOptions: _config.effectiveOptions];
     C4ReplicatorParameters params = {
         .push = mkmode(isPush(_config.replicatorType), _config.continuous),
         .pull = mkmode(isPull(_config.replicatorType), _config.continuous),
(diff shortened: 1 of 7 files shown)


Objective-C/CBLReplicator.mm
```diff
@@ -261,6 +264,26 @@ - (bool) _setupC4Replicator: (C4Error*)outErr {
...
+    NSUInteger collectionCount = _config.collectionConfigs.count;
+    C4ReplicationCollection cols[collectionCount];
+    NSUInteger i = 0;
+    for (CBLCollection* col in _config.collectionConfigs) {
+        CBLCollectionConfiguration* colConfig = _config.collectionConfigs[col];
+        AllocedDict dict = [self encodedCollectionOptions: colConfig];
+        C4ReplicationCollection c = {
+            .collection = col.c4spec,
+            .push = mkmode(isPush(_config.replicatorType), _config.continuous),
+            .pull = mkmode(isPull(_config.replicatorType), _config.continuous),
```
pasin, line 276: The same .push and .pull setup here should be removed as the default collection when using database should be in the collectionConfigs as well:

https://github.com/couchbase/couchbase-lite-ios/pull/2987/files#diff-db9ac2c53f3d61c425b37311410aa8be9d6fcafb43e203cd6235c63b18455b0bR250-R251

Objective-C/CBLReplicator.mm
```diff
@@ -699,9 +729,21 @@ - (void) resolveConflict: (CBLReplicatedDocument*)doc {
 - (void) _resolveConflict: (CBLReplicatedDocument*)doc {
     CBLLogInfo(Sync, @"%@: Resolve conflicting version of '%@'", self, doc.id);
     NSError* error = nil;
-    if (![[_config.database defaultCollectionOrThrow] resolveConflictInDocument: doc.id
-                                                           withConflictResolver: _config.conflictResolver
-                                                                          error: &error]) {
+    CBLCollection* collection = nil;
+    if (doc.collection) {
+        collection = [_config.database collectionWithName: doc.collection scope: doc.scope error: &error];
```
pasin, line 734: The collection should be gotten from _config.collectionConfigs. I think you could set callbackContext in C4ReplicationCollection with the CBLCollection object so that you could get the CBLCollection object back from CBLReplicatedDocument.

https://github.com/couchbase/couchbase-lite-core/blob/master/C/include/c4ReplicatorTypes.h#L118

Objective-C/CBLReplicator.mm
```diff
@@ -730,26 +772,43 @@ static bool pushFilter(C4CollectionSpec collectionSpec,
...
                 pushing: (bool)pushing
 {
-    CBLCollection* c = [_config.database defaultCollectionOrThrow];
-    auto doc = [[CBLDocument alloc] initWithCollection: c
+    CBLCollection* collection = nil;
+    if (c4spec.name.buf) {
+        NSError* error = nil;
+        NSString* name = slice2string(c4spec.name);
+        NSString* scopeName = slice2string(c4spec.scope);
+        collection = [_config.database collectionWithName: name
```
pasin, line 799: This should be gotten from the _config.collectionConfigs. For the conflict resolver, I suggested to set callbackContext in C4ReplicationCollection with the Collection object. However, when setting that the context for the filter could be the collection object as well which might not be working here (check with LiteCore or test it yourself). 

If there is conflict of setting the context object, another option would be lazily creating a dictionary of full collection name (< scope > . < collection >) and the CBLCollection object out of _config.collectionConfig in order to get to the CBLCollection object form the c4spec.

pasin (changes requested):

Objective-C/CBLReplicator.mm
jayahariv: but the 'CBLCollectionConfiguration' class doesn't include the ReplicatorMode(push/pull) in it, and we are not copying the values as well.

jayahariv: ok. I think I get it, I will remove the C4ReplicatorParameters.push/pull setting, as the default collection config also is present here.

Objective-C/CBLReplicator.mm
jayahariv: Setting C4ReplicationCollection is not making any difference. `onDocumentsEnded`, is still returning the replicator we set at 'C4ReplicatorParameters.callbackContext'. I guess, that param is not used currently.

Objective-C/CBLReplicator.mm
```diff
@@ -276,7 +274,7 @@ - (bool) _setupC4Replicator: (C4Error*)outErr {
             .pull = mkmode(isPull(_config.replicatorType), _config.continuous),
             .pushFilter = filter(_config.pushFilter, true),
             .pullFilter = filter(_config.pullFilter, false),
-            .callbackContext    = (__bridge void*)self,
+            .callbackContext    = (__bridge void*)col,
```
pasin, line 277: As you are using collection map, I think you can just set it to self. Also, once LiteCore starts to send back this context, you will not be able to get to the replicator object to call the filter function.

pasin (changes requested):

pasin (approved):
Change that one bit back to self. I think it's ready to go.

PR #2988, opened 2022-07-26
Fix podspecs

jeffdgr8:
Spaces were changed to underscores in #2654, which broke the local podspec files. This fixes the scheme names and output framework paths.

Also .gitignore the framework builds and logs, as well as the AppCode .idea IDE configuration folder.


biozal:
@pasin @rajagp

PR #2989, opened 2022-07-27
[test] [swift] 8.1 default collection/scope & 8.2 collections

jayahariv:
* Add unit tests for 8.1 & 8.2
* CBL-3477

Update the code to return nil with CBLErrorNotFound, in case of collection/scope is missing. So that,  swift can check and return nil. 

If not we might end up using a hack to check for 
```
if nsError.code == 0, nsError.domain == "Foundation._GenericObjCError" {
        // handle empty error code! 
}

### DIFF
--- a/Objective-C/CBLDatabase.mm
+++ b/Objective-C/CBLDatabase.mm
@@ -681,12 +681,12 @@ static const C4DatabaseConfig2 kDBConfig = {
 
 - (nullable CBLCollection*) defaultCollection: (NSError**)error {
     CBL_LOCK(_mutex) {
-        // db closed or deleted
+        // db closed
         if (![self mustBeOpen: error])
             return nil;
         
-        // collection removed
-        BOOL isValid = _defaultCollection.isValid;
+        // if collection is invalid (deleted or LC marked as invalid)
+        BOOL isValid = [_defaultCollection collectionIsValid: error];
         if (!isValid) {
             _defaultCollection = nil;
             convertError({LiteCoreDomain, kC4ErrorNotFound}, error);
--- a/Objective-C/CBLDatabase.mm
+++ b/Objective-C/CBLDatabase.mm
@@ -686,7 +686,7 @@ static const C4DatabaseConfig2 kDBConfig = {
             return nil;
         
         // if collection is invalid (deleted or LC marked as invalid)
-        BOOL isValid = [_defaultCollection collectionIsValid: error];
+        BOOL isValid = _defaultCollection.isValid;
         if (!isValid) {
             _defaultCollection = nil;
             convertError({LiteCoreDomain, kC4ErrorNotFound}, error);
--- a/Objective-C/CBLDatabase.mm
+++ b/Objective-C/CBLDatabase.mm
@@ -668,8 +668,10 @@ static const C4DatabaseConfig2 kDBConfig = {
         
         CBLStringBytes sname(scopeName);
         BOOL exists = c4db_hasScope(_c4db, sname);
-        if (!exists)
+        if (!exists) {
+            convertError({LiteCoreDomain, kC4ErrorNotFound}, error);
             return nil;
+        }
         
         return [[CBLScope alloc] initWithDB: self name: scopeName];
     }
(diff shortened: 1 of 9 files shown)

### CODE REVIEW

File: Objective-C/CBLDatabase.mm [hunk 1]
```diff
@@ -668,8 +668,10 @@ - (nullable CBLScope*) scopeWithName: (NSString*)name error: (NSError**)error {
         
         CBLStringBytes sname(scopeName);
         BOOL exists = c4db_hasScope(_c4db, sname);
-        if (!exists)
+        if (!exists) {
+            convertError({LiteCoreDomain, kC4ErrorNotFound}, error);
```
pasin, line 672: Ugh. This needs to mention in the API doc.

Objective-C/CBLDatabase.mm
```diff
@@ -679,12 +681,16 @@ - (nullable CBLScope*) scopeWithName: (NSString*)name error: (NSError**)error {
...
+        // db closed or deleted
         if (![self mustBeOpen: error])
             return nil;
         
+        // collection removed
         BOOL isValid = _defaultCollection.isValid;
-        if (!isValid)
+        if (!isValid) {
             _defaultCollection = nil;
+            convertError({LiteCoreDomain, kC4ErrorNotFound}, error);
```
pasin, line 692: Need to differentiate b/w null default collection and invalid default collection. If it's invalid (e.g. db closed), not open error should be returned.

Also, returning nil with CBLErrorNotFound should be mentioned in the API doc.

Objective-C/CBLDatabase.mm
```diff
@@ -802,8 +808,10 @@ - (nullable CBLCollection*) collectionWithName: (NSString*)name
             return nil;
         }
         
-        if (!c)
+        if (!c) {
+            convertError({LiteCoreDomain, kC4ErrorNotFound}, error);
```
pasin, line 812: This needs to mention in the API doc.

pasin (changes requested):
I wished there is a way to not set the NotFound error.

Objective-C/CBLDatabase.mm
jayahariv: line:685, it will return the db closed/deleted error.

jayahariv:
> I wished there is a way to not set the NotFound error.

There is a way without setting error, but the swift need to hack it like the description. 
```
if nsError.code == 0, nsError.domain == "Foundation._GenericObjCError" {
        // handle empty error code! 
}

File: Objective-C/CBLDatabase.mm (hunk 2 above)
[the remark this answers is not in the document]
Comment (jayahariv) [reply Objective-C/CBLDatabase.mm:692]: same:
line:685, it will return the db closed/deleted error.

**Comment by pasin** [comment]
I mean without using that hack as well.

File: Objective-C/CBLDatabase.mm (hunk 2 above)
Comment (jayahariv) [reply Objective-C/CBLDatabase.mm:692]: I am assuming the scenarios are: 
1. db is closed / deleted : in this case `_c4db` will be `nullptr`, `[self mustBeOpen]` function will return false with CBLErrorNotOpen. 
2. invalid collection: in this case collection is deleted or marked as invalid. In both cases, collection doesn't exist? So we return CBLErrorNotFound?

File: Objective-C/CBLDatabase.mm [hunk 4]
```diff
@@ -681,12 +681,12 @@ - (nullable CBLScope*) scopeWithName: (NSString*)name error: (NSError**)error {
 
 - (nullable CBLCollection*) defaultCollection: (NSError**)error {
     CBL_LOCK(_mutex) {
-        // db closed or deleted
+        // db closed
```
jayahariv, line 684: this check will return CBLErrorNotOpen, in case of db closed/deleted?

Objective-C/CBLDatabase.mm
```diff
@@ -681,12 +681,12 @@ - (nullable CBLScope*) scopeWithName: (NSString*)name error: (NSError**)error {
...
         if (![self mustBeOpen: error])
             return nil;
         
-        // collection removed
-        BOOL isValid = _defaultCollection.isValid;
+        // if collection is invalid (deleted or LC marked as invalid)
+        BOOL isValid = [_defaultCollection collectionIsValid: error];
```
pasin, line 689: It doesn't seem like passing error here is having any effects as it will get converted to NotFound anyway. So if it doesn't do anything at the end, just pass nil.

pasin (approved):

PR #2993, opened 2022-07-29
[test] [swift] SQL++, QueryBuilder

jayahariv:
* Add unit tests
* CBL-3480

### DIFF
--- a/Swift/Select.swift
+++ b/Swift/Select.swift
@@ -27,17 +27,9 @@ public final class Select: Query, FromRouter {
     /// - Parameter dataSource: The DataSource object.
     /// - Returns: The From object that represent the FROM clause of the query.
     public func from(_ dataSource: DataSourceProtocol) -> From {
-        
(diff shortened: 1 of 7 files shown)


Swift/Select.swift
```diff
@@ -27,9 +27,17 @@ public final class Select: Query, FromRouter {
     /// - Parameter dataSource: The DataSource object.
     /// - Returns: The From object that represent the FROM clause of the query.
     public func from(_ dataSource: DataSourceProtocol) -> From {
+        
+        var db: Database!
+        if let colSource = dataSource.source() as? Collection {
+            db = colSource._db
```
pasin, line 33: Maybe we should have a proper internal property for the _db.

pasin (approved):

Swift/Select.swift
jayahariv: I created a getter from Swift.

PR #2994, opened 2022-08-02
[feat] [test] [swift] replicator config collection update

jayahariv:
* Replicator config collection update
* CBL-3508

### DIFF
--- a/Swift/ReplicatorConfiguration.swift
+++ b/Swift/ReplicatorConfiguration.swift
@@ -59,11 +59,8 @@ public struct ReplicatorConfiguration {
     /// The local database to replicate with the replication target.
     @available(*, deprecated, message: "Use config.collections instead.")
     public var database: Database {
-        guard let db = _db else {
-            NSException(name: .internalInconsistencyException,
-                        reason: "Attempt to access database property but no collections added",
-                        userInfo: nil).raise()
-            fatalError()
+        guard let db = self.db else {
+            fatalError("Attempt to access database property but no collections added")
         }
         return db
     }
@@ -250,7 +247,7 @@ public struct ReplicatorConfiguration {
     
     /// The collections used for the replication.
     public var collections: [Collection] {
-        return Array(_collectionConfigs.keys)
+        return Array(self.collectionConfigs.keys)
     }
     
     /// Initializes a ReplicatorConfiguration's builder with the given
@@ -261,7 +258,7 @@ public struct ReplicatorConfiguration {
     ///   - target: The replication target.
     @available(*, deprecated, message: " Use init(target:) instead. ")
     public init(database: Database, target: Endpoint) {
-        self._db = database
+        self.db = database
         self.target = target
         
         initCollectionConfigs()
@@ -285,21 +282,16 @@ public struct ReplicatorConfiguration {
                               config: CollectionConfiguration? = nil) -> ReplicatorConfiguration {
         
         if !collection._impl.isValid {
-            NSException(name: .invalidArgumentException,
-                        reason: "Attempt to add an invalid collection.",
-                        userInfo: nil).raise()
-            fatalError()
+            fatalError("Attempt to add an invalid collection.")
         }
         
         let db = collection._db
-        if let db1 = _db {
+        if let db1 = self.db {
             if db1._impl != db._impl {
-                NSException(name: .invalidArgumentException,
-                            reason: "Attempt to add collection from different databases.",
-                            userInfo: nil).raise()
+                fatalError("Attempt to add collection from different databases.")
             }
         } else {
-            _db = db
+            self.db = db
         }
         
         var colConfig: CollectionConfiguration!
@@ -309,7 +301,7 @@ public struct ReplicatorConfiguration {
             colConfig = CollectionConfiguration()
         }
         
-        _collectionConfigs[collection] = colConfig
+        self.collectionConfigs[collection] = colConfig
         
         return self
     }
@@ -324,10 +316,7 @@ public struct ReplicatorConfiguration {
                                         config: CollectionConfiguration? = nil) -> ReplicatorConfiguration {
         
         if collections.count == 0 {
-            NSException(name: .invalidArgumentException,
-                        reason: "Attempt to add empty collection array",
-                        userInfo: nil).raise()
-            fatalError()
+            fatalError("Attempt to add empty collection array.")
         }
         
         for col in collections {
@@ -340,10 +329,10 @@ public struct ReplicatorConfiguration {
     /// Remove the collection. If the collection doesn’t exist, this operation will be no ops.
     @discardableResult
     public mutating func removeCollection(_ collection: Collection) -> ReplicatorConfiguration {
-        _collectionConfigs.removeValue(forKey: collection)
+        self.collectionConfigs.removeValue(forKey: collection)
         
-        if _collectionConfigs.isEmpty {
-            _db = nil
+        if self.collectionConfigs.isEmpty {
+            self.db = nil
         }
         
         return self
@@ -355,7 +344,7 @@ public struct ReplicatorConfiguration {
     /// - Parameter collection The collection whose config is needed.
     /// - Returns The collection config if exists.
     public func collectionConfig(_ collection: Collection) -> CollectionConfiguration? {
-        return _collectionConfigs[collection]
+        return self.collectionConfigs[collection]
     }
     
     /// Initializes a ReplicatorConfiguration's builder with the given
@@ -363,7 +352,7 @@ public struct ReplicatorConfiguration {
     ///
     /// - Parameter config: The configuration object.
     public init(config: ReplicatorConfiguration) {
-        self._db = config.database
+        self.db = config.database
         self.target = config.target
         self.replicatorType = config.replicatorType
         self.continuous = config.continuous
@@ -376,10 +365,12 @@ public struct ReplicatorConfiguration {
         self.maxAttemptWaitTime = config.maxAttemptWaitTime
         self.enableAutoPurge = config.enableAutoPurge
         
-        for (col, config) in config._collectionConfigs {
-            if col.isValid {
-                _collectionConfigs[col] = config
+        for (col, config) in config.collectionConfigs {
+            if !col.isValid {
+                fatalError("Tries to add invalid collection")
             }
+            
+            self.collectionConfigs[col] = config
         }
         
         #if os(iOS)
@@ -395,10 +386,10 @@ public struct ReplicatorConfiguration {
     
     var defaultCollectionConfig: CollectionConfiguration {
         guard let col = try? self.database.defaultCollection() else {
-            Database.throwNotOpenError()
+            fatalError("Default collection is missing!")
         }
         
-        guard let colConfig = _collectionConfigs[col] else {
+        guard let colConfig = self.collectionConfigs[col] else {
             fatalError("default collection config missing")
         }
         
@@ -407,10 +398,10 @@ public struct ReplicatorConfiguration {
     
     mutating func setDefaultCollectionConfig(_ config: CollectionConfiguration) {
         guard let col = try? self.database.defaultCollection() else {
-            Database.throwNotOpenError()
+            fatalError("Default collection is missing!")
         }
         
-        _collectionConfigs[col] = config
+        self.collectionConfigs[col] = config
     }
     
     func toImpl() -> CBLReplicatorConfiguration {
@@ -466,7 +457,7 @@ public struct ReplicatorConfiguration {
     }
     
     mutating func initCollectionConfigs() {
-        if let db = _db {
+        if let db = self.db {
             guard let col = try? db.defaultCollection() else {
                 Database.throwNotOpenError()
             }
@@ -476,6 +467,6 @@ public struct ReplicatorConfiguration {
         }
     }
     
-    var _collectionConfigs = [Collection: CollectionConfiguration]()
-    var _db: Database?
+    var collectionConfigs = [Collection: CollectionConfiguration]()
+    var db: Database?
 }
--- a/Swift/ReplicatorConfiguration.swift
+++ b/Swift/ReplicatorConfiguration.swift
@@ -58,7 +58,15 @@ public struct ReplicatorConfiguration {
     
     /// The local database to replicate with the replication target.
     @available(*, deprecated, message: "Use config.collections instead.")
-    public let database: Database
+    public var database: Database {
+        guard let db = _db else {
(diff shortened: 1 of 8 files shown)


Swift/ReplicatorConfiguration.swift
```diff
@@ -58,7 +58,15 @@ public struct ReplicatorConfiguration {
     
     /// The local database to replicate with the replication target.
     @available(*, deprecated, message: "Use config.collections instead.")
-    public let database: Database
+    public var database: Database {
+        guard let db = _db else {
+            NSException(name: .internalInconsistencyException,
+                        reason: "Attempt to access database property but no collections added",
+                        userInfo: nil).raise()
+            fatalError()
```
pasin, line 66: I think now we are mixing b/w using fatalError() and NSException() in Swift code. I think we should decide which one to use. If you are relying on NSException in test code, you can continue to use NSException but please change the rest of the code in Swift to all use NSException for consistency. You can decide here.

Create function to throw exception so that you don't need to call fatalError().
```
throwException(e: NSException) -> Never() {
    e.raise();
}
```

Swift/ReplicatorConfiguration.swift
```diff
@@ -221,16 +273,43 @@ public struct ReplicatorConfiguration {
...
+    @discardableResult
+    public mutating func addCollection(_ collection: Collection,
                               config: CollectionConfiguration? = nil) -> ReplicatorConfiguration {
         
-        // TODO: Add implementation
+        if !collection._impl.isValid {
+            NSException(name: .invalidArgumentException,
+                        reason: "Attempt to add an invalid collection.",
+                        userInfo: nil).raise()
+            fatalError()
```
pasin, line 291: Same as above.

Swift/ReplicatorConfiguration.swift
```diff
@@ -221,16 +273,43 @@ public struct ReplicatorConfiguration {
...
+        }
+        
+        let db = collection._db
+        if let db1 = _db {
+            if db1._impl != db._impl {
+                NSException(name: .invalidArgumentException,
```
pasin, line 297: Same as above.

Swift/ReplicatorConfiguration.swift
```diff
@@ -240,18 +319,32 @@ public struct ReplicatorConfiguration {
...
+    @discardableResult
+    public mutating func addCollections(_ collections: Array<Collection>,
+                                        config: CollectionConfiguration? = nil) -> ReplicatorConfiguration {
         
-        // TODO: Add implementation
+        if collections.count == 0 {
+            NSException(name: .invalidArgumentException,
+                        reason: "Attempt to add empty collection array",
+                        userInfo: nil).raise()
+            fatalError()
```
pasin, line 330: Same as above.

Swift/ReplicatorConfiguration.swift
```diff
@@ -262,35 +355,33 @@ public struct ReplicatorConfiguration {
...
         self.heartbeat = config.heartbeat
         self.maxAttempts = config.maxAttempts
         self.maxAttemptWaitTime = config.maxAttemptWaitTime
-        self.pullFilter = config.pullFilter
-        self.pushFilter = config.pushFilter
         self.enableAutoPurge = config.enableAutoPurge
         
+        for (col, config) in config._collectionConfigs {
+            if col.isValid {
+                _collectionConfigs[col] = config
```
pasin, line 381: Throw exception if collection is not valid.

Swift/ReplicatorConfiguration.swift
```diff
@@ -302,6 +393,26 @@ public struct ReplicatorConfiguration {
     
     // MARK: Internal
     
+    var defaultCollectionConfig: CollectionConfiguration {
+        guard let col = try? self.database.defaultCollection() else {
+            Database.throwNotOpenError()
```
pasin, line 398: No need to check here as it has already been checked in initCollectionConfigs.

Swift/ReplicatorConfiguration.swift
```diff
@@ -302,6 +393,26 @@ public struct ReplicatorConfiguration {
...
+        guard let colConfig = _collectionConfigs[col] else {
+            fatalError("default collection config missing")
+        }
+        
+        return colConfig
+    }
+    
+    mutating func setDefaultCollectionConfig(_ config: CollectionConfiguration) {
+        guard let col = try? self.database.defaultCollection() else {
+            Database.throwNotOpenError()
```
pasin, line 410: No need to check again.

Swift/ReplicatorConfiguration.swift
```diff
@@ -354,4 +465,17 @@ public struct ReplicatorConfiguration {
...
+                Database.throwNotOpenError()
+            }
+            
+            let colConfig = CollectionConfiguration()
+            addCollection(col, config: colConfig)
+        }
+    }
+    
+    var _collectionConfigs = [Collection: CollectionConfiguration]()
+    var _db: Database?
```
pasin, line 480: I think we should stop using _ at some point as it's objective-c not swift.

Swift/Tests/ReplicatorTest+Collection.swift
```diff
@@ -0,0 +1,421 @@
...
+        
+        var config = ReplicatorConfiguration(database: self.db, target: target)
+        
+        let conflictResolver = TestConflictResolver({ (con: Conflict) -> Document? in
+            return con.remoteDocument
+        })
+        config.conflictResolver = conflictResolver
+        XCTAssertNotNil(conflictResolver.uid)
+        
+
```
pasin, line 93: extra empty line.

Swift/Tests/ReplicatorTest+Collection.swift
```diff
@@ -0,0 +1,421 @@
...
+        
+        let conflictResolver = TestConflictResolver({ (con: Conflict) -> Document? in
+            return con.remoteDocument
+        })
+        config.conflictResolver = conflictResolver
+        XCTAssertNotNil(conflictResolver.uid)
+        
+        
+        var colConfig = collectionConfig(config, col: try defaultCollection())
+        XCTAssertEqual(conflictResolver.uid, (config.conflictResolver as! TestConflictResolver).uid)
```
pasin, line 95: As TestConflictResolver is a class, can you just compare b/w conflictResolver and config.conflictResolver without using uid? Basically, they will be the same object.

pasin (changes requested):

Swift/ReplicatorConfiguration.swift
jayahariv:
```
throwException(e: NSException) -> Never() {
    e.raise();
}
```

Above function return type is `Void` and not `Never`. For that either we should add `fatalError()` or `abort()`

jayahariv: I am only worried about the testing when fatalError or abort is used, instead of the NSException. I am more inclined to use fatalError, since that's more swift style.

pasin: You can decide on this.

Swift/ReplicatorConfiguration.swift
jayahariv: Even if the default collection is present when we created the replicator config, but it can be invalid when it set properties(channels, documentIDs, filters etc..) later. Also swift optionals will make us check for empty when default collection is fetched, as well as collection config.

Swift/ReplicatorConfiguration.swift
jayahariv: same? check for optional collection

Swift/ReplicatorConfiguration.swift
jayahariv: yea. we can update to avoid using underscore('_')

Swift/ReplicatorConfiguration.swift
jayahariv: will use the fatalError

Swift/Tests/ReplicatorTest+Collection.swift
jayahariv: Made the TestConflictResolver equatable, so now it can be checked for equality in swift. But its still same as checking `resolver.uid == resolver2.uid`.

Swift/Tests/ReplicatorTest+Collection.swift
```diff
@@ -163,7 +162,7 @@ class ReplicatorTest_Collection: ReplicatorTest {
         XCTAssertNotNil(colConfig.pullFilter)
     }
     
-    func testCreateConfigWithEndpointOnly() {
+    func _testCreateConfigWithEndpointOnly() {
```
pasin, line 165: Why disabled?

Swift/Tests/ReplicatorTest+Collection.swift
```diff
@@ -84,39 +84,38 @@ class ReplicatorTest_Collection: ReplicatorTest {
         
         var config = ReplicatorConfiguration(database: self.db, target: target)
         
-        let conflictResolver = TestConflictResolver({ (con: Conflict) -> Document? in
+        let conflictResolver: TestConflictResolver = TestConflictResolver({ (con: Conflict) -> Document? in
```
pasin, line 87: I'm curious why it needs explicit type `: TestConflictResolver`?

Swift/Tests/ReplicatorTest+Collection.swift
```diff
@@ -84,39 +84,38 @@ class ReplicatorTest_Collection: ReplicatorTest {
...
         })
-        XCTAssertNotEqual(conflictResolver2.uid, conflictResolver3.uid)
+        XCTAssertNotEqual(conflictResolver2, conflictResolver3)
         colConfig = collectionConfig(config, col: try defaultCollection())
         colConfig.conflictResolver = conflictResolver3
         config.addCollection(try defaultCollection(), config: colConfig)
         
-        XCTAssertEqual(conflictResolver3.uid, (config.conflictResolver as! TestConflictResolver).uid)
-        XCTAssertEqual(conflictResolver3.uid, (colConfig.conflictResolver as! TestConflictResolver).uid)
+        XCTAssertEqual(conflictResolver3, (config.conflictResolver as! TestConflictResolver))
```
pasin, line 117: `XCTAssertTrue(conflictResolver3 === config.conflictResolver)` This will compare the reference when there is no need for uid.

pasin (changes requested):

Swift/Tests/ReplicatorTest+Collection.swift
jayahariv: will remove, since the test validates the fatalError, which is not possible.

Swift/Tests/ReplicatorTest+Collection.swift
jayahariv: not necessary

Swift/Tests/ReplicatorTest+Collection.swift
```diff
@@ -84,38 +84,38 @@ class ReplicatorTest_Collection: ReplicatorTest {
...
+        let conflictResolver = TestConflictResolver({ (con: Conflict) -> Document? in
             return con.remoteDocument
         })
         config.conflictResolver = conflictResolver
         XCTAssertNotNil(conflictResolver)
         
         var colConfig = collectionConfig(config, col: try defaultCollection())
-        XCTAssertEqual(conflictResolver, (config.conflictResolver as! TestConflictResolver))
-        XCTAssertEqual(conflictResolver, (colConfig.conflictResolver as! TestConflictResolver))
+        XCTAssert(conflictResolver === (config.conflictResolver as! TestConflictResolver))
```
pasin, line 94: I don't think `(config.conflictResolver as! TestConflictResolver)` is needed? Just `conflictResolver === config.conflictResolver`.

Swift/Tests/ReplicatorTest+Collection.swift
```diff
@@ -84,38 +84,38 @@ class ReplicatorTest_Collection: ReplicatorTest {
...
             return con.localDocument
         })
-        XCTAssertNotEqual(conflictResolver, conflictResolver2)
+        XCTAssert(conflictResolver !== conflictResolver2)
         config.conflictResolver = conflictResolver2
         
-        XCTAssertEqual(conflictResolver2, (config.conflictResolver as! TestConflictResolver))
         colConfig = collectionConfig(config, col: try defaultCollection())
-        XCTAssertEqual(conflictResolver2, (colConfig.conflictResolver as! TestConflictResolver))
+        XCTAssert(conflictResolver2 === (config.conflictResolver as! TestConflictResolver))
```
pasin, line 105: Same as above.

Swift/Tests/ReplicatorTest+Collection.swift
```diff
@@ -236,13 +226,13 @@ class ReplicatorTest_Collection: ReplicatorTest {
         XCTAssertNotNil(config1.pullFilter)
         XCTAssertEqual(config1.channels, ["channel1", "channel2", "channel3"])
         XCTAssertEqual(config1.documentIDs, ["doc1", "doc2", "doc3"])
-        XCTAssertEqual((config1.conflictResolver as! TestConflictResolver), conflictResolver)
+        XCTAssert((config1.conflictResolver as! TestConflictResolver) === conflictResolver)
```
pasin, line 229: Same as above.

Swift/Tests/ReplicatorTest+Collection.swift
```diff
@@ -84,38 +84,38 @@ class ReplicatorTest_Collection: ReplicatorTest {
...
-        XCTAssertNotEqual(conflictResolver2, conflictResolver3)
+        XCTAssert(conflictResolver2 !== conflictResolver3)
         colConfig = collectionConfig(config, col: try defaultCollection())
         colConfig.conflictResolver = conflictResolver3
         config.addCollection(try defaultCollection(), config: colConfig)
         
-        XCTAssertEqual(conflictResolver3, (config.conflictResolver as! TestConflictResolver))
-        XCTAssertEqual(conflictResolver3, (colConfig.conflictResolver as! TestConflictResolver))
+        XCTAssert(conflictResolver3 === (config.conflictResolver as! TestConflictResolver))
+        XCTAssert(conflictResolver3 === (colConfig.conflictResolver as! TestConflictResolver))
```
pasin, line 118: Same as above `conflictResolver3 === config.conflictResolver`.

pasin (changes requested):

Swift/Tests/ReplicatorTest+Collection.swift
jayahariv: type difference, config.conflictResolver is type ConflictResolverProtocol. Other one is class.

pasin (approved):

PR #3006, opened 2022-08-31
[fix] remove the underscore style from swift

jayahariv:
CBL-3631


pasin (approved):
I haven't reviewed this throughly as I assumed that the PR is only removing _.

jayahariv:
Yes there isn't anything much, but only removing the variable.

PR #3011, opened 2022-09-16
[bug-fix] CBL-2973 - Xcode 13.x Update

jayahariv:
- This was working fine on earlier Xcode, since the preprocessor was going through different flow. 
- Pull request script updated to use iPhone 8. Previous SE(2nd get) was unavailable

### DIFF
--- a/Objective-C/Internal/Replicator/CBLTrustCheck.mm
+++ b/Objective-C/Internal/Replicator/CBLTrustCheck.mm
@@ -186,7 +186,7 @@ static BOOL sOnlyTrustAnchorCerts;
             NSData* certData = nil;
             for (CFIndex i = 0; i < count; i++) {
                 CFArrayRef certs = SecTrustCopyCertificateChain(_trust);
-                SecCertificateRef cert = (SecCertificateRef)CFArrayGetValueAtIndex(certs, 0);
+                SecCertificateRef cert = (SecCertificateRef)CFArrayGetValueAtIndex(certs, i);
                 certData = CFBridgingRelease(SecCertificateCopyData(cert));
                 CFRelease(certs);
                 if ([_pinnedCertData isEqual: certData]) {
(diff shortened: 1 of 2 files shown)


Objective-C/Internal/Replicator/CBLTrustCheck.mm
```diff
@@ -186,7 +186,7 @@ - (NSURLCredential*) checkTrust: (NSError**)outError {
             NSData* certData = nil;
             for (CFIndex i = 0; i < count; i++) {
                 CFArrayRef certs = SecTrustCopyCertificateChain(_trust);
-                SecCertificateRef cert = (SecCertificateRef)CFArrayGetValueAtIndex(certs, 0);
+                SecCertificateRef cert = (SecCertificateRef)CFArrayGetValueAtIndex(certs, i);
```
pasin, line 189: Does this mean that the pinned cert is still matched with the leaf cert? 

I think recently there is a question from support about this as a customer has been asked for the feature, and I have informed the support that we do support this. We may need to port the fix to 3.0.x and raise this to Priya and support?

pasin: Related CBSE-12463 (Show the customer and support who were informed about the feature)

jayahariv: I assume, this won't happen to any released product, since this flow will work only with Xcode 13+, before it was going through the below check. That's why the test was passing before updating the Xcode in build machine. 


https://github.com/couchbase/couchbase-lite-ios/blob/9e0451c2d847b79b03930544877d69f9111af318/Objective-C/Internal/Replicator/CBLTrustCheck.mm#L201

pasin (approved):

pasin:
As we have upgraded xcode in the build machine, please port the fix (or have a ticket) to the lithium branch as well.

jayahariv:
https://github.com/couchbase/couchbase-lite-ios/blob/release/lithium/Objective-C/Internal/Replicator/CBLTrustCheck.mm#L184-L190

For lithium, it's not necessary, since we haven't used the latest API in Lithium.

PR #3016, opened 2022-09-19
fixes for the Jenkins failures

jayahariv:
(1) multiple calls to the expectations  
(2) change listener called in multiple sets

3.1.0 - 235 build failures
```
3.1.0-EE-Swift

Failing tests:
	CBL_EE_Swift_Tests:
		MessageEndpointTest.testCollectionsContinuousPushPullReplication()
		MessageEndpointTest.testCollectionsSingleShotPushPullReplication()
		MessageEndpointTest.testMismatchedCollectionReplication()


Reason: Multiple calls made to expectation. 


3.1.0-CE-swift

testLiveQueryReturnsEmptyResultSet

Reason: change listener can be called in mutiple sets.  

3.1.0-EE-ObjC 
Failing tests:
	CBL_EE_ObjC_Tests:
		-[MessageEndpointTest testCollectionsContinuousPushPullReplication]
		-[MessageEndpointTest testPullDoc]
		-[MessageEndpointTest testPullDocContinuous]
		-[MessageEndpointTest testPushDoc]


Reason: Multiple calls made to expectation. 
```

### DIFF
--- a/Objective-C/Tests/MessageEndpointTest.m
+++ b/Objective-C/Tests/MessageEndpointTest.m
@@ -171,6 +171,8 @@ MCSessionDelegate, CBLMessageEndpointDelegate, MultipeerConnectionDelegate>
 }
 
 - (void) startDiscovery {
+    // TODO: check whether this is a new issue introduced?
+    // https://issues.couchbase.com/browse/CBL-3699
     _serverConnected = [self allowOverfillExpectationWithDescription: @"Server Connected"];
     _serverPeer = [[MCPeerID alloc] initWithDisplayName: @"server"];
     _serverSession = [[MCSession alloc] initWithPeer:_serverPeer
@@ -184,6 +186,8 @@ MCSessionDelegate, CBLMessageEndpointDelegate, MultipeerConnectionDelegate>
     _advertiser.delegate = self;
     [_advertiser startAdvertisingPeer];
     
+    // TODO: check whether this is a new issue introduced?
+    // https://issues.couchbase.com/browse/CBL-3699
     _clientConnected = [self allowOverfillExpectationWithDescription: @"Client Connected"];
     _clientPeer = [[MCPeerID alloc] initWithDisplayName: @"client"];
     _clientSession = [[MCSession alloc] initWithPeer: _clientPeer
--- a/Objective-C/Tests/MessageEndpointTest.m
+++ b/Objective-C/Tests/MessageEndpointTest.m
@@ -171,7 +171,7 @@ MCSessionDelegate, CBLMessageEndpointDelegate, MultipeerConnectionDelegate>
 }
 
 - (void) startDiscovery {
-    _serverConnected = [self expectationWithDescription: @"Server Connected"];
+    _serverConnected = [self allowOverfillExpectationWithDescription: @"Server Connected"];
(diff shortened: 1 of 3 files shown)


Objective-C/Tests/MessageEndpointTest.m
```diff
@@ -171,7 +171,7 @@ - (void)tearDown {
 }
 
 - (void) startDiscovery {
-    _serverConnected = [self expectationWithDescription: @"Server Connected"];
+    _serverConnected = [self allowOverfillExpectationWithDescription: @"Server Connected"];
```
pasin, line 174: Is it a bug that is causing multiple filling?

Swift/Tests/QueryTest.swift
```diff
@@ -1917,26 +1917,28 @@ class QueryTest: CBLTestCase {
         
         let x1 = expectation(description: "wait-for-live-query-changes")
         var count = 0;
+        var docCount = 0;
```
pasin, line 1920: Doesn't seem to get used.

Swift/Tests/QueryTest.swift
```diff
@@ -1917,26 +1917,28 @@ class QueryTest: CBLTestCase {
...
-            count = count + 1
             let rows =  Array(change.results!)
-            if count == 1 {
-                // initial notification about already existing result set!
-                XCTAssertEqual(rows.count, 9)
-                XCTAssertEqual(rows[0].int(at: 0), 1)
-            } else {
-                // deleted the doc should return empty result set!
-                XCTAssertEqual(rows.count, 8)
+            count = count + rows.count
```
pasin, line 1923: I do not understand why the previous code doesn't work. I expected that there will be only 2 callbacks (initial ones and the one after purge).

Objective-C/Tests/MessageEndpointTest.m
jayahariv: This might be, I will create a ticket and add a todo for now, will work in parallel. 
https://issues.couchbase.com/browse/CBL-3699

Swift/Tests/QueryTest.swift
jayahariv: But the callbacks can come in multiple sets right? one with all docIds or two with all docIds. There is no restriction in how the set will be delivered?

I have seen, sometimes, it returns in two set, but not always.

Objective-C/Tests/MessageEndpointTest.m
jayahariv: I believe, it started happening with the Xcode upgrade. I am trying to reproduce the issue locally(command line), it won't happen when running through Xcode.

Swift/Tests/QueryTest.swift
pasin: I am not sure I understand what you mean by the callback is called in multiple sets. The callback reports change in the query result, and the result will not be separated in multiple set.

jayahariv: I will go through the code once again.

jayahariv: You are right, my wrong understanding. Will revert and see, if the issue with unit test keep happening.

jayahariv: reverted the change!

pasin (approved):

PR #3033, opened 2022-10-21
Import the defaults file generated

jayahariv:
Fix the build errors
Fix the unit test to pass basic ones - more detailed unit tests will be added later


Objective-C/CBLDefaults.h
```diff
@@ -0,0 +1,73 @@
...
+//  limitations under the License.
+//
+
+// THIS IS AN AUTOGENERATED FILE, MANUAL CHANGES SHOULD BE EXPECTED TO
+// BE OVERWRITTEN
+
+#import "CBLReplicatorTypes.h"
+
+/** [NO] Plaintext is not used, and instead binary encoding is used in log files */
+extern const BOOL kCBLDefaultLogFileUsePlainText;
```
pasin, line 26: I think it's better to group them into sections using configuration class names so that it's easier to know which ones are for which configuration such as 

#pragma mark - CBLReplicatorConfiguration

Objective-C/CBLDefaults.m
```diff
@@ -0,0 +1,41 @@
...
+const BOOL kCBLDefaultReplicatorAllowReplicatingInBackground = NO;
+const NSTimeInterval kCBLDefaultReplicatorHeartbeat = 300;
+const NSUInteger kCBLDefaultReplicatorMaxAttemptsSingleShot = 9;
+const NSUInteger kCBLDefaultReplicatorMaxAttemptsContinuous = NSUIntegerMax;
+const NSTimeInterval kCBLDefaultReplicatorMaxAttemptWaitTime = 300;
+const BOOL kCBLDefaultReplicatorEnableAutoPurge = YES;
+const unsigned short kCBLDefaultListenerPort = 0;
+const BOOL kCBLDefaultListenerDisableTls = NO;
+const BOOL kCBLDefaultListenerReadOnly = NO;
+const BOOL kCBLDefaultListenerEnableDeltaSync = NO;
```
pasin, line 40: Same here. For easier to read, group them into section.

Objective-C/CBLFullTextIndexConfiguration.h
```diff
@@ -25,7 +25,7 @@ NS_ASSUME_NONNULL_BEGIN
 @interface CBLFullTextIndexConfiguration : CBLIndexConfiguration
 
 /**
- Set the true value to ignore accents/diacritical marks. The default value is false.
+ Set the true value to ignore accents/diacritical marks.
```
pasin, line 28: Better to link to constant variable.

https://developer.apple.com/documentation/xcode/formatting-your-documentation-content#Link-to-symbols-and-other-content

Objective-C/CBLReplicatorConfiguration.h
```diff
@@ -136,30 +136,36 @@ __deprecated_msg(" Use [... initWithTarget:] and [config addCollection: config:]
 /**
  The heartbeat interval in second.
 
- The interval when the replicator sends the ping message to check whether the other peer is still alive. Set the value to zero(by default)
- means using the default heartbeat of 300 seconds.
+ The interval when the replicator sends the ping message to check whether the other peer is still alive.
+ Default heartbeat is 300 seconds.
```
pasin, line 140: Link to the constant variable instead of mentioning the actual value here so when the value gets changed, there is only once place to update.

Example:

```
Default heartbeat is ``kCBLDefaultReplicatorHeartbeat``
```

Please check if the generated API doc links to the constant variable correctly.

pasin: Same comments for the others.

Swift/Defaults.swift
```diff
@@ -0,0 +1,75 @@
...
+//
+
+// THIS IS AN AUTOGENERATED FILE, MANUAL CHANGES SHOULD BE EXPECTED TO
+// BE OVERWRITTEN
+
+import Foundation
+
+public struct Defaults {
+	/// [false] Plaintext is not used, and instead binary encoding is used in log files
+	public static let defaultLogFileUsePlainText: Bool = kCBLDefaultLogFileUsePlainText.boolValue
```
pasin, line 27: Same as objective-c. Group them into section. Alternatively I think extension can be used here as well.

pasin (changes requested):

Objective-C/CBLFullTextIndexConfiguration.h
jayahariv: For objc, in this case, this value is not nullable. So user has to provide either NO or YES. So there is no default value in this case.

Swift/Defaults.swift
jayahariv: I think for consistency with Defaults files, it's better to keep in same Defaults struct itself, and group using `// MARK: group name`

```
public struct Defaults {

// MARK: LogFileConfiguration 

// MARK: FullTextIndexConfiguration

// MARK: ReplicatorConfiguration 

}
```

We could separate out into extension classes, sometime in future when it goes out of control?

pasin: IMO, it doesn't need to be in sync with Objective-C. Which option do you prefer as a user? You can make decision.

jayahariv: >  Which option do you prefer as a user?

My understanding, this constant will be used by very few users. Since it's the default value, so no need to set it. It will be useful for CBL devs, to set it. Am I right?

Personally, for me I am fine with both ways, 
* more cleaner/organized code seems like extensions, 
* we only have very limited number of constants and it is fit in one page itself(not feels like, worth spending formatting it may be?)

pasin: I don't think it's about the number of constants. It's about API and API doc.

pasin: > My understanding, this constant will be used by very few users. Since it's the default value, so no need to set it. It will be useful for CBL devs, to set it. Am I right?

They are served references for the default values. It's a user's choice to use it for setting the config values programmatically.

pasin: If you think using Defaults is preferred because it's easier to use and organized enough, go for it. 

But if you think that using extension is better, but we only have 17 configs now so let don't spend more time on it. I feel that it's not a right argument. Changing the API later also means that the old API needs to deprecated.

jayahariv: ok. I think, I will agree and go for extension.

I was thinking from a stand point where we are keeping the constants in a single place and exposed outside just for reference. Not given enough thought from an API / API doc stand point. I get your point now.

jayahariv: Updated the new format as comment: https://issues.couchbase.com/browse/CBL-3744

Swift/LogFileConfiguration.swift
```diff
@@ -26,22 +26,22 @@ public class LogFileConfiguration {
...
     
     /// To use plain text file format instead of the default binary format.
-    public var usePlainText: Bool = false  {
+    public var usePlainText: Bool = Defaults.defaultLogFileUsePlainText  {
         willSet(newValue) {
             checkReadOnly()
         }
     }
     
     /// The maximum size of a log file before being rotated in bytes. The default is 500 Kilobytes.
```
jayahariv, line 35: default value -> do link to symbol in documentation

pasin (approved):

PR #3040, opened 2022-11-07
[fix] simulator issue + unit test failures are disabled 

jayahariv:
* get the latest from simctl and use it instead of hard coding it
* disable the unit tests which are consistently failing, with Jira ticket.


pasin (approved):
All these query test fails are weird.

jayahariv:
I suspect there is something wrong in the latest commit from LC

PR #3068, opened 2022-12-07
Fix DNS Service, Socket Connect and Cleanup

pasin:
[AddressInfo]

* Updated port to sockaddr directly by using const_cast - this is not the best practice but it is the simplest way to do.
* Added addrIn and addrIn6 for easier accessing to the ip address structure.
* Added host and interface for debugging and logging purpose.

[CBLDNSService]

* Renamed the class to CBLDNSService for consistency
* Added the logic to check whether the host is already IP Address and notify the result right away.
* Ensured to notify result or error on the dispatch queue.
* Remove kDNSServiceFlagsTimeout flas as we are doing our own timeout to better match the default litecore connection timeout of 15 seconds.
* Used dispatch after instead of NSTimer as NSTimer does work in dispatch queue correctly.
* Updated log messages.
* Updated license

[CBLWebSocket]
* Fiexed the code that creates and connects the socket; Correct len passed to connect() function.
* Updated log messages
* Cleaned up imports

[CBLReplicatorConfiguration]
* Reordered the experimental options per priority


jayahariv (approved):
looks good and tested and working

PR #3070, opened 2022-12-08
fix: build error on Xcode 13.x

jayahariv:
error: implicit conversion from integral type 'unsigned int' to 'BOOL' [-Werror,-Wobjc-signed-char-bool-implicit-int-conversion] probably the lower Xcode error, its not happening locally


jayahariv:
Validating the fix on Jenkins machine!

pasin (approved):

PR #3075, opened 2022-12-16
Update DNS resolution to use DNSService

jayahariv:
* getaddrinfo is using default device network interface, so updated to use the DNSService

### DIFF
--- a/CouchbaseLite.xcodeproj/project.pbxproj
+++ b/CouchbaseLite.xcodeproj/project.pbxproj
@@ -122,6 +122,12 @@
 		1AAB2787227793E50037A880 /* CBLConflict.m in Sources */ = {isa = PBXBuildFile; fileRef = 1AAB273C22739EBB0037A880 /* CBLConflict.m */; };
 		1AAF6372226A8B060016754C /* QueryTest+Meta.m in Sources */ = {isa = PBXBuildFile; fileRef = 1AAF6371226A8B060016754C /* QueryTest+Meta.m */; };
 		1AAF6383226A8DBF0016754C /* QueryTest+Join.m in Sources */ = {isa = PBXBuildFile; fileRef = 1AAF6382226A8DBF0016754C /* QueryTest+Join.m */; };
+		1AC3E443294C977800156AE0 /* CBLDNSService.h in Headers */ = {isa = PBXBuildFile; fileRef = 1AC3E436294C977700156AE0 /* CBLDNSService.h */; };
+		1AC3E444294C977800156AE0 /* CBLDNSService.h in Headers */ = {isa = PBXBuildFile; fileRef = 1AC3E436294C977700156AE0 /* CBLDNSService.h */; };
+		1AC3E445294C977800156AE0 /* CBLDNSService.mm in Sources */ = {isa = PBXBuildFile; fileRef = 1AC3E442294C977700156AE0 /* CBLDNSService.mm */; };
+		1AC3E446294C977800156AE0 /* CBLDNSService.mm in Sources */ = {isa = PBXBuildFile; fileRef = 1AC3E442294C977700156AE0 /* CBLDNSService.mm */; };
+		1AC3E447294C979200156AE0 /* CBLDNSService.mm in Sources */ = {isa = PBXBuildFile; fileRef = 1AC3E442294C977700156AE0 /* CBLDNSService.mm */; };
+		1AC3E448294C979400156AE0 /* CBLDNSService.mm in Sources */ = {isa = PBXBuildFile; fileRef = 1AC3E442294C977700156AE0 /* CBLDNSService.mm */; };
 		1AC7EC29249DA24E00978C2E /* Foundation+CBL.h in Headers */ = {isa = PBXBuildFile; fileRef = 1AC7EC27249DA24E00978C2E /* Foundation+CBL.h */; };
 		1AC7EC2A249DA24E00978C2E /* Foundation+CBL.h in Headers */ = {isa = PBXBuildFile; fileRef = 1AC7EC27249DA24E00978C2E /* Foundation+CBL.h */; };
 		1AC7EC2B249DA24E00978C2E /* Foundation+CBL.mm in Sources */ = {isa = PBXBuildFile; fileRef = 1AC7EC28249DA24E00978C2E /* Foundation+CBL.mm */; };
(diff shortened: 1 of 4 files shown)


pasin (changes requested):

CouchbaseLite.xcodeproj/project.pbxproj
```diff
@@ -122,6 +122,12 @@
...
 		1AAF6372226A8B060016754C /* QueryTest+Meta.m in Sources */ = {isa = PBXBuildFile; fileRef = 1AAF6371226A8B060016754C /* QueryTest+Meta.m */; };
 		1AAF6383226A8DBF0016754C /* QueryTest+Join.m in Sources */ = {isa = PBXBuildFile; fileRef = 1AAF6382226A8DBF0016754C /* QueryTest+Join.m */; };
+		1AC3E443294C977800156AE0 /* CBLDNSService.h in Headers */ = {isa = PBXBuildFile; fileRef = 1AC3E436294C977700156AE0 /* CBLDNSService.h */; };
+		1AC3E444294C977800156AE0 /* CBLDNSService.h in Headers */ = {isa = PBXBuildFile; fileRef = 1AC3E436294C977700156AE0 /* CBLDNSService.h */; };
+		1AC3E445294C977800156AE0 /* CBLDNSService.mm in Sources */ = {isa = PBXBuildFile; fileRef = 1AC3E442294C977700156AE0 /* CBLDNSService.mm */; };
+		1AC3E446294C977800156AE0 /* CBLDNSService.mm in Sources */ = {isa = PBXBuildFile; fileRef = 1AC3E442294C977700156AE0 /* CBLDNSService.mm */; };
+		1AC3E447294C979200156AE0 /* CBLDNSService.mm in Sources */ = {isa = PBXBuildFile; fileRef = 1AC3E442294C977700156AE0 /* CBLDNSService.mm */; };
+		1AC3E448294C979400156AE0 /* CBLDNSService.mm in Sources */ = {isa = PBXBuildFile; fileRef = 1AC3E442294C977700156AE0 /* CBLDNSService.mm */; };
```
pasin, line 130: Just noticed the double record here for CBLDNSService.mm.

Objective-C/Internal/Replicator/CBLWebSocket.mm
```diff
@@ -372,63 +370,79 @@ - (void) connectToHostWithName: (NSString*)hostname
...
+        NSString* msg = $sprintf(@"Failed to find network interface %@ with errno %d", interface, errNo);
         CBLWarnError(WebSocket, @"%@: %@", self, msg);
-        [self closeWithError: posixError(EPERM, msg)];
+        [self closeWithError: posixError(errNo, msg)];
         return;
     }
-    
-    // Create socket:
-    [self _socketConnect: interface];
+
```
pasin, line 382: Move from _socketConnect: to here : 
CBLLogVerbose(WebSocket, @"%@: Interface '%@' is mapped to index '%u'", self, _networkInterface, index);

Objective-C/Internal/Replicator/CBLWebSocket.mm
```diff
@@ -372,63 +370,79 @@ - (void) connectToHostWithName: (NSString*)hostname
...
         CBLWarnError(WebSocket, @"%@: %@", self, msg);
         [self closeWithError: posixError(errNo, msg)];
         return;
     }
     
     // Set network interface:
-    if (interface) {
-        NSError* err;
-        bool useIPv4 = (_addr->ai_family == AF_INET);
+    if (_networkInterface) {
```
pasin, line 426: There is no case that this will be nil. You could also just do the assert that `info.interface > 0`.

Objective-C/Internal/Replicator/CBLWebSocket.mm
```diff
@@ -372,63 +370,79 @@ - (void) connectToHostWithName: (NSString*)hostname
...
+        UInt32 index = info.interface;
+        CBLLogVerbose(WebSocket, @"%@: Interface '%@' is mapped to index '%u'", self, _networkInterface, index);
```
pasin, line 428: Move to where it gets mapped above.

Objective-C/Internal/Replicator/CBLWebSocket.mm
```diff
@@ -372,63 +370,79 @@ - (void) connectToHostWithName: (NSString*)hostname
...
+        } else if (info.addr->sa_family == AF_INET6 ){
+            result = setsockopt(_sockfd, IPPROTO_IPV6, IPV6_BOUND_IF, &index, sizeof(index));
+        }
         
-        if (![self setSocketOptForInterface: interface useIPv4: useIPv4 error: &err]) {
-            [self closeWithError: err];
+        if (result < 0) {
+            int errNo = errno;
+            NSString* msg = $sprintf(@"Failed to set network interface %@ with errno %d (%@)",
+                                     _networkInterface, errNo, info);
```
pasin, line 439: As there is already a log about mapping network interface to index above, here we can log with just index as this method only uses index.

Objective-C/Internal/Replicator/CBLWebSocket.mm
```diff
@@ -372,63 +370,79 @@ - (void) connectToHostWithName: (NSString*)hostname
...
+            CBLWarnError(WebSocket, @"%@: %@", self, msg);
+            [self closeWithError: posixError(errNo, msg)];
             return;
         }
+        
+        CBLLogVerbose(WebSocket, @"%@: Successfully set network interface %@ to socket option", self, _networkInterface);
```
pasin, line 445: Same here log with `index` as this method is only using index.

Objective-C/Internal/Replicator/CBLWebSocket.mm
```diff
@@ -472,8 +486,8 @@ - (void) _socketConnect: (NSString*)interface {
             });
         } else {
             int errNo = errno;
-            NSString* msg = interface ?
-            $sprintf(@"Failed to connect via the specified network interface %@ with errno %d", interface, errNo) :
+            NSString* msg = _networkInterface ?
+            $sprintf(@"Failed to connect via the specified network interface %@ with errno %d", _networkInterface, errNo) :
```
pasin, line 490: There is no case that _networkInterace == nil here after all. Let log with the index instead of _networkInterace as this method is using the index.

pasin (changes requested):

CouchbaseLite.xcodeproj/project.pbxproj
jayahariv: Those four references might be for the '.mm' file included in 4 targets(CBL_Swift, CBL_ObjC, CBL_EE_ObjC, CBL_EE_Swift)? Same is for below `Foundation+CBL.mm`?

pasin: I do not know how this actually works as there are exactly the same.

pasin (approved):

CouchbaseLite.xcodeproj/project.pbxproj
jayahariv: I tried remove and added back, it's same 4 file references

jayahariv: 1AC3E445294C977800156AE0 - CBL_Swift
1AC3E446294C977800156AE0 - CBL_EE_Swift
1AC3E447294C979200156AE0 - CBL_ObjC
1AC3E448294C979400156AE0 - CBL_EE_ObjC

First ID differs and points to each target

PR #3082, opened 2022-12-21
CBL-3953: documentWithID shouldn't throw, if document doesn't exists

jayahariv:
* NonExisting document shouldn't throw NotFound error inside documentWithID, will skip the error populating. 
* Default Swift API from ObjC overriden(NS_SWIFT_NOTHROW) to avoid throw nilError; 
* ObjC API ending with ':(NSError**)' treated as throwable compulsory return type, but we need nullable(optional) return type.

### DIFF
--- a/Objective-C/CBLCollection.h
+++ b/Objective-C/CBLCollection.h
@@ -87,7 +87,8 @@ extern NSString* const kCBLDefaultCollectionName;
  @param error On return, the error if any.
  @return The CBLDocument object.
  */
-- (nullable CBLDocument*) documentWithID: (NSString*)documentID error: (NSError**)error;
+- (nullable CBLDocument*) documentWithID: (NSString*)documentID
+                                   error: (NSError**)error NS_SWIFT_NOTHROW;
 
 #pragma mark - Save, Delete, Purge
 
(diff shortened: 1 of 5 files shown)


Objective-C/CBLCollection.h
```diff
@@ -87,7 +87,8 @@ extern NSString* const kCBLDefaultCollectionName;
  @param error On return, the error if any.
  @return The CBLDocument object.
  */
-- (nullable CBLDocument*) documentWithID: (NSString*)documentID error: (NSError**)error;
+- (nullable CBLDocument*) documentWithID: (NSString*)documentID
+                                   error: (NSError**)error NS_SWIFT_NOTHROW;
```
pasin, line 91: I do not know that there is NS_SWIFT_NOTHROW.

pasin (approved):

Objective-C/CBLCollection.h
jayahariv: We are already using the same in [isDocumentPending API](https://github.com/couchbase/couchbase-lite-ios/blob/master/Objective-C/CBLReplicator.h#L181)

PR #3083, opened 2022-12-22
Fixing a crash “Attempt to perform an operation on a closed database”

RockyCognizant:
**Bug**: “Attempt to perform an operation on a closed database”

**Reason**: self.mustBeOpen may raise an exception but the code cannot capture it without this PR, so usually when the error happen, the app will crash

**Fix**: Call `self.mustBeOpen` with an error parameter.

### DIFF
--- a/Objective-C/CBLDatabase.mm
+++ b/Objective-C/CBLDatabase.mm
@@ -345,8 +345,15 @@ static const C4DatabaseConfig2 kDBConfig = {
     CBLAssertNotNil(block);
     
     CBL_LOCK(_mutex) {
-        [self mustBeOpen];
-        
+        NSError* exception = nil;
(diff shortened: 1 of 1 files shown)


RockyCognizant:
Also there is a possibility that could trigger the same error:
[CBLReplicator.mm#L176](https://github.com/couchbase/couchbase-lite-ios/blob/master/Objective-C/CBLReplicator.mm#L176)

```obj-c
 [_config.database addActiveStoppable: self];
 ```
Suggest to add a try-catch but please review this PR first.

jayahariv:
thanks @RockyCognizant, unfortunately we cannot merge these changes right now(to master). We are slowly moving away from throwing exception in ObjC/Swift, but this will be done in a later release. We will be tracking it here CBL-4092

PR #3090, opened 2023-01-10
CBL-4052: objc API ends with error updated with NoThrow

jayahariv:
* No throw for objc APIs ends with error
* Make it consistent for all APIs instead of do..try..catch in save, delete functions

### DIFF
--- a/Objective-C/CBLDatabase.h
+++ b/Objective-C/CBLDatabase.h
@@ -111,7 +111,7 @@ __deprecated_msg("Use [database defaultCollection].count instead.");
  @return The CBLDocument object.
  */
 - (nullable CBLDocument*) documentWithID: (NSString*)id
-__deprecated_msg("Use [database defaultCollection] documentWithID:] instead.") NS_SWIFT_NOTHROW;
+__deprecated_msg("Use [database defaultCollection] documentWithID:] instead.");
 
 
 #pragma mark - Subscript
--- a/Objective-C/CBLDatabase.h
+++ b/Objective-C/CBLDatabase.h
@@ -530,7 +530,7 @@ __deprecated_msg("Use [[database defaultCollection] getDocumentExpirationWithID:
  @return list of collections in the scope, or nil if an error occurred.
  */
 - (nullable NSArray<CBLCollection*>*) collections: (nullable NSString*)scope
-                                            error: (NSError**)error NS_SWIFT_NOTHROW;
+                                            error: (NSError**)error;
 
 /**
  Create a named collection in the specified scope.
@@ -543,7 +543,7 @@ __deprecated_msg("Use [[database defaultCollection] getDocumentExpirationWithID:
  */
 - (nullable CBLCollection*) createCollectionWithName: (NSString*)name
                                                scope: (nullable NSString*)scope
-                                               error: (NSError**)error NS_SWIFT_NOTHROW;
+                                               error: (NSError**)error;
 
 /**
  Get a collection in the specified scope by name.
@@ -581,7 +581,7 @@ __deprecated_msg("Use [[database defaultCollection] getDocumentExpirationWithID:
  @param error On return, the error if any.
  @return Default Scope, or nil if an error occurred.
  */
-- (nullable CBLScope*) defaultScope: (NSError**)error NS_SWIFT_NOTHROW;
+- (nullable CBLScope*) defaultScope: (NSError**)error;
 
 /**
  Get the default collection. If the default collection is deleted, nil will be returned.
@@ -590,7 +590,7 @@ __deprecated_msg("Use [[database defaultCollection] getDocumentExpirationWithID:
         default collection is deleted.
  @return Default collection, or nil if an error occurred.
  */
-- (nullable CBLCollection*) defaultCollection: (NSError**)error NS_SWIFT_NOTHROW;
+- (nullable CBLCollection*) defaultCollection: (NSError**)error;
 
 @end
 
--- a/Objective-C/CBLDatabase.h
+++ b/Objective-C/CBLDatabase.h
@@ -111,7 +111,7 @@ __deprecated_msg("Use [database defaultCollection].count instead.");
  @return The CBLDocument object.
  */
 - (nullable CBLDocument*) documentWithID: (NSString*)id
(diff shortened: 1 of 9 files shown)


Objective-C/CBLDatabase.h
```diff
@@ -111,7 +111,7 @@ __deprecated_msg("Use [database defaultCollection].count instead.");
  @return The CBLDocument object.
  */
 - (nullable CBLDocument*) documentWithID: (NSString*)id
-__deprecated_msg("Use [database defaultCollection] documentWithID:] instead.");
+__deprecated_msg("Use [database defaultCollection] documentWithID:] instead.") NS_SWIFT_NOTHROW;
```
pasin, line 114: Does it need here as well when no out error?

Objective-C/CBLDatabase.h
```diff
@@ -543,7 +543,7 @@ __deprecated_msg("Use [[database defaultCollection] getDocumentExpirationWithID:
  */
 - (nullable CBLCollection*) createCollectionWithName: (NSString*)name
                                                scope: (nullable NSString*)scope
-                                               error: (NSError**)error;
+                                               error: (NSError**)error NS_SWIFT_NOTHROW;
```
pasin, line 546: Is there a case that a null collection returned without an error?

Objective-C/CBLDatabase.h
```diff
@@ -581,7 +581,7 @@ __deprecated_msg("Use [[database defaultCollection] getDocumentExpirationWithID:
  @param error On return, the error if any.
  @return Default Scope, or nil if an error occurred.
  */
-- (nullable CBLScope*) defaultScope: (NSError**)error;
+- (nullable CBLScope*) defaultScope: (NSError**)error NS_SWIFT_NOTHROW;
```
pasin, line 584: I don't think there is a case that null is returned without an error.

Objective-C/CBLDatabase.h
```diff
@@ -590,7 +590,7 @@ __deprecated_msg("Use [[database defaultCollection] getDocumentExpirationWithID:
         default collection is deleted.
  @return Default collection, or nil if an error occurred.
  */
-- (nullable CBLCollection*) defaultCollection: (NSError**)error;
+- (nullable CBLCollection*) defaultCollection: (NSError**)error NS_SWIFT_NOTHROW;
```
pasin, line 593: There is no case as of now to return a null default collection without an error.

Objective-C/CBLQueryFactory.h
```diff
@@ -33,7 +33,7 @@ NS_ASSUME_NONNULL_BEGIN
  @param error On return, the error if any., the given query string is invalid(e.g., syntax error).
  @return query created using the given expression string, or nil if an error occurred.
  */
-- (nullable CBLQuery*) createQuery: (NSString*)query error: (NSError**)error;
+- (nullable CBLQuery*) createQuery: (NSString*)query error: (NSError**)error NS_SWIFT_NOTHROW;
```
pasin, line 36: I don't think there is a case that a null is returned without error.

Objective-C/CBLReplicator.h
```diff
@@ -169,7 +169,7 @@ __deprecated_msg("Use [replicator pendingDocumentIDsForCollection:error:] instea
  @return A  set of document Ids, each of which has one or more pending revisions. If error, nil.
  */
 - (nullable NSSet<NSString*>*) pendingDocumentIDsForCollection: (CBLCollection*)collection
-                                                         error: (NSError**)error;
+                                                         error: (NSError**)error NS_SWIFT_NOTHROW;
```
pasin, line 172: Is there a case that null is returned without an error?

Objective-C/CBLScope.h
```diff
@@ -53,7 +53,7 @@ extern NSString* const kCBLDefaultScopeName;
  @param error On return, the error if any.
  @return Collections in the scope, or nil if an error occurred.
  */
-- (nullable NSArray<CBLCollection*>*) collections: (NSError**)error;
+- (nullable NSArray<CBLCollection*>*) collections: (NSError**)error NS_SWIFT_NOTHROW;
```
pasin, line 56: I think there is no case to return null collections without error.

pasin (commented):
I think this PR is risky right be helium and I don't see it necessary as well.

jayahariv:
This PR is necessary! 

Example: 
```- (nullable Collection) collection: (NSError**)error ```

This ObjC Api will generate swift API with a non-optional return type. 

```func collection() throws -> Collection ```

So If the ObjC API returns nil in some case, it will cause swift to receive an exception/error.

File: Objective-C/CBLDatabase.h (hunk 1 above)
Comment (jayahariv) [reply Objective-C/CBLDatabase.h:114]: its not necessary, I might have accidentally updated with collection documentWithID:error:

File: Objective-C/CBLQueryFactory.h (hunk 5 above)
Comment (pasin) [inline Objective-C/CBLQueryFactory.h:36]: No cases that null will be returned without an error?

File: Objective-C/CBLReplicator.h (hunk 6 above)
Comment (pasin) [inline Objective-C/CBLReplicator.h:172]: No cases that null will be returned without an error?

File: Objective-C/CBLScope.h (hunk 7 above)
Comment (pasin) [inline Objective-C/CBLScope.h:56]: No cases that null will be returned without an error?

File: Objective-C/CBLDatabase.h [hunk 8]
```diff
@@ -530,7 +530,7 @@ __deprecated_msg("Use [[database defaultCollection] getDocumentExpirationWithID:
  @return list of collections in the scope, or nil if an error occurred.
  */
 - (nullable NSArray<CBLCollection*>*) collections: (nullable NSString*)scope
-                                            error: (NSError**)error;
+                                            error: (NSError**)error NS_SWIFT_NOTHROW;
```
pasin, line 533: No cases that null will be returned without an error?

pasin (changes requested):
Basically, please change the PR to add NS_SWIFT_NOTHROW only on the methods that need to return null with null error.

Objective-C/CBLDatabase.h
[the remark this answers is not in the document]
jayahariv: I have made consistent with other call, that's why.

pasin (approved):

PR #3105, opened 2023-03-16
CBL-4308 : Add subscript function to Collection class

pasin:
* We have a subscript function in the Database class for getting a document from the database using the subscript syntax such as database[“doc1”]. This is Apple platform only API.  The subscript function in the database class should be deprecated and there should be the same functionality implemented in the Collection class for an alternative API.

* Deprecated subscript function the in Database class for both Swift and Objective-C.

* Implemented subscript function in Collection class.

### DIFF
--- a/Objective-C/CBLCollection.h
+++ b/Objective-C/CBLCollection.h
@@ -23,6 +23,7 @@
 
 @class CBLDocument;
 @class CBLDocumentChange;
+@class CBLDocumentFragment;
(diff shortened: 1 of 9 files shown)


velicuvlad (approved):
LGTM

bmeike (commented):
Is this a user-visible feature?  Is that why we are adding it, now?

pasin:
Yes, it's a public API.

PR #3118, opened 2023-04-13
CBL-4401: Update tests to avoid use of deprecated api 1

velicuvlad:
- add defaultCollection within CBLTestCase so it can be accessed in every test
- update calls in:
    - DatabaseEncryptionTest
    - DatabaseTest 

defaultCollection should be optional such that it gets released and set to nil prior to database reset/deletion. Otherwise, some tests which involve the operations above would fail.

### DIFF
--- a/Swift/Tests/CBLTestCase.swift
+++ b/Swift/Tests/CBLTestCase.swift
@@ -40,7 +40,9 @@ class CBLTestCase: XCTestCase {
     
     let otherDatabaseName = "otherdb"
     
-    var defaultCol: Collection?
+    var defaultCollection: Collection?
+    
+    var otherDB_defaultCollection: Collection?
     
     #if COUCHBASE_ENTERPRISE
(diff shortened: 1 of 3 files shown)


Swift/Tests/CBLTestCase.swift
```diff
@@ -40,6 +40,8 @@ class CBLTestCase: XCTestCase {
     
     let otherDatabaseName = "otherdb"
     
+    var defaultCol: Collection?
```
pasin, line 43: Sorry for a bit picky here. Can we use full name, `defaultCollection`? In CBL-C test, it's also `defaultCollection`.

pasin (approved):

Swift/Tests/CBLTestCase.swift
velicuvlad: will do

PR #3119, opened 2023-04-14
CBL-4428 : Fix database exclusive calls to c4queryobs* functions

pasin:
* As c4queryobs_* (except c4queryobs_create) functions need to be callded under database-exclusive lock, made sure those function calls are under database’s lock.

* Combined stop() and stopAndFree() together to have only stop().

* Ensured not call to get the query result in postQueryChange() when the observer is already stopped.


borrrden (approved):

pasin:
Closed as will do this in Beryllium instead.

PR #3121, opened 2023-04-17
CBL-4403: Update tests to avoid use of deprecated api 3

velicuvlad:
- DocumentExpirationTest
- FragmentTest
- Logtest
- MessageEndpointTest

For the moment, listener tests contain workaround in place to deal with [CBL-4438](https://issues.couchbase.com/browse/CBL-4438) - so is safe to merge. 

Will revise those tests once when the issue is fixed.

### DIFF
--- a/Swift/Tests/MessageEndpointTest.swift
+++ b/Swift/Tests/MessageEndpointTest.swift
@@ -21,7 +21,7 @@ import XCTest
 import MultipeerConnectivity
 import CouchbaseLiteSwift
 
-protocol MultipeerConnectionDelegate: class {
+protocol MultipeerConnectionDelegate: AnyObject {
     
     func connectionDidOpen(_ connection: MessageEndpointConnection)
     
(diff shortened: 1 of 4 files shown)


Swift/Tests/MessageEndpointTest.swift
```diff
@@ -21,7 +21,7 @@ import XCTest
 import MultipeerConnectivity
 import CouchbaseLiteSwift
 
-protocol MultipeerConnectionDelegate: class {
+protocol MultipeerConnectionDelegate: AnyObject {
```
pasin, line 24: What is the difference b/w AnyObject and class here? What is the issue with class?

pasin (approved):

pasin (approved):
Can you also mark `CBL-4438` with a comment where you put the workaround?

Swift/Tests/MessageEndpointTest.swift
velicuvlad: xcode deprecation warning - for class-constrained protocols, AnyObject should be used instead of class.

PR #3130, opened 2023-05-11
CBL-4440 : Fix collection leaks when not removing document change listener token

pasin:
* Made collection weak reference in CBLDocumentChangeNotifier so that there is no cirular reference b/w colletion and its CBLDocumentChangeNotifier objects.

* Added warning when skipping the notification as the collection object has been released and improved warning logs in CBLDocumentChangeNotifier.

* Added tests to ensure that there is no memory leak when not removing listener token in collection (Note : we have checked for C4 Object Leaks in Tests’ teardown).

### DIFF
--- a/Objective-C/Tests/CollectionTest.m
+++ b/Objective-C/Tests/CollectionTest.m
@@ -648,6 +648,23 @@
     AssertEqual(changeListenerFired, 0);
 }
 
+- (void) testCollectionChangeListenerWithoutRemoveToken {
+    @autoreleasepool {
+        NSError* error = nil;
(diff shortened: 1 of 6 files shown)


Objective-C/Tests/CollectionTest.m
```diff
@@ -648,6 +648,23 @@ - (void) testCollectionChangeListener {
     AssertEqual(changeListenerFired, 0);
 }
 
+- (void) testCollectionChangeListenerWithoutRemoveToken {
```
borrrden, line 651: Seems like it could use comments since the actual assertions that are important are handled in the teardown (same for all new tests in this PR)?

borrrden (approved):

PR #3132, opened 2023-05-18
CBL-4405: Update tests to avoid use of deprecated api 5

velicuvlad:
- replicator tests
- url endpoint listener tests
- small adjustments to base class and aux - the majority is name changing

### DIFF
--- a/Swift/Tests/CBLTestCase.swift
+++ b/Swift/Tests/CBLTestCase.swift
@@ -92,8 +92,9 @@ class CBLTestCase: XCTestCase {
     
     override func tearDown() {
         self.defaultCollection = nil
+        self.otherDB_defaultCollection = nil
         try! db.close()
-        try! otherDB?.close()
+        try? otherDB?.close()
         super.tearDown()
     }
     
--- a/Swift/Tests/CBLTestCase.swift
+++ b/Swift/Tests/CBLTestCase.swift
@@ -94,7 +94,7 @@ class CBLTestCase: XCTestCase {
         self.defaultCollection = nil
         self.otherDB_defaultCollection = nil
         try! db.close()
-        try? otherDB?.close()
+        try! otherDB?.close()
         super.tearDown()
     }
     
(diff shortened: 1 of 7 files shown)


Swift/Tests/CBLTestCase.swift
```diff
@@ -92,8 +92,9 @@ class CBLTestCase: XCTestCase {
     
     override func tearDown() {
         self.defaultCollection = nil
+        self.otherDB_defaultCollection = nil
         try! db.close()
-        try! otherDB?.close()
+        try? otherDB?.close()
```
pasin, line 97: Not sure, can it be `try!`?

Swift/Tests/URLEndpointListenerTest.swift
```diff
@@ -21,8 +21,8 @@ import XCTest
 
 @available(macOS 10.12, iOS 10.3, *)
 class URLEndpointListenerTest: ReplicatorTest {
-    let wsPort: UInt16 = 4984
-    let wssPort: UInt16 = 4985
+    let wsPort: UInt16 = 2
+    let wssPort: UInt16 = 3
```
pasin, line 25: Port 2 and 3 are reserved. Is there a reason to change the port number here? If it is because that they are the same as default SG port number, you can change to something like 4084 and 4085.

pasin (changes requested):
A couple comments.

Swift/Tests/URLEndpointListenerTest.swift
velicuvlad: yes, that was exactly the reason I wanted to change. Will change to 4084 and 4085. 

I've also set the kConnRefusedTarget to port 1. Will change that to 4083 as well.

Swift/Tests/CBLTestCase.swift
velicuvlad: I just tested, `try!` works as well... I don't remember why I put it that way.. will update

pasin (approved):

PR #3147, opened 2023-05-26
CBL-4438 : Prevent crash when notifying doc change for already-released collection

pasin:
* Ported the fix b48ba0e524b884d6455b3f4fab171fdaa3373784 from master branch.

* Changed to weak self and added warning log when collection ref has been released.


bmeike (approved):
The new code looks right.  Not sure what "unowned" does... ;-P

pasin:
It's equivalent to a naked pointer with being retained.

PR #3152, opened 2023-06-03
CBL-4580 : Fix Document’s contains(key:) returning wrong result

pasin:
Fixed CBLDocument’s containsValueForKey: to call the right function when checking if the key exists in the document.


velicuvlad (approved):
LGTM

bmeike (approved):

PR #3154, opened 2023-06-06
CBL-3873: PR validation should run on iOS_App_Tests target

velicuvlad:


pasin (commented):
Do you know the reason behind this change?

velicuvlad:
It seems to be mostly the same tests, so from that standpoint it wouldn't make much difference, but iOS is more restrictive than macOS so it seemed like a good idea to try EE Jenkins straight on iOS.

pasin (approved):

PR #3161, opened 2023-06-20
CBL-2780: Add action to run SG tests

velicuvlad:
- different scheme for SG tests
- use Walrus and pick-up cert from LiteCore for SSL (.der)
- GH Actions validation
- changed ports for URLEndpointListener tests to not clash with sgw so full test suite can be at once.


velicuvlad:
Jenkins PR validation won't pass because of submodule ref. I believe this can be merged as it is .. we will update LiteCore with latest once there's a green build

pasin (commented):
Look good. One question, can we reuse the existing scheme (CBL_ObjC) that we have instead of creating a new one as it seems to be the same?

pasin (approved):

PR #3173, opened 2023-06-30
CBL-2862: Add package validation actions

velicuvlad:


Sources/CouchbaseLiteSwift/CouchbaseLiteSwift.swift
```diff
@@ -0,0 +1,3 @@
+struct CouchbaseLiteSwift {
```
pasin, line 1: Couchbase Lite Swift package is a binary swift package. Can you explain why we need this file? If we really need it, can we make it an empty struct without the "Hello, World!"?

.github/workflows/check_package.yml
```diff
@@ -0,0 +1,38 @@
...
+
+    steps:
+      - uses: actions/checkout@v2
+      - uses: swift-actions/setup-swift@v1.23.0
+        with:
+            swift-version: ${{ matrix.swift }} 
+      - name: Build & test MacOS
+        run: |
+          swift build -v
+          swift test -v
```
pasin, line 27: I do not know how this works when trying to build the binary swift package which is generated from swift version 5.3.

pasin: Umm ... we are doing something similar here but it works.

https://github.com/couchbase/couchbase-lite-swift-ee/blob/master/.github/workflows/swift.yml

borrrden (approved):

PR #3185, opened 2023-08-24
CBL-2766: Deprecate SortOrder (with QuerySortOrder) 

velicuvlad:


velicuvlad:
Yes, indeed. I thought we are already expecting Beryllium to be the next major. Will close this PR.

PR #3190, opened 2023-09-04
CBL-4658: Enable CLANG_WARN_OBJC_IMPLICIT_RETAIN_SELF

velicuvlad:


pasin (changes requested):
There are some swift build error as a result of this change. Can you fix those errors?

pasin (approved):

PR #3191, opened 2023-09-11
Update LiteCore to 3.2.0-94

velicuvlad:


Objective-C/CBLDatabase.mm
```diff
@@ -796,7 +796,7 @@ - (nullable CBLCollection*) collectionWithName: (NSString*)name
         C4Error c4err = {};
         C4Collection* c4col = c4db_getCollection(_c4db, spec, &c4err);
         if (!c4col) {
-            if (c4err.code != 0) {
+            if (c4err.code != 7) {
```
pasin, line 799: Can we use the constant variable instead of the number value? Also we will need to check for the error domain as well.

pasin (changes requested):

Objective-C/CBLDatabase.mm
```diff
@@ -796,7 +796,7 @@ - (nullable CBLCollection*) collectionWithName: (NSString*)name
         C4Error c4err = {};
         C4Collection* c4col = c4db_getCollection(_c4db, spec, &c4err);
         if (!c4col) {
-            if (c4err.code != 0) {
+            if (c4err.code != kC4ErrorNotFound && c4err.domain == LiteCoreDomain) {
```
pasin, line 799: It should be 

```
!(c4err.code == kC4ErrorNotFound && c4err.domain == LiteCoreDomain)
```
or 

```
c4err.code != kC4ErrorNotFound || c4err.domain != LiteCoreDomain)
```

pasin (changes requested):

pasin (approved):

PR #3198, opened 2023-11-09
CBL-5035 : Fix backgrounding logic to cover conflict resolution

pasin:
* The extending background task shouldn’t be ended if there are still pending conflict resolutions.

* When the app was suspended by the background monitor, any current pending conflict resolutions should be canceled. All cancelled conflict resolutions will be notified and put into the queue again when the replicator is resumed or is restarted.

* Fixed bug that the backgrounding monitor may not be restarted after it was ended.

* Renamed some methods and add comments to code.

* Added test to test suspending conflict resolution.

### DIFF
--- a/Objective-C/CBLReplicator.mm
+++ b/Objective-C/CBLReplicator.mm
@@ -83,8 +83,9 @@ typedef enum {
     CBLChangeNotifier<CBLReplicatorChange*>* _changeNotifier;
     CBLChangeNotifier<CBLDocumentReplication*>* _docReplicationNotifier;
     BOOL _resetCheckpoint;          // Reset the replicator checkpoint
-    unsigned _conflictCount;        // Current number of conflict resolving tasks
-    BOOL _deferReplicatorNotification; // Defer replicator notification until finishing all conflict resolving tasks
+    BOOL _conflictResolutionSuspended;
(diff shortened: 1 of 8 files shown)


Objective-C/CBLReplicator.mm
```diff
@@ -521,59 +538,63 @@ - (void) c4StatusChanged: (C4ReplicatorStatus)c4Status {
...
+                // conflicts are resolved.
+                if ([self hasPendingConflicts]) {
+                    _state = kCBLStateStopping;
+                    _deferChangeNotification = YES;
+                } else {
+                    [self stopped];
+                }
+            } else if (c4Status.level == kC4Idle) {
+                // When the replicator is idle, check if there are any pending conflicts
+                // to be resolved. If none, he change listeners will be notified about
```
velicuvlad, line 577: "the" change

velicuvlad (approved):

bmeike (commented):
This is a gigantic change to a version that is end of life.
Unless it is a bug that was introduced since 2.8.x, I suggest we consider asking customers to upgrade, to get the fix.

pasin:
Confirmed from the PM that this fix will go into 3.0.x branch. The fix will be ported to the other release branches.

PR #3202, opened 2023-11-20
Update Swift Package and Podspecs to 3.1.3

velicuvlad:


pasin (approved):
Thanks for keeping up with the supported version.

PR #3205, opened 2023-12-04
Re-enable some query tests

velicuvlad:
Todo has been fixed a while back and these have not been re-enabled since


pasin:
@velicuvlad is this still a valid PR?

velicuvlad:
I forgot about this. As far I remember all the 3 tests enables were passing on local but one of them hangs in Jenkins. I will rebase this of master and have a look

PR #3223, opened 2024-02-09
Check collection equality using isEqual:

jeffdgr8:
Use `CBLCollection isEqual:` for collection equality checks in order to allow separate references to the same collection. This aligns with the behavior in the Java and C SDKs, which allow this.

Also check `CBLCollection` database instances are equal for equality check, [similar to Java SDK](https://github.com/couchbase/couchbase-lite-java-common/blob/b94e70d55289539f966057438ad22d2874310321/common/main/java/com/couchbase/lite/Collection.java#L513-L514).

### DIFF
--- a/Objective-C/CBLCollection.mm
+++ b/Objective-C/CBLCollection.mm
@@ -525,7 +525,8 @@ NSString* const kCBLDefaultCollectionName = @"_default";
     
     if (!(other && [self.name isEqual: other.name] &&
           [self.scope.name isEqual: other.scope.name] &&
-          [self.database.path isEqual: other.database.path])) {
+          // don't use isEqual: here!  The database must be the exact same instance.
+          self.database == other.database)) {
         return NO;
     }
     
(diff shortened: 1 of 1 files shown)


Objective-C/CBLCollection.mm
```diff
@@ -525,7 +525,8 @@ - (BOOL) isEqual: (id)object {
     
     if (!(other && [self.name isEqual: other.name] &&
           [self.scope.name isEqual: other.scope.name] &&
-          [self.database.path isEqual: other.database.path])) {
+          // don't use isEqual: here!  The database must be the exact same instance.
+          self.database == other.database)) {
```
pasin, line 529: Have you seen a problem when comparing the database paths?

pasin (commented):
I'm checking if there could be any issues when saving a document to a different collection object which is not the one that the document is from.

Objective-C/CBLCollection.mm
jeffdgr8: Yes, with just the first change from `!=` to `![ isEqual:]` it broke [these tests](https://github.com/couchbase/couchbase-lite-ios/actions/runs/7847831080/job/21421330937). The equality check needs to do the check for separate database instances, which then make the collections not equal.

pasin:
Thanks a lot for creating this PR. I have talked to @bmeike and @borrrden today and we have learnt that each platform has implemented different validation logic when saving the documents and in database / collection equality check. We will have tickets to sync the implementation on each platform. So let me take care this issue.

PR #3230, opened 2024-02-15
CBL-5324: Vector Index Tests 1-5

velicuvlad:
- First 5 vector search tests passing on local
- Update LiteCore to 3.2.0-137

Disabled tests so that they dont run in Jenkins as the library doesn't exist.


velicuvlad:
Once you open the project with XCode, add the vector search extension under `Frameworks` in root directory and set as follows: 
image.

Objective-C/Tests/CBLTestCase.m
```diff
@@ -131,7 +135,7 @@ - (void) reopenDB {
         Assert([_db close: &error], @"Close error: %@", error);
         _db = nil;
     }
-    [self openDB];
+    [self initDB];
```
pasin, line 138: Should call [self openDB] otherwise if calling this in the VectorSearchTest, it will do the copy again.

Objective-C/Tests/VectorSearchTest.m
```diff
@@ -0,0 +1,142 @@
...
+    CBLDatabaseConfiguration* config = [[CBLDatabaseConfiguration alloc] init];
+    config.directory = self.directory;
+    
+    NSString* path = [self databasePath: @"words_db.cblite2" inDirectory: @"vectorsearch"];
+    [CBLDatabase copyFromPath:path toDatabase: kDatabaseName withConfig: config error: &error];
+
+    [self openDB];
+}
+
+- (NSString*) databasePath: (NSString*)fileName inDirectory: (NSString*)dir {
```
pasin, line 41: Better to move this method to CBLTestCase so it can be shared with MigrationTest.

Objective-C/Tests/VectorSearchTest.m
```diff
@@ -0,0 +1,142 @@
...
+        NSLog(@"Invalid Argument test for centroids = 64001");
+    }];
+}
+
+- (void) _testCreateVectorIndex {
+    NSError* error;
+    CBLCollection* collection = [_db collectionWithName: @"words" scope: nil error: nil];
+    CBLVectorIndexConfiguration* config = [[CBLVectorIndexConfiguration alloc] initWithExpression: @"vector" dimensions: 300 centroids: 10];
+    
+    [collection createIndexWithName: @"words_index" config: config error: &error];
```
pasin, line 120: Assert the boolean result to check that the create index was successful.

```
Assert([collection createIndexWithName: @"words_index" config: config error: &error]);
```

Objective-C/Tests/VectorSearchTest.m
```diff
@@ -0,0 +1,142 @@
...
+    
+    NSArray* names = [collection indexes: &error];
+    AssertEqual(names.count, 1u);
+    AssertEqualObjects(names, (@[@"words_index"]));
+    
+    NSString* sql = @"select meta().id, word from _default.words where vector_match(words_index, $vector, 300)";
+    CBLMutableArray* dinnerArray = [[CBLMutableArray alloc] initWithData: kDinnerVector];
```
pasin, line 127: You can just do [parameters setValue: kDinnerVector forName: @"vector"].

Objective-C/Tests/VectorSearchTest.m
```diff
@@ -0,0 +1,142 @@
...
+    
+    CBLQueryParameters* parameters = [[CBLQueryParameters alloc] init];
+    [parameters setArray: dinnerArray forName:@"vector"];
+    
+    CBLQuery* q = [_db createQuery: sql error: &error];
+    [q setParameters: parameters];
+    NSString* explain = [q explain: &error];
```
pasin, line 134: More readable when having an empty line before this line as explain is not related to set parameter at all.

pasin (changes requested):

pasin:
The testCollectionConflictResolver failure is not related to this PR. I have fixed the bug in the test in 3.1 branch and that needs to be ported here.

pasin (approved):

PR #3244, opened 2024-03-01
CBL-5441 : Update EE configs to workaround XCode editor faulty error

pasin:
* XCode editor doesn’t recognize Debug-EE config set in the scheme. Also, XCode editor always choose CE’s CouchbaseLite.h regardless of the active target based on the selected scheme. As a result most of the EE types are not recongized and codes under COUCHBASE_ENTERPRISE ifdef are not active in the editor. There are no issues when built or run the tests though (only the editors having the problems).

* EE targets now have their own xcconfigs. With this change, each target’s release mode (Debug or Release) will have the same config for both EE and CE. For example, Debug and Debug_EE config for each target will be the same. This change actually makes sense because we don’t have a target that can be both CE and EE. However, we still need to have a separated CE and EE config to be compatible with LiteCore when building the project.

* Renamed xcconfig files to have the same names as their targets.

* Combined CE and EE's CouchbaseLite.h and added public CBLEdition.h to define COUCHBASE_ENTERPRISE marco definition.

* Added EE schemes to the repo instead of copying the schemes from the EE repo. This will allow us to uptake any updates to the scheme files by XCode instead of keep using the non-updated scheme files created several year ago.

* CBL-5457 : Add missing exported symbols to CouchbaseLite.exp file.

* Companion PR : https://github.com/couchbaselabs/couchbase-lite-ios-ee/pull/211


velicuvlad:
NICE!

velicuvlad (approved):

PR #3245, opened 2024-03-02
CBL-5461 : Remove ios submodule from ios-ee repo

pasin:
We don't really need couchbase-lite-ios submodule inside couchbase-lite-ios-ee repo as in build we always checkout couchbase-lite-ios submodule to the top of the master or release branch.

The reason that we have the submodule is to try to keep CE and EE repo in sync which turns out to be an unnecessary unproductivity task. We are now removing the ios submdoule from the ee repo now so that we don’t need to make them in-sync anymore.

Plus : Move download_vector_search_extension.sh script to ee repo.
Plus : Minor fix to the encryption key property in CBLDatabaseConfiguration.m to add check readonly.
Note : Need to work on updating beryllium build manifest and script in jenkins's configuration before merge.


Jenkinsfile
velicuvlad, line 1: the env var **PRODUCT** is not needed anymore in this case, so it should be removed

velicuvlad (approved):

PR #3247, opened 2024-03-05
CBL-5434: Vector Index API Tests 21-26 [Obj-C]

velicuvlad:
- as per the last spec, the whole obj-c vector index api test suite is now complete
- 16/26 are currently running, the rest being disabled and have their respective ticket at the start of the test (in a comment)

- update TestVectorMatchOnNonExistingIndex off spec to check for consistency with fts
- added 2 more index related tests to confirm behaviour - I think we can leave them here as they don't require the creation of a query such that to be moved under the query tests... and also not so many that we need to have a separate file for now

### DIFF
--- a/Objective-C/Tests/VectorSearchTest.m
+++ b/Objective-C/Tests/VectorSearchTest.m
@@ -267,7 +267,6 @@
 }
 
 - (void) testCreateVectorIndexUsingPredictionModel {
-    CBLDatabase.log.console.level = kCBLLogLevelVerbose;
     NSError* error;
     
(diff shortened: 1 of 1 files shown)


Objective-C/Tests/VectorSearchTest.m
```diff
@@ -6,6 +6,7 @@
 //  COUCHBASE CONFIDENTIAL -- part of Couchbase Lite Enterprise Edition
 //
 
+#import "CBLErrors.h"
```
pasin, line 9: This is already included in CouchbaseLite.h which is included in the CBLTestCase.h so this is not needed.

pasin (approved):

PR #3254, opened 2024-03-08
CBL-5435: Implement Tests for Vector Index API 1-10 Swift

velicuvlad:
- WordEmbeddingModel impl
- Init method changes in CBLTestCase (same was done for Obj-C) so that we can override and import words_db
- Added the VS extension framework to Swift EE tests (mac and ios) targets

- Fixed Embed&Sign setting
- Fixed building error of test target because it cant find CBLDatabaseConfiguration (obj-c type)... i dont know how this got through pipeline?
- First 10 tests for VS on Swift

### DIFF
--- a/Swift/Tests/Util/WordEmbeddingModel.swift
+++ b/Swift/Tests/Util/WordEmbeddingModel.swift
@@ -1,0 +1,50 @@
+//
+//  WordEmbeddingModel.swift
+//  CouchbaseLite
+//
+//  Copyright (c) 2024 Couchbase, Inc. All rights reserved.
+//  COUCHBASE CONFIDENTIAL -- part of Couchbase Lite Enterprise Edition
+//
+
+import Foundation
+import CouchbaseLiteSwift
+
+public struct WordEmbeddingModel: PredictiveModel {
+    
+    var db: Database!
+    
+    public init (db: Database) {
+        self.db = db;
+    }
+    
+    public func getWordVector(word: String, collection: String) -> Any? {
+        let sql = "select vector from \(collection) where word = '\(word)'"
+        let q = try! self.db.createQuery(sql)
+        let rs: ResultSet = try! q.execute()
+        let results = rs.allResults()
+    
+        if (results.count == 0) {
+            return nil;
+        }
+            
+        let result = results[0];
+        return result["vector"]
+    }
+        
+    
+    public func predict(input: CouchbaseLiteSwift.DictionaryObject) -> CouchbaseLiteSwift.DictionaryObject? {
+        guard let inputWord = input.string(forKey: "word") else {
+            fatalError("No word input !!!")
+        }
+        
+        let result = getWordVector(word: inputWord, collection: "words")
+        
+        if (result == nil) {
+            return nil
+        }
+        
+        let output = MutableDictionaryObject()
+        output.setValue(result, forKey: "vector")
+        return output
+    }
+}
--- a/Swift/Tests/Util/WordEmbeddingModel.swift
+++ b/Swift/Tests/Util/WordEmbeddingModel.swift
@@ -17,29 +17,27 @@ public struct WordEmbeddingModel: PredictiveModel {
         self.db = db;
     }
     
-    public func getWordVector(word: String, collection: String) -> Any? {
+    public func getWordVector(word: String, collection: String) -> ArrayObject? {
         let sql = "select vector from \(collection) where word = '\(word)'"
         let q = try! self.db.createQuery(sql)
         let rs: ResultSet = try! q.execute()
         let results = rs.allResults()
     
-        if (results.count == 0) {
-            return nil;
+        guard let result = results.first else {
+            return nil
         }
-            
-        let result = results[0];
-        return result["vector"]
+        
+        return result.array(forKey: "vector")
     }
         
     
-    public func predict(input: CouchbaseLiteSwift.DictionaryObject) -> CouchbaseLiteSwift.DictionaryObject? {
+    public func predict(input: DictionaryObject) -> DictionaryObject? {
         guard let inputWord = input.string(forKey: "word") else {
             fatalError("No word input !!!")
         }
-        
-        let result = getWordVector(word: inputWord, collection: "words")
-        
-        if (result == nil) {
+
+        guard let result = self.getWordVector(word: inputWord, collection: "words") ??
+                           self.getWordVector(word: inputWord, collection: "extwords") else {
             return nil
         }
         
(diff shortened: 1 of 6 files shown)


Swift/Tests/Util/WordEmbeddingModel.swift
```diff
@@ -0,0 +1,50 @@
...
+        
+    
+    public func predict(input: CouchbaseLiteSwift.DictionaryObject) -> CouchbaseLiteSwift.DictionaryObject? {
+        guard let inputWord = input.string(forKey: "word") else {
+            fatalError("No word input !!!")
+        }
+        
+        let result = getWordVector(word: inputWord, collection: "words")
+        
+        if (result == nil) {
```
pasin, line 42: Use guard let instead.

Swift/Tests/VectorSearchTest.swift
```diff
@@ -0,0 +1,400 @@
...
+    override func tearDownWithError() throws {
+        try super.tearDownWithError()
+    }
+    
+    override func initDB() throws {
+        var config = DatabaseConfiguration()
+        config.directory = self.directory
+     
+        let res = ("Support/databases/vectorsearch" as NSString).appendingPathComponent("words_db")
+        let path = Bundle(for: Swift.type(of:self)).path(forResource: res, ofType: "cblite2")
```
pasin, line 28: Move these two lines inside do { ... } as they are only needed when database doesn't exist.

Swift/Tests/VectorSearchTest.swift
```diff
@@ -0,0 +1,400 @@
...
+            - expression: "vector"
+            - dimensions: 300
+            - centroids: 20
+        2. Get and check the following property values:
+            - encoding: 8-Bit Scalar Quantizer Encoding
+            - metric: Euclidean Distance
+            - minTrainingSize: 25 * centroids
+            - maxTrainingSize: 256 * centroids
+        3. To check the encoding type, platform code will have to expose some internal property to the tests for verification.
+ */
```
pasin, line 66: Swift is using `///` for func's comment. Can you also adjust the identation to be aligned with `func`'s indentation?

Swift/Tests/VectorSearchTest.swift
```diff
@@ -0,0 +1,400 @@
...
+    func testVectorIndexConfigurationDefaultValue() {
+        let config = try! VectorIndexConfiguration(expression: "vector", dimensions: 300, centroids: 20)
```
pasin, line 68: VectorIndexConfiguration has no `throws` so `try !` is not needed. Same for the other tests.

Swift/Tests/VectorSearchTest.swift
```diff
@@ -0,0 +1,400 @@
...
+            - dimensions: 300
+            - centroids: 20
+        2. Check that the config can be created without an error thrown.
+        3. Use the config to create the index and check that the index
+ can be created successfully.
+        4. Change the dimensions to 1 and 2049.
+        5. Check that an invalid argument exception is thrown.
+ */
+    func testDimensionsValidation() {
+        let collection = try! db.collection(name: "words")
```
pasin, line 128: Make test function throws instead of doing `try!`. Same for the other test.

Swift/Tests/VectorSearchTest.swift
```diff
@@ -0,0 +1,400 @@
...
+        
+        let config1 = try? VectorIndexConfiguration(expression: "vector", dimensions: 2, centroids: 20)
```
pasin, line 130: VectorIndexConfiguration has no throws.

pasin (changes requested):

Swift/Tests/VectorSearchTest.swift
```diff
@@ -0,0 +1,628 @@
...
+    ///    3. Use the config to create the index and check that the index can be created successfully.
+    ///    4. Change the dimensions to 1 and 2049.
+    ///    5. Check that an invalid argument exception is thrown.
+
+    func testDimensionsValidation() throws {
+        let collection = try db.collection(name: "words")!
+        
+        let config1 = VectorIndexConfiguration(expression: "vector", dimensions: 2, centroids: 20)
+        
+        XCTAssertNotNil(config1)
```
pasin, line 128: Config is never be null so no need to check.

Swift/Tests/VectorSearchTest.swift
```diff
@@ -0,0 +1,628 @@
...
+        try collection.createIndex(withName: "words_index_1", config: config1)
+        
+        var names = try collection.indexes()
+        XCTAssert(names.contains("words_index_1"))
+        
+        let config2 = VectorIndexConfiguration(expression: "vector", dimensions: 2048, centroids: 20)
+        XCTAssertNotNil(config2)
```
pasin, line 135: Config is never be null so no need to check.

Swift/Tests/VectorSearchTest.swift
```diff
@@ -0,0 +1,628 @@
...
+    ///     3. Use the config to create the index and check that the index can be created successfully.
+    ///     4. Change the centroids to 0 and 64001.
+    ///     5. Check that an invalid argument exception is thrown.
+
+    func testCentroidsValidation() throws {
+        let collection = try db.collection(name: "words")!
+        
+        let config1 = VectorIndexConfiguration(expression: "vector", dimensions: 300, centroids: 1)
+        
+        XCTAssertNotNil(config1)
```
pasin, line 169: Config is never be null so no need to check.

Swift/Tests/VectorSearchTest.swift
```diff
@@ -0,0 +1,628 @@
...
+        try collection.createIndex(withName: "words_index_1", config: config1)
+
+        var names = try collection.indexes()
+        XCTAssert(names.contains("words_index_1"))
+        
+        let config2 = VectorIndexConfiguration(expression: "vector", dimensions: 300, centroids: 64000)
+        
+        XCTAssertNotNil(config2)
```
pasin, line 177: Config is never be null so no need to check.

Swift/Tests/VectorSearchTest.swift
```diff
@@ -0,0 +1,628 @@
...
+    ///           WHERE vector_match(words_index, <dinner vector>, 20)
+    ///     6. Check the explain() result of the query to ensure that the "words_index" is used.
+    ///     7. Execute the query and check that 20 results are returned.
+
+    func testCreateVectorIndex() throws{
+        let collection = try db.collection(name: "words")!
+        
+        let config1 = VectorIndexConfiguration(expression: "vector", dimensions: 300, centroids: 1)
+        
+        XCTAssertNotNil(config1)
```
pasin, line 215: Config is never be null so no need to check.

pasin (approved):
Please take care a few things then merge.

PR #3264, opened 2024-03-27
CouchbaseLite 3.2.0 Beta 1 package + readme

velicuvlad:


velicuvlad:
Dependency check fails as the head of LiteCore is not `release/3.2`. That was created after RCs were submitted. `release/3.2` for LiteCore exists now and will be update for beta 2

pasin (approved):

PR #3271, opened 2024-04-16
CBL-5365: Remove CBLErrors.h from Swift Public API

velicuvlad:
- EE pair: https://github.com/couchbaselabs/couchbase-lite-ios-ee/pull/222
- wish that it existed another way rather than importing this almost in every implementation file... but we do relate to obj-c types for most of them

Note: I've found _@testable_ flag and that it can be used to access only private methods in tests, I will probably add this next sprint to more files such that we can add tests for those to improve code coverage


Swift/Tests/CollectionTest.swift
```diff
@@ -18,7 +18,8 @@
 //
 
 import XCTest
-import CouchbaseLiteSwift
+@testable import CouchbaseLiteSwift
+@testable import CouchbaseLiteSwift_Private
```
pasin, line 22: CouchbaseLiteSwift_Private mostly contains objective-c classes. Is it required here in Swift Tests?

velicuvlad: yes as some tests check certain values with <CBL...> values

pasin (approved):

PR #3272, opened 2024-04-19
CBL-5508: Update Min macOS Support Version to 12.0

velicuvlad:
- these changes fix building issues when targeting macOS 12
- update LiteCore to current release/3.2 to pick some required changes as well
- pair https://github.com/couchbaselabs/couchbase-lite-ios-ee/pull/223


velicuvlad:
Remove MYAnonymousIdentity altogether from the project as is not used.o Only x86_64 complained on compile, I am still not certain why... seems like it might the case that it is not deprecated on Catalyst, but on x86_64 only.

velicuvlad:
(([[rs allObjects] count]) equal to (0u)) failed: ("1") is not equal to ("0") [2024-04-22T14:57:47.191Z] Test Case '-[QueryTestWithMeta testExpiryNoGreaterThanDate]' failed (0.447 seconds).

xcconfigs/Project.xcconfig
```diff
@@ -22,7 +22,7 @@ CBL_BUILD_NUMBER                   = 0
 CBL_COPYRIGHT_YEAR                 = 2021
 
 IPHONEOS_DEPLOYMENT_TARGET         = 12.0
-MACOSX_DEPLOYMENT_TARGET           = 10.14
+MACOSX_DEPLOYMENT_TARGET           = 12
```
pasin, line 25: Can you change it to 12.0 for format consistency?

pasin (approved):

PR #3275, opened 2024-04-22
CBL-5659 : Fix a released query context may be used in observer callback

pasin:
Implemented CBLContextManager class for retaining and mapping the object with its pointer value which can be used as the context for LiteCore's callbacks (e.g. use when creating c4queryobserver objects). The implementation simply stores the object in a map by using its memory address as the key and returns the memory address as the pointer value.

Updated CBLQueryObserver to use CBLContextManager to register its query context and unregister the context when the C4QueryObserver is disabled. The C4QueryObserver's callback now can verified that the query context is still valid before using it. Also when using the query context, the query context is retained in the callback itself.

Updated CBLQueryObserver to get the C4QueryEnumerator object inside the callback as the enumerator object to prevent using the already freed C4QueryObserver in the notification queue.

Added a test for [CBSE-16662](https://issues.couchbase.com/browse/CBSE-16662) which is related to this issue and two tests to check that there are no notification received without crash ([CBL-5659](https://issues.couchbase.com/browse/CBL-5659)) after removing the token.

Added an internal debug build only C4QueryObserverCallbackDelayInterval config for testing the fix.

Found that the retain cyble between CBLQueryObserver and CBLQuery and CBLListenerToken can be improved.

Updated LiteCore to 3.1.7-1 to get the fix for [CBSE-16662](https://issues.couchbase.com/browse/CBSE-16662).

### DIFF
--- a/Objective-C/Tests/QueryTest+Main.m
+++ b/Objective-C/Tests/QueryTest+Main.m
@@ -31,6 +31,10 @@
 #import "Foundation+CBL.h"
 #import "CollectionUtils.h"
 
+#ifdef DEBUG
+#import "CBLQueryObserver.h"
+#endif
+
 @interface QueryTest_Main : QueryTest
 
 @end
@@ -40,6 +44,13 @@
 // TODO: Remove https://issues.couchbase.com/browse/CBL-3206
 #pragma clang diagnostic push
 #pragma clang diagnostic ignored "-Wdeprecated-declarations"
+
+- (void) tearDown {
+#ifdef DEBUG
+    [CBLQueryObserver setC4QueryObserverCallbackDelayInterval: 0.0];
+#endif
+    [super tearDown];
+}
 
 #pragma mark - Where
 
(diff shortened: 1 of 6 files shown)


Objective-C/Tests/QueryTest+Main.m
```diff
@@ -2061,6 +2072,145 @@ - (void) testLiveQueryUpdateQueryParam {
...
+        } else {
+            [changedExp fulfill];
+        }
+    }];
+    
+    [self waitForExpectations: @[changedExp] timeout: 5.0];
+    
+    // Remove the token:
+    tokenRemoved = YES;
+    [token remove];
```
velicuvlad, line 2154: If `remove` is not successful, will the test/app crash? If not, we might want to invert these 2 so we set `tokenRemoved = YES` after we actually successfully remove it.

Objective-C/Tests/QueryTest+Main.m
```diff
@@ -2061,6 +2072,145 @@ - (void) testLiveQueryUpdateQueryParam {
...
+        
+        // Add more docs:
+        [self createDocNumbered: 101 of: 200];
+        
+        // Wait a little for the query to pickup the change:
+        [NSThread sleepForTimeInterval: 1.0];
+        
+        // Remove the token:
+        tokenRemoved = YES;
+        [token remove];
```
velicuvlad, line 2205: same here

velicuvlad (approved):

PR #3278, opened 2024-04-25
CBL-5222: MutableDocument should be usable before creating a database instance

velicuvlad:
User can now create a MutableDocument prior of defining a database object as it will initialize logging by itself if that's the case.

### DIFF
--- a/Objective-C/Internal/CBLLog+Admin.h
+++ b/Objective-C/Internal/CBLLog+Admin.h
@@ -1,5 +1,5 @@
 //
-//  CBLLogg+Admin.h
+//  CBLLog+Admin.h
 //  CouchbaseLite
 //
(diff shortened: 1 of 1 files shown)


Objective-C/Internal/CBLLog+Admin.h
```diff
@@ -1,8 +1,8 @@
 //
-//  CBLLogging+Admin.h
+//  CBLLogg+Admin.h
```
pasin, line 2: Typo with two gg.

pasin (approved):

velicuvlad:
Some tests started to be flaky, I can't figure out why yet... it is only happening on iOS Actions. It is not the same tests every time and all of them are passing on my machine via simulator.

PR #3281, opened 2024-05-01
CBL-5693 : Fix missing exported symbols

pasin:
* Ported the fix from release/3.1 branch.

* Used the same approach as CBL-C to generate CE and EE exported symbole file. The workflow will be updating CBL.txt or CBL_EE.txt (Contains EE only symbols) at Objective-C/Exports folder and run generate_exports.sh script. The final export symbol files will be in the Objective-C/Exports/Generated folder.

* Ensured to include all symbols by cross checking with the symbols extracted from the header files.

* Removed obsoleted predefined classes which don’t exist anymore.


pasin:
The test failures seems to be flaky and is not related to this PR.

velicuvlad:
This is the flaky I've told you about - `LITECORE OBJECTS STILL NOT FREED`. I've ran it a couple of time on both mac and ios and I couldn't reproduce

velicuvlad (approved):

pasin:
Some times things are freed in a background thread and when we checked that there are any objects left in the memory, they are not freed yet. Need to look at the test to see if it's because of that or not and see if there are things that can prevent the flakyness.

PR #3284, opened 2024-05-09
CBL-5613: Swift API docs for Scope using Obj-c reference

velicuvlad:


Swift/Collection.swift
```diff
@@ -2,7 +2,7 @@
 //  Collection.swift
 //  CouchbaseLite
 //
-//  Copyright (c) 2022 Couchbase, Inc All rights reserved.
+//  Copyright (c) 2024-present Couchbase, Inc All rights reserved.
```
pasin, line 5: 2024-Present


Can you change to cap 'P'?

pasin (approved):

Swift/Collection.swift
pasin, line 5: IIRC, we have received this format from legal and all CBL platforms are using the same format. 

Before changing the format even though it is small and ok, it would be best to check with the team.

Also we would prefer to have one PR to change the format for all files for consistency.

pasin (changes requested):

Swift/Collection.swift
velicuvlad: I've talked with SGW and they use a tool for updating the copyright paragraph and also to generate one for every new file created. very interesting. I'll look into it if time permits.

I've revert the license date changes.

bmeike: Yeah.... I if "present" works, it is the first I've heard of it.   It would be very nice if we can use it, though.  Please let me know how this goes....

bmeike (approved):
you get all sorts of brownie points for doing this stuff...

pasin (approved):

PR #3285, opened 2024-05-09
CBL-5660 : Fix a released query context may be used in observer callback

pasin:
* Ported the fix directly from release/3.1 branch.

* Implemented CBLContextManager class for retaining and mapping the object with its pointer value which can be used as  the context for LiteCore's callbacks (e.g. use when creating c4queryobserver objects). The implementation simply stores the object in a map by using its memory address as the key and returns the memory address as the pointer value.

* Updated CBLQueryObserver to use CBLContextManager to make sure that the released query context can be detected and not be used. When using the query context, the query context is retained.

* Added a test for CBSE-16662 which is related to this issue and two tests to check that there are no notification received without crash (CBL-5660) after removing the token.

* Added an internal debug build only C4QueryObserverCallbackDelayInterval config for testing the fix.


pasin:
After all, need a new LiteCore that includes https://github.com/couchbase/couchbase-lite-core/pull/2025.

velicuvlad:
will merge after https://github.com/couchbase/couchbase-lite-ios/pull/3288

velicuvlad (approved):

pasin:
A failure doc expiry test is not related to this PR.

PR #3287, opened 2024-05-14
CBL-5514: Swift MutableDocument's collection is not set when a new document is saved

velicuvlad:
- change Document.collection API to have a private setter
- set collection on save
- add 1 tests for each side (Obj-C and Swift)

### DIFF
--- a/Swift/Collection.swift
+++ b/Swift/Collection.swift
@@ -107,6 +107,7 @@ public final class Collection : CollectionChangeObservable, Indexable, Equatable
     /// Throws an NSError with the CBLError.notOpen code, if the collection is deleted or
     /// the database is closed.
     public func save(document: MutableDocument) throws {
+        document.collection = self
         try impl.save(document.impl as! CBLMutableDocument)
     }
     
--- a/Swift/Collection.swift
+++ b/Swift/Collection.swift
@@ -107,10 +107,10 @@ public final class Collection : CollectionChangeObservable, Indexable, Equatable
     /// Throws an NSError with the CBLError.notOpen code, if the collection is deleted or
     /// the database is closed.
     public func save(document: MutableDocument) throws {
-        if (document.collection == nil){
+        try impl.save(document.impl as! CBLMutableDocument)
+        if (document.collection == nil) {
             document.collection = self
         }
-        try impl.save(document.impl as! CBLMutableDocument)
     }
     
     /// Save a document into the collection with a specified concurrency control. When specifying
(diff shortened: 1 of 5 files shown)


velicuvlad:
Tests failing is unrelated to this PR. Somewhat flaky as it seems the timeout period is not big enough... all of them pass on local. Will investigate as part of the code coverage ticket.

bmeike (approved):
Be sure you get confirmation from someone who actually knows what they are doing, for this.  There was *lots* of discussion during the original implementation of Collections, about whether the platform code was allowed to hold the reference to one, or not.  I believe the issue is that the thing to which you hold a reference might become invalid, behind your back.    Worth checking.

velicuvlad:
I can confirm that this is behaviour is already there for Obj-C. The Obj-C tests are just to confirm the behaviour.

A new MutableDoc is getting created without any collection ref, meanwhile it gets one when it is saved (as part of the process). I initially thought the change would be bigger as I didn't know that Swift has private setters (Pasin made me aware - kudos). For Swift the collection attribute was non-modifiable as it was a `let`.

What exactly do you refer as getting invalid? The C4Collection object? We do not allow the change of `doc.collection` so in the case you want to move a document, a brand new copy must be made anyway... I would expect that if the collection is invalid by magic, as it cannot be changed, that document is pretty much orphan and the solution becomes the same -> make a brand new copy of it

When we pull a doc out of the db, it does come attached with a collection so we already hold ref to it. I can't think of a case for which that collection would be invalided at the exact time of the save, when save is a collection operation - so if doesn't exist it will fail on save in the first place

Swift/Collection.swift
```diff
@@ -107,6 +107,9 @@ public final class Collection : CollectionChangeObservable, Indexable, Equatable
     /// Throws an NSError with the CBLError.notOpen code, if the collection is deleted or
     /// the database is closed.
     public func save(document: MutableDocument) throws {
+        if (document.collection == nil){
+            document.collection = self
+        }
```
pasin, line 112: This should be done after saving is successful.

Swift/Collection.swift
```diff
@@ -107,6 +107,9 @@ public final class Collection : CollectionChangeObservable, Indexable, Equatable
     /// Throws an NSError with the CBLError.notOpen code, if the collection is deleted or
     /// the database is closed.
     public func save(document: MutableDocument) throws {
+        if (document.collection == nil){
```
pasin, line 110: Format: A space after ).

Swift/Collection.swift
pasin: There are another two save functions that need set the collection to the saved doc as well.

pasin (changes requested):

pasin (approved):

PR #3289, opened 2024-05-28
CBL-5668: Get Code Coverage at least 80%

velicuvlad:
- add a couple of tests for some API methods that weren't already tested.
- update LiteCore to fix orphan commit
- update VS ext

This is what I've found to be the fastest way to increase the code coverage. There's a big portion of MessageEndpoint class in Swift that is not tested at all but I have to give it time to understand how that works and how to properly test it as there's some setup from what I could grasp from the Obj-C code lines. 

This should get Swift to 80% or very very close. It takes a good amount of time to generate the whole release zip on local for all platforms so that it generates the code coverage as well. Hence, if needed, I will follow up with another PR after the new API impl as there will be new tests there as well.

### DIFF
--- a/Swift/Tests/DictionaryTest.swift
+++ b/Swift/Tests/DictionaryTest.swift
@@ -23,6 +23,8 @@ import CouchbaseLiteSwift
 
 class DictionaryTest: CBLTestCase {
     
+    let kTestDate = "2017-01-01T00:00:00.000Z"
+    
     func testCreateDictionary() throws {
         let address = MutableDictionaryObject()
         XCTAssertEqual(address.count, 0)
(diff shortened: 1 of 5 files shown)


bmeike (commented):
There is a separate ticket to port this up to Boron?
Normally, the idea is to keep changes to a release branch to things that block the release.  While not strictly a blocker, this is pertinent to the release.  Go for it. 
I can't comment on the code itself.  Pasin can approve.

bmeike:
Right.   I completely agree.  These are tests for exactly the key feature of the 3.2 release.  Do it.

Swift/Tests/DictionaryTest.swift
```diff
@@ -75,6 +77,47 @@ class DictionaryTest: CBLTestCase {
         })
     }
     
+    func testSettersDictionary() throws {
+        let dict = MutableDictionaryObject()
+        dict.setValue("value", forKey: "key")
+        XCTAssertNotNil(dict.value(forKey: "key"))
```
pasin, line 83: The test should check if the set value is equals to the get value.

Swift/Tests/DictionaryTest.swift
```diff
@@ -75,6 +77,47 @@ class DictionaryTest: CBLTestCase {
...
+        dict.setValue(0, forKey: "key")
+        
+        dict.setString("string", forKey: "key")
+        XCTAssertNotNil(dict.string(forKey: "key"))
+        
+        dict.setNumber(1, forKey: "key")
+        XCTAssertNotNil(dict.number(forKey: "key"))
+        
+        dict.setInt(0, forKey: "key")
```
pasin, line 92: I would just non zero number as the getter will return zero for non integer as well.

Swift/Tests/DictionaryTest.swift
```diff
@@ -75,6 +77,47 @@ class DictionaryTest: CBLTestCase {
...
+        XCTAssertNotNil(dict.int(forKey: "key"))
+        
+        dict.setInt64(Int64("0")!, forKey: "key")
```
pasin, line 95: Use non zero number value instead of string "0".

Swift/Tests/DictionaryTest.swift
```diff
@@ -88,7 +131,7 @@ class DictionaryTest: CBLTestCase {
         
         let level3 = MutableDictionaryObject()
         level3.setValue("n3", forKey: "name")
-        level2.setValue(level3, forKey: "level3")
+        level2.setDictionary(level3, forKey: "level3")
```
pasin, line 134: This is not concistent as the rest is using setDictionary.

Swift/Tests/DocumentTest.swift
```diff
@@ -48,7 +48,7 @@ class DocumentTest: CBLTestCase {
         let array = MutableArrayObject()
         array.addValue("650-123-0001")
         array.addValue("650-123-0002")
-        doc.setValue(array, forKey: "array")
+        doc.setArray(array, forKey: "array")
```
pasin, line 51: It's not consistent to use setArray here while the other places in this test use setValue.

Swift/Tests/DictionaryTest.swift
pasin: You could convert this tests to all use setDictionary.

pasin (changes requested):

Swift/Tests/DictionaryTest.swift
velicuvlad: I thought of ... but then there's no other test for setValue with a dict value, so I thought interchanging them would prove they behave the same and as expected.

Swift/Tests/DocumentTest.swift
velicuvlad: same here -  I thought interchanging them would prove they behave the same and as expected. Most of the tests have originally been made using setValue for ease of use/read.  

setValue is used in most of the other tests whenever we manipulate a document, so I can certainly change all these to use setArray.

Swift/Tests/DictionaryTest.swift
pasin: Then it’s better to have different tests to test them both. Objective-C test has a large number of tests that test all getters and setters. You may check and design what should be the best and optimized test cases for swift.

pasin: One idea would be for specific tests that test dictionary, we can use dictionary methods. Then we can have a spec test to test get/set value method.

Swift/Tests/DocumentTest.swift
velicuvlad: I realized this was a leftover... reverted

velicuvlad:
I've made a couple of other modifications. I figured that `populateData()` uses `setValue()` for all types. All Document `getters` tests do use `populateData()` and then type specific getters. Thus, for the `setters` tests, I changed everything to type specific setters.

pasin (approved):

PR #3290, opened 2024-06-03
CBL-3385: Allow null expression parameter for Function.count(expression)

velicuvlad:


velicuvlad:
failing tests unrelated, there's a hope to address these in a another, already existing, PR.

Swift/Function.swift
```diff
@@ -39,8 +39,12 @@ public final class Function {
     ///
     /// - Parameter expression: The expression.
     /// - Returns: The COUNT(expr) function.
-    public static func count(_ expression: ExpressionProtocol) -> ExpressionProtocol {
-        return QueryExpression(CBLQueryFunction.count(expression.toImpl()))
+    public static func count(_ expression: ExpressionProtocol? = nil) -> ExpressionProtocol {
+        if let expression = expression {
```
bmeike, line 43: I'm not sure what idiomatic Obj-C looks like so ignore this if it is irrelevant.
I would be inclined to have the value of the `let` be either expression.toImpl() or expression.property('*) and to have a single return statement.

bmeike (approved):
LGTM

Swift/Function.swift
```diff
@@ -39,8 +39,12 @@ public final class Function {
...
+            return QueryExpression(CBLQueryFunction.count(expression.toImpl()))
+        } else {
+            return QueryExpression(CBLQueryFunction.count(Expression.property("*").toImpl()))
```
pasin, line 46: If expression is nil, nil can be passed to Objective-C count directory.

```
return QueryExpression(CBLQueryFunction.count(expression != nil ? expression.toImpl() : nil))
```

pasin: Need to update the `+ (CBLQueryExpression*) count: (CBLQueryExpression*)expression;` signature to allow null as well.

pasin (changes requested):

pasin (approved):

PR #3302, opened 2024-07-02
CBL-5893: Throw exception for everything if finish() was successfully called beforehand

velicuvlad:
- Included Test 27 into Test 26: `count()`, `valueAtIndex()`, `setVector()`, `skipVector()` and `finish()` will throw an NSException if the updater is already finished.
- Updated to LiteCore 3.2.0-211

### DIFF
--- a/Objective-C/Tests/VectorLazyIndexTest.m
+++ b/Objective-C/Tests/VectorLazyIndexTest.m
@@ -1244,9 +1244,10 @@
  *     - Get the word string from the IndexUpdater.
  *     - Query the vector by word from the _default.words collection.
  *     - Convert the vector result which is an array object to a platform's float array.
- *     - Call setVector() with the platform's float array at the index..
- * 8. Call finish() and check that the finish() is successfully called.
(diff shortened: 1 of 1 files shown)


bmeike (approved):
Are these documented tests now?  There was a (somewhat silly) rule, at one point, that nobody was allowed to test anything that everybody didn't test.  These tests are obviously useful, so they should be in the test spec, if they are not.

velicuvlad:
@bmeike Spec was update last night so, in short yes. - Test 27. Also, as noted in test 27 description, itself and test 26 can be merged into a single on and this is what I've done in this PR. Now, it might be the case that I chose the wrong exception to throw, so I will be waiting for Pasin's review. `IllegalStateArgumentException` I can't find this exact exception, so I've used `NSInternalInconsistencyException`

bmeike:
Boy I hate these changes....

pasin:
We had have a long discussion in Slack last week about this. It was a conclusion that we will make it illegal to use the index updater after calling finish. Sorry, I didn't have time to update the test spec after the discussion. For iOS, it's correct to throw NSInternalInconsistencyException.

borrrden (approved):

PR #3309, opened 2024-07-23
CBL-5329 : Fix Replicator ListenerToken.remove() (#3213)

pasin:
* Fixed the bug that the Replicator’s ListenerToken was created without the CBLRemovableListenerToken delegate being set.

* Ported the [fix](https://github.com/couchbase/couchbase-lite-ios/commit/b17116c75033b38474b37446a44ee4aa38f94685) from release/3.2 branch.


pasin:
The x2 was changed to be an inverted check (exp2.inverted = YES) so if it's fullfill within the given timeout, the test will fail. It's a better check than using XTCFail.

PR #3310, opened 2024-07-23
CBL-6023 : Sometimes query becomes nil before is stopped

pasin:
* Make sure the query is retained and not nil when using it.

* Port the [fix](https://github.com/couchbase/couchbase-lite-ios/pull/3292/commits/7656730f3f22ded3e02454f37627fbb176eeaa59) from releae/3.2 branch.


bmeike (approved):
Ahh!  Took me a minute to see that.

PR #3311, opened 2024-07-23
CBL-6024: Swift MutableDocument's collection is not set when a new document is saved

pasin:
Port the [fix](https://github.com/couchbase/couchbase-lite-ios/commit/87d307b0b8c005a2dc29d52691e0a1cf33c20af1) from release/3.2 branch.

### DIFF
--- a/Swift/Document.swift
+++ b/Swift/Document.swift
@@ -42,7 +42,7 @@ public class Document : DictionaryProtocol, Equatable, Hashable, Sequence {
     }
     
     /// The collection that the document belongs to.
-    public let collection: Collection?
+    internal(set) public var collection: Collection?
     
     // MARK: Edit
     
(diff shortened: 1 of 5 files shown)


Swift/Document.swift
```diff
@@ -42,7 +42,7 @@ public class Document : DictionaryProtocol, Equatable, Hashable, Sequence {
     }
     
     /// The collection that the document belongs to.
```
bmeike, line 44: I don't know swift.  Does this create a publicly mutable property?  Is that what internal(set) prevents?

bmeike (approved):
I'm kind of surprised that this ever worked... ;-P

Swift/Document.swift
pasin: Yes, that is what `internal(set) prevents` prevents it to be public.

PR #3323, opened 2024-08-19
CBL-5396: Use os_log instead of NSLog for Console Log

velicuvlad:


pasin:
For some reason, a test was hang for 14 hour. I just canceled it.

Objective-C/Tests/Util/CBLWordEmbeddingModel.m
```diff
@@ -40,7 +40,8 @@ - (CBLDictionary*) predict: (CBLDictionary*)input {
     NSString* inputWord = [input stringForKey: @"word"];
     
     if (!inputWord) {
-        NSLog(@"No word input !!!");
+        os_log_t log = os_log_create("CouchbaseLite", "OSLogging");
+        os_log(log, "No word input !!!");
```
pasin, line 44: For this one we can use NSLog, I guessed or change it to assertion instead of logging.

Objective-C/CBLConsoleLogger.m
```diff
@@ -45,7 +45,8 @@ - (void) logWithLevel: (CBLLogLevel)level domain: (CBLLogDomain)domain message:
     
     NSString* levelName = CBLLog_GetLevelName(level);
     NSString* domainName = CBLLog_GetDomainName(domain);
-    NSLog(@"CouchbaseLite %@ %@: %@", domainName, levelName, message);
+    os_log_t log = os_log_create("CouchbaseLite", "OSDebug");
```
pasin, line 48: What is "OSDebug"? Where does it appear in the log?

Objective-C/CBLDatabase.mm
```diff
@@ -115,7 +115,8 @@ to be done in CBLInit()which is called only once when the first CBLDatabase is c
  */
 + (void) initialize {
     if (self == [CBLDatabase class]) {
-        NSLog(@"%@", [CBLVersion userAgent]);
+        os_log_t log = os_log_create("CouchbaseLite", "OSLogging");
```
pasin, line 118: What is "OSLogging"? Where does it appear in the log?

Objective-C/CBLConsoleLogger.m
```diff
@@ -45,7 +45,8 @@ - (void) logWithLevel: (CBLLogLevel)level domain: (CBLLogDomain)domain message:
...
+    os_log(log, "CouchbaseLite %@ %@: %@", domainName, levelName, message);
```
pasin, line 49: There are different types of os_log function based on the log level. I think we should use those functions. What is .NET using?

Objective-C/CBLConsoleLogger.m
velicuvlad: `os_log_create(const char *subsystem, const char *category)` is a place holder for category. These will not be present in the logs as we know them, but only on debugging via XCode. The userAgent will have the same exact shape/info. So I just put a placeholder without any meaning.

I can see that Jim did `private CoreFoundation.OSLog _logger = new CoreFoundation.OSLog("CouchbaseLite", "dotnet");` so maybe I should change to "Platform" instead of "OSDebug" and "OSLogging" ? IMO having the platform name is not relevant as it should be already known what platform it is.

Objective-C/CBLConsoleLogger.m
velicuvlad: .NET is mapping what we have to os_log as
```new Dictionary<LogLevel, CoreFoundation.OSLogLevel>
            {
                [LogLevel.Debug] = CoreFoundation.OSLogLevel.Debug,
                [LogLevel.Verbose] = CoreFoundation.OSLogLevel.Info, // No verbose level in Apple
                [LogLevel.Info] = CoreFoundation.OSLogLevel.Info,
                [LogLevel.Warning] = CoreFoundation.OSLogLevel.Error, // No warning level in Apple
                [LogLevel.Error] = CoreFoundation.OSLogLevel.Error,
            };
  ```          
  
My understanding is that these NSLog are only informative and have nothing to do with the actual CBL Logging, thus a similar change with .NET is not needed

File: Objective-C/Tests/Util/CBLWordEmbeddingModel.m (hunk 1 above)
Comment (velicuvlad) [reply Objective-C/Tests/Util/CBLWordEmbeddingModel.m:44]: will change to assert

File: Objective-C/CBLConsoleLogger.m (hunk 2 above)
Comment (velicuvlad) [reply Objective-C/CBLConsoleLogger.m]: changed to `com.couchbase.lite.ios` and the category now represents a domain (CBL)

**Review by borrrden — APPROVED** [review]

File: Objective-C/CBLDatabase.mm [hunk 5]
```diff
@@ -115,7 +114,8 @@ to be done in CBLInit()which is called only once when the first CBLDatabase is c
  */
 + (void) initialize {
     if (self == [CBLDatabase class]) {
-        NSLog(@"%@", [CBLVersion userAgent]);
+        os_log_t logger = os_log_create("com.couchbase.lite.ios", "UserAgent");
+        os_log(logger, "%@", [CBLVersion userAgent]);
```
pasin, line 118: I think you can add a static method at CBLConsoleLog to log the message regardless of the log level so that there is only one place that manages the logger? The category could be just "Database" as well.

Objective-C/CBLLog.mm
```diff
@@ -155,11 +155,12 @@ - (instancetype) initWithDefault {
 #ifdef DEBUG
         // Check if user overrides the default callback log level:
         NSString* userLogLevel = [NSUserDefaults.standardUserDefaults objectForKey: @"CBLLogLevel"];
+        oslogger = os_log_create("com.couchbase.lite.ios", "DebugBuild");
         if (userLogLevel) {
             callbackLogLevel = string2level(userLogLevel);
         }
         if (callbackLogLevel != kC4LogWarning) {
-            NSLog(@"CouchbaseLite minimum log level is %s", kLevelNames[callbackLogLevel]);
+            os_log(oslogger, "CouchbaseLite minimum log level is %s", kLevelNames[callbackLogLevel]);
```
pasin, line 163: Same as the comment above. The category could be "Log".

pasin (changes requested):

Objective-C/CBLConsoleLogger.m
```diff
@@ -79,4 +80,17 @@ - (void) initializeOSLogDomains {
     osLogDictionary[@(kCBLLogDomainListener)] = os_log_create("com.couchbase.lite.ios", "Listener");
     #endif
 }
+
++ (os_log_t) internalLogger {
+    if(!logger){
+        logger = os_log_create("com.couchbase.lite.ios", "Log");
```
pasin, line 86: This is not thread safe. You can use `dispatch_once` which is thread-safe to make it singleton.

pasin (changes requested):

Objective-C/CBLConsoleLogger.m
```diff
@@ -43,9 +46,51 @@ - (void) logWithLevel: (CBLLogLevel)level domain: (CBLLogDomain)domain message:
...
+            return OS_LOG_TYPE_ERROR;
+        default:
+            return OS_LOG_TYPE_DEFAULT; // Default log type
+    }
+}
+
+- (void) initializeOSLogDomains {
+    osLogDictionary = [NSMutableDictionary dictionary];
+        
+    osLogDictionary[@(kCBLLogDomainDatabase)] = os_log_create("com.couchbase.lite.ios", "Database");
```
pasin, line 74: We should also get "com.ouchbase.lite.ios" from the framework bundle identifier instead of hard code the string here.

pasin (changes requested):

Objective-C/CBLConsoleLogger.m
```diff
@@ -32,6 +32,7 @@ - (instancetype) initWithLogLevel: (CBLLogLevel)level {
     if (self) {
         _level = level;
         _domains = kCBLLogDomainAll;
+        _sysID = [[NSBundle mainBundle] bundleIdentifier];
```
pasin, line 35: I'm not sure if this is right as it think it will return the bundle identifier of the app not the framework. What is the value when you run the test? If this returns bundle id of the app, you may try to get the bundle id from one of the classes such as CBLConsoleLogger itself.

Objective-C/CBLConsoleLogger.m
```diff
@@ -70,21 +71,22 @@ static os_log_type_t osLogTypeForLevel(CBLLogLevel level) {
...
     #endif
 }
 
 + (os_log_t) internalLogger {
-    if(!logger){
-        logger = os_log_create("com.couchbase.lite.ios", "Log");
-    }
+    static dispatch_once_t onceToken;
+    dispatch_once(&onceToken, ^{
+        logger = os_log_create("com.couchbase.lite.ios", "Internal");
```
pasin, line 88: This one still uses the hardcode string, "com.couchbase.lite.ios".

pasin (changes requested):

Objective-C/CBLConsoleLogger.m
velicuvlad: you are right, will fix

velicuvlad:
I've changed `internalLogger` to `logAlways` which will log under DB domain in the default scope of os_log (hence the name), independently of the log level set by CBL, that is only used in `logWithLevel`. Also tested a bit and it works as expected.

Objective-C/CBLLog.mm
```diff
@@ -190,7 +191,7 @@ - (instancetype) initWithDefault {
                 C4LogDomain domain = c4log_getDomain(domainName, true);
                 C4LogLevel level = string2level(defaults[key]);
                 c4log_setLevel(domain, level);
-                NSLog(@"CouchbaseLite logging to %s domain at level %s", domainName, kLevelNames[level]);
+                os_log(oslogger, "CouchbaseLite logging to %s domain at level %s", domainName, kLevelNames[level]);
```
pasin, line 194: Should it use CBLConsoleLogger's logAlways as well?

pasin (approved):

Objective-C/CBLConsoleLogger.m
```diff
@@ -43,9 +47,43 @@ - (void) logWithLevel: (CBLLogLevel)level domain: (CBLLogDomain)domain message:
...
+    osLogDictionary[@(kCBLLogDomainReplicator)] = os_log_create([_sysID UTF8String], "Replicator");
+    osLogDictionary[@(kCBLLogDomainNetwork)] = os_log_create([_sysID UTF8String], "Network");
+    
+    #ifdef COUCHBASE_ENTERPRISE
+    osLogDictionary[@(kCBLLogDomainListener)] = os_log_create([_sysID UTF8String], "Listener");
+    #endif
+}
+
++ (void) logAlways: (NSString*)message {
+    os_log(osLogDictionary[@(kCBLLogDomainDatabase)], "%@", message);
```
pasin, line 86: Just noticed this. If `CBLConsoleLogger` is not inited yet, osLogDictionary will be empty.

pasin (changes requested):

PR #3327, opened 2024-09-06
CBL-5180: Implement Array Index API

velicuvlad:


Objective-C/CBLArrayIndexConfiguration.m
```diff
@@ -0,0 +1,62 @@
...
+
+- (instancetype) initWithPath: (NSString*) path
+                  expressions: (nullable NSArray<NSString*>*)expressions {
+    
+    if(!expressions){
+        self = [super initWithIndexType: kC4ArrayIndex
+                            expressions: @[@""]];
+    } else {
+        for (NSString* expression in expressions) {
+                if (!expression) {
```
pasin, line 35: * NSArray/NSMutableArray doesn't allow nil so there is no need to check this for iOS.
* The only thing that is needed to evaluate is whether expressions is empty or not.

pasin (changes requested):

borrrden (approved):
Probably fine but I also check that there is at least one non-empty element in the array before sending it to Core.

As you know, [""] works, but I would guess more often than not this is an error in the user's logic that ended up with empty string, rather than the user purposefully sending an empty string in the array (as opposed to nil).  If there is at least one non-empty string in the list then any empty strings will then be caught and errored by Core.

velicuvlad:
@borrrden if is nil, I pass to Core [""] anyway. IMO, if we go ahead and check whether a non-empty list has any empty values, we might as well fail straight away in the case that is nil as well. Which then makes us handle most, if not all, scenarios, instead of letting Core to do it.

velicuvlad:
`["class", "", "name"]` will result into `class,,name` string representing n1ql. Now, as n1ql, that is invalid, but Jens mentioned on slack that as long as we pass a comma sep string to Core, it should be good enough for Core to sort out if is good or bad.

I've chat with Jianmin, and confirmed that will throw as invalid n1ql, which can be confusing depending of it actually showing where the problem is, but my understanding is that Core shouldn't even query with the above from Jens, but throw. I will add this additional check if it ends up not being covered by Core.

pasin:
@velicuvlad 
> I know for nil passed, the path must contain scalar values. If that's not the case, I guess LiteCore throws? (we can't figure that out on client side so we don't care)

LiteCore will index the complex value which will turn out to be unuseful. There will be no error.

pasin (approved):

PR #3330, opened 2024-10-02
CBL-5181: Array Index API tests and adjustments

velicuvlad:
- updated LiteCore to 3.2.1-5
- moved `Internal` section at the end of file for `Objective-C/CBLCollection.mm`
- implement internal `indexesInfo` which is exactly `coll.indexes()` but keeping all information for each index found, not only stripping the name out of it
- fixed `expectExcepion` typo
- tests for both Obj-C and Swift for Array Index API
- contains [CBL-5899](https://jira.issues.couchbase.com/browse/CBL-589) as well

[CBL-5899]: https://couchbasecloud.atlassian.net/browse/CBL-5899?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Objective-C/Internal/CBLCollection+Internal.h
```diff
@@ -33,7 +35,7 @@
 
 NS_ASSUME_NONNULL_BEGIN
 
-@interface CBLCollection () <CBLRemovableListenerToken>
+@interface CBLCollection () <CBLRemovableListenerToken, CBLIndexableInternal>
```
pasin, line 38: I don't see a reason to create a protocol for CBLIndexableInternal as CBLCollection will not be used as CBLIndexableInternal type. I think you can just add a method `indexesInfo` in CBLCollection() directly.

Objective-C/Internal/CBLIndexable+Internal.h
```diff
@@ -0,0 +1,31 @@
...
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+//
+
+#import "CouchbaseLite/CBLIndexable.h"
+
+NS_ASSUME_NONNULL_BEGIN
+
+@protocol CBLIndexableInternal <CBLIndexable>
```
pasin, line 24: Don't need to create the protocol for this as the method can be put directly to CBLCollection Internal. 

Also no need to have underscored prefix as this function will not be a public API.

Objective-C/CBLCollection.mm
```diff
@@ -976,4 +831,171 @@ - (nullable CBLQueryIndex*) indexWithName: (nonnull NSString*)name
...
+            return nil;
+        
+        C4Error err = {};
+        C4SliceResult res = c4coll_getIndexesInfo(_c4col, &err);
+        if (err.code != 0){
+            convertError(err, error);
+            return nil;
+        }
+        
+        FLDoc doc = FLDoc_FromResultData(res, kFLTrusted, nullptr, nullslice);
```
pasin, line 991: Wondering if FLDoc requires the `res` to be a dictionary only or not (I don't have the info).

pasin (changes requested):

pasin (approved):
One comment and the rest is looking good.

PR #3331, opened 2024-10-07
CBL-6191 : Fix null URL for proxy CONNECT request (#3324)

velicuvlad:
NSURL cannot be created with only host and port anymore starting from iOS 17. Changed to use CFURLCreateWithString directly.

### DIFF
--- a/Objective-C/Internal/Replicator/CBLHTTPLogic.m
+++ b/Objective-C/Internal/Replicator/CBLHTTPLogic.m
@@ -144,7 +144,7 @@ static NSDictionary* sOverrideProxySettings;
     CFHTTPMessageRef httpMsg;
     if (_proxyType == kCBLHTTPProxy && _useProxyCONNECT) {
         NSString *destination = $sprintf(@"%@:%d", url.host, url.my_effectivePort);
-        CFURLRef requestURL = CFURLCreateWithString(kCFAllocatorDefault, (__bridge CFStringRef)destination, false);
+        CFURLRef requestURL = CFURLCreateWithString(kCFAllocatorDefault, (__bridge CFStringRef)destination, NULL);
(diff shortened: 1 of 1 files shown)


Objective-C/Internal/Replicator/CBLHTTPLogic.m
```diff
@@ -143,11 +143,13 @@ - (CFHTTPMessageRef) newHTTPRequest {
     // Create the CFHTTPMessage:
     CFHTTPMessageRef httpMsg;
     if (_proxyType == kCBLHTTPProxy && _useProxyCONNECT) {
-        NSURL *requestURL = [NSURL URLWithString: $sprintf(@"%@:%d", url.host, url.my_effectivePort)];
+        NSString *destination = $sprintf(@"%@:%d", url.host, url.my_effectivePort);
+        CFURLRef requestURL = CFURLCreateWithString(kCFAllocatorDefault, (__bridge CFStringRef)destination, false);
```
pasin, line 147: I just noticed that the last parameter is false. It should be NULL. I guess objective is kind of converting false to NULL but it should be fixed here and in 3.2 branch as well.

pasin (changes requested):

bmeike (approved):
I have no idea what is going on here. :-)
 I'm going to assume that it is necessary and correct.

pasin (approved):

pasin:
@bmeike 

From iOS 17, `NSURL's URLWithString:` function doesn't allow to create an URL without scheme anymore but we do need this for requesting a proxy's CONNECT request. So we are using a low level CFURLCreateWithString to create CFRULRef directly instead.

bmeike:
Wow.  Now I get it!

PR #3340, opened 2024-11-04
CBL-5736: Implement copy for MutableDocument and test

velicuvlad:
- Obj-C `CBLMutableDocument` now implements `(CBLMutableDocument*) copy`

- Swift `MutableDocument` `toMutable()` marked as deprecated and `copy()` replacement


Objective-C/CBLMutableDocument.h
```diff
@@ -110,6 +110,12 @@ NS_ASSUME_NONNULL_BEGIN
                                 json: (NSString*)json
                                error: (NSError**)error;
 
+/**
+ Returns a mutable copy of the document.
+ 
+ @return The CBLMutableDocument object.
+ */
+- (CBLMutableDocument*) copy;
```
pasin, line 118: I might be wrong as it has been awhile. I don't remember if we need to explicity define this as the copy is a part of NSObject protocol.

Objective-C/CBLMutableDocument.mm
```diff
@@ -119,6 +119,10 @@ - (CBLMutableDocument*) mutableCopyWithZone: (NSZone *)zone {
     return [[CBLMutableDocument alloc] initAsCopyWithDocument: self dict: _dict];
 }
 
+- (CBLMutableDocument*) copy {
+    return [self mutableCopyWithZone: nil];
```
pasin, line 123: I don't remember if the default copy calls mutableCopyWithZone or not. Can you double check?

Swift/MutableDocument.swift
```diff
@@ -97,10 +97,18 @@ public final class MutableDocument : Document, MutableDictionaryProtocol {
     /// Returns the same MutableDocument object.
     ///
     /// - Returns: The MutableDocument object.
+    @available(*, deprecated, message: "Use mutableDocument.copy() instead.")
```
pasin, line 100: We shouldn't deprecate this as toMutable() is our cross-platform API.

pasin (changes requested):

velicuvlad:
Will close this PR as is not relevant

PR #3344, opened 2024-11-13
3.2.1 SPM

velicuvlad:


velicuvlad:
Jenkins - `testCBLVersionAndLiteCoreVersion` failing due to the fact that underlying LiteCore XCode project remained under 3.2.0 (known issue).

bmeike (approved):
Again: should this change have gone into 3.2.1
Should this now be 3.2.2?

pasin (approved):

velicuvlad:
@bmeike the only thing that could've been was the project version change, but that is taken automatically from the build args when we do binaries, but the Jenkins would've fail for every PR because of the mismatch. This was caught by Jim last week, after Core RC. 

The rest is SPM specific, as we hold the CE here next to the source. This is a thing that will be changed for 4.0, will move under the separate repo https://github.com/couchbase/couchbase-lite-swift-ee , next to the EE, so this additional commit wont be needed finally.

PR #3346, opened 2024-11-20
Re-enable SGW tests on 3.2 line

velicuvlad:


velicuvlad:
Jenkins failing because of 1 test... which is the version check for Litecore. Litecore xcode project version has not been updated

pasin (approved):

PR #3349, opened 2024-12-17
Objc-c new logging api implementation

pasin:
* Implemented the logic to init CBLLogSinks.
* Implemented the logic to setup each log sink.
* Implemented a minium locking with a note comment in the CBLLogSinks implementation.
* Changed the old logging API to use the new log sink API.
* No tests for the new logging API implemented.
* No guard for using both old and new API implemented.
* PR validation can't be run when having a PR to non release branch or main branch for some reason.


velicuvlad:
yes, PR validation is only set for master/release. It needs modification into the yaml to run for every branch - there is no way to enable it on-fly.

I've tested locally and everything is fine.

I looked around today on how to handle the situation of old/new api use. Apple doesn't provide any "intelligent" way to do this and as far as I've skim around for ways to do such thing, I will move forward with a similar solution of what you've done for C. 

I will merge this now as I already started making changes on local on top. I do have a question take will take offline - it is concept related.

velicuvlad (approved):

PR #3351, opened 2024-12-20
New Logging API Obj-C

velicuvlad:
- [CBL-6455](https://jira.issues.couchbase.com/browse/CBL-6455)
- old Logging API uses the new LogSink API, thanks @pasin 
- do NOT allow the use new and old API at the same time.
- [CBL-6460](https://jira.issues.couchbase.com/browse/CBL-6455) changed/adapted existing tests to use the new API.
- renamed old tests under `LogTestOld`, which will be removed altogether with the old api, when time comes
- mark the old api as deprecated


velicuvlad:
Jenkins still failing on `[MiscTest testCBLVersionAndLiteCoreVersion]` ... again, is just a sanity check - will be fixed once we get a 3.2 LiteCore including https://github.com/couchbase/couchbase-lite-core/pull/2198

Objective-C/LogSink/CBLConsoleLogSink.mm
```diff
@@ -0,0 +1,76 @@
...
+- (void) writeLogWithLevel: (CBLLogLevel)level domain: (CBLLogDomain)domain message: (NSString*)message { 
+    if (level < self.level || (self.domain & domain) == 0) {
+        return;
+    }
+    
+    NSString* levelName = logLevelNames[level];
+    NSString* domainName = [self domainName: domain];
+    NSLog(@"CouchbaseLite %@ %@: %@", domainName, levelName, message);
+}
+
```
bmeike, line 51: Isn't this an attribute of the LiteCore interface?  Should it be somewhere generic for use by all of the loggers?

Objective-C/LogSink/CBLFileLogSink.mm
```diff
@@ -0,0 +1,103 @@
...
+    if (logSink) {
+        if (![self setupLogDirectory: logSink.directory error: &error]) {
+            CBLWarnError(Database, @"Cannot setup log directory at %@: %@", logSink.directory, error);
+            return;
+        }
+        
+        options = {
+            .base_path = CBLStringBytes(logSink.directory),
+            .log_level = (C4LogLevel)logSink.level,
+            .max_rotate_count = static_cast<int32_t>(logSink.maxKeptFiles - 1),
```
bmeike, line 71: Nice!  I missed this...

Objective-C/LogSink/CBLLogSinks.mm
```diff
@@ -0,0 +1,211 @@
...
+    CBLConsoleLogSink* console = self.console;
+    CBLLogLevel consoleLevel = console != nil ? console.level : kCBLLogLevelNone;
+    
+    CBLCustomLogSink* custom = self.custom;
+    CBLLogLevel customLevel = custom.logSink != nil ? custom.level : kCBLLogLevelNone;
+    
+    CBLFileLogSink* file = self.file;
+    CBLLogLevel fileLevel = file != nil ? file.level : kCBLLogLevelNone;
+    
+    CBLLogLevel callbackLevel = std::min(consoleLevel, customLevel);
```
bmeike, line 132: At first I thought that the two vars were redundant... but this is really nice and readable.

Objective-C/LogSink/CBLLogTypes.h
```diff
@@ -0,0 +1,54 @@
...
+//  limitations under the License.
+//
+
+NS_ASSUME_NONNULL_BEGIN
+
+/**
+ Log domain.
+ */
+typedef NS_OPTIONS(NSUInteger, CBLLogDomain) {
+    kCBLLogDomainAll            = 1 << 0 | 1 << 1 | 1 << 2 | 1 << 3 | 1 << 4, ///< All domains
```
bmeike, line 26: Would this be better as
kCBLLogDomainDatabase | kCBLLogDomainQuery | ...

bmeike (approved):
Looks pretty good.  Not a close read but the parts I understand seem fine.

Objective-C/LogSink/CBLLogSinks.mm
```diff
@@ -0,0 +1,211 @@
...
+    [CBLLogSinks.console writeLogWithLevel: level domain: domain message :message];
+    [CBLLogSinks.custom writeLogWithLevel: level domain: domain message :message];
+}
+
++ (void) writeCBLLog: (C4LogDomain)c4domain level: (C4LogLevel)c4level message: (NSString*)message {
+    CBLLogLevel level = (CBLLogLevel) c4level;
+    CBLLogDomain domain = toCBLLogDomain(c4domain);
+    
+    CBLStringBytes c4msg(message);
+    c4slog(c4domain, c4level, c4msg);
```
pasin, line 170: When looking at this, I think CBLFileLogSink should implement CBLLogSinkProtocol protocol so this can be moved to CBLFileLogSink's writeLogWithLevel() the same way as the console and custom log sink below.

Objective-C/LogSink/CBLLogSinks.mm
```diff
@@ -0,0 +1,211 @@
...
+        };
+    }
+    
+    NSString* domainName = [NSString stringWithUTF8String: c4log_getDomainName(domain)];
+    NSNumber* mapped = [domainDictionary objectForKey: domainName];
+    return mapped ? mapped.integerValue : kCBLLogDomainDatabase;
+}
+
++ (void)checkLogApiVersion: (LogAPI)version {
+    if (!isInitialized)
```
pasin, line 201: Not sure if it's required to be checked? Is there a case that the initialize() is not called and what will happen if not checking at all?

Objective-C/LogSink/CBLLogTypes.h
pasin, line 26: Maybe 0xFF?

velicuvlad: I actually did that for Swift as the namings are way shorter. As you now mentioned it now, I will change that for readability for Obj-C as well. Thanks!

Objective-C/Tests/VectorSearchTest.m
```diff
@@ -76,12 +76,11 @@ - (void) setUp {
     AssertNotNil(_wordDB);
     
     _logger = [[CustomLogger alloc] init];
-    _logger.level = kCBLLogLevelInfo;
-    CBLDatabase.log.custom = _logger;
+    CBLLogSinks.custom = [[CBLCustomLogSink alloc] initWithLevel: kCBLLogLevelInfo logSink: _logger];
```
pasin, line 79: I don't remember why we have a custom logger set for the VectorSearchTest?

Objective-C/Tests/LogTest.m
```diff
@@ -57,33 +49,24 @@ - (void) tearDown {
...
-
 - (void) backupLoggerConfig {
-    _backup = [[FileLoggerBackup alloc] init];
-    _backup.level = CBLDatabase.log.file.level;
-    _backup.config = CBLDatabase.log.file.config;
-    _backupConsoleLevel = CBLDatabase.log.console.level;
-    _backupConsoleDomain = CBLDatabase.log.console.domains;
+    _backup = [[CBLFileLogSink alloc] initWithLevel: CBLLogSinks.file.level directory: logFileDirectory];
+    _backupConsoleLevel = CBLLogSinks.console.level;
+    _backupConsoleDomain = CBLLogSinks.console.domain;
```
pasin, line 55: The backup could just be some thing like this:
```
_file = CBLLogSinks.file;
_console = CBLLogSinks.console;
```

I think after all the only backup is the console log sink for the case that we have enabled the console log sink for debugging purpose.

Objective-C/Tests/LogTestOld.m
```diff
@@ -0,0 +1,450 @@
...
+@end
+
+@interface LogTestOld : CBLTestCase
+
+@end
+
+@implementation LogTestOld {
+    FileLoggerBackup* _backup;
+    CBLLogLevel _backupConsoleLevel;
+    CBLLogDomain _backupConsoleDomain;
```
pasin, line 39: I think, there is no need to backup anything for testing the old API. The backup is for the case that we enable a console log in the base test case for debugging purpose.

pasin: How do you handle the new and the old API are both used as the new logging tests and the old logging tests will be both run? Doesn't an exception will be thrown?

pasin (changes requested):

Objective-C/LogSink/CBLConsoleLogSink.mm
velicuvlad: Not sure what you mean here? NSLog is an Obj-C native method to write to console only.

Objective-C/LogSink/CBLLogTypes.h
velicuvlad: I will go with Blake's recommendation of using `kCBL...`. I don't really want to go for 0xFF as we only have 5 bits not 8. The first thought I have for using 0xFF is that if LiteCore adds a new domain, `All` will include the new one as well and it can go unnoticed. As if we get a new domain from LiteCore we most probably going to need a singular entry as well for the new domain.

Objective-C/Tests/VectorSearchTest.m
velicuvlad: It is used to check for `Untrained index; queries may be slow`. I remember a conversation about this vaguely, but I wasn't sure enough to remove such checks. So I kept the custom logger in place. Do we still want to check for untrained? If not, I will remove this logger altogether.

pasin: I see. Let's keep it for now I think.

Objective-C/Tests/LogTestOld.m
velicuvlad: > How do you handle the new and the old API are both used as the new logging tests and the old logging tests will be both run? Doesn't an exception will be thrown?

Both suites run one after the other. I've used `isInitialized` to allow this only once, for the old custom logger needed in the old tests. Without that check, it will throw indeed

Objective-C/LogSink/CBLLogSinks.mm
velicuvlad: needed to bypass 1 time (every test instance, at init) in order to be able to run both old and new tests. Should I just disable old tests (and remove this workaround)?

pasin: I don't understand how this works. I don't see any code reset isInitialized. I think reset _vAPI to none would be enough for testing both old and new API?

pasin: After reading the code, I have noticed that the isInitialized is reset to NO in CBLLogSinks's initialize. I couldn't find where that method is called in this PR. Also I think a better way is to have an internal method that the test can used to reset the API version.

velicuvlad:
ready for review now

... `testCBLVersionAndLiteCoreVersion` failing only

Objective-C/CBLConsoleLogger.m
```diff
@@ -19,33 +19,68 @@
...
     }
     return self;
 }
 
 - (void) setLevel: (CBLLogLevel)level {
-    _level = level;
-    [[CBLLog sharedInstance] synchronizeCallbackLogLevel];
+    LogAPI version;
+    CBL_LOCK(self) {
+        version = [CBLLogSinks vAPI];
```
pasin, line 40: Why do we need to get the current API and set it at the end? 

I think what it needs to be done is something like this :

```
CBL_LOCK(self) {
  _level = level;
  CBLConsoleLogSink* consoleSink = [[CBLConsoleLogSink alloc] initWithLevel: _level domain: _domains];
  consoleSink.api = kLogAPIOld; // alternative, you can add an internal constructor to include the API as well
  CBLLogSinks.console = consoleSink;
}
```
Then in `CBLLogSinks.console` you can check whether the old or the new API has been used.

pasin (changes requested):

Objective-C/LogSink/CBLLogTypes.h
pasin: OK.

Objective-C/CBLLog.mm
```diff
@@ -149,69 +50,40 @@ static void sendToCallbackLogger(C4LogDomain d, C4LogLevel l, NSString* message)
...
-                C4LogLevel level = string2level(defaults[key]);
-                c4log_setLevel(domain, level);
-                NSLog(@"CouchbaseLite logging to %s domain at level %s", domainName, kLevelNames[level]);
-            }
-        }
-#endif
-        
-        // Keep the current callback log level:
-        _callbackLogLevel = (CBLLogLevel)callbackLogLevel;
+        [CBLLogSinks setVAPI: LogAPINone];
```
velicuvlad, line 56: here

Objective-C/CBLLog.mm
```diff
@@ -149,69 +50,40 @@ static void sendToCallbackLogger(C4LogDomain d, C4LogLevel l, NSString* message)
...
         
         // Create console logger:
-        _console = [[CBLConsoleLogger alloc] initWithLogLevel: _callbackLogLevel];
+        _console = [[CBLConsoleLogger alloc] initWithDefault];
         
-        // Create file logger which will enable file logging immediately with default log rotation:
+        // Create file logger:
         _file = [[CBLFileLogger alloc] initWithDefault];
+        
+        [CBLLogSinks setVAPI: LogAPINone];
```
velicuvlad, line 64: here

Objective-C/LogSink/CBLLogSinks.mm
```diff
@@ -0,0 +1,202 @@
...
+//
+// This class is implemented with a minimum logging. The interface defines
+// console, custom, and file property as nonatomic, so the expectation
+// is not to set up the log sinks object concurrently. However, at least,
+// the lock to access console and custom is required to avoid crash from accessing
+// garbage objects as LiteCore's log callback may be called from a background thread.
+
+@implementation CBLLogSinks
+
+static LogAPI _vAPI = LogAPINone;
```
velicuvlad, line 49: this doesn't have to me set. will change...

Objective-C/Tests/CBLTestCase.m
```diff
@@ -90,6 +91,10 @@ - (void) tearDown {
     [super tearDown];
 }
 
++ (void) tearDown {
+    [CBLLogSinks setVAPI: LogAPINone];
```
velicuvlad, line 95: here to reset after each test class run. So it doesn't crash when running in pipeline.

Objective-C/CBLConsoleLogger.m
pasin: For the nil case, the checkLogApiVersion: can just be no-ops, I think. It's not 100% perfect but it will serve the purpose enough that we don't want users to use two API to setup the log at the same time. Otherwise, you will need to add another layer for the implementation to allow the perfect check.

```
+ (void) checkLogApiVersion: (id<CBLLogApiSource>)source {
    if (!source) {
      return;
    }
    ...
}
```

pasin: > Those LogApi version grabs whatever the current version is CBLLogSinks under. And it matters if is None or Old. And just sets it back. So in the case Old is being called and Old has been used previously, we want to keep that (guard against a new, new api, call). And CBLLogSinks control None value as well, so it will only happen when set to LogApiNone in the internal implementation. It will never go to None from a user API call perspective.

I don't understand this point. The None is the status before one of the API is used. When the old API is used, the `+ (void) checkLogApiVersion: (id<CBLLogApiSource>)source` will keep the value as OLD so that the old API can be continue to be used.

Objective-C/CBLLog.mm
pasin: It is better do be done in `[CBLLogSinks init]` so the logic is controlled in a single place and done only once.

Objective-C/CBLLog.mm
pasin: It is better do be done in [CBLLogSinks init] so the logic is controlled in a single place and done **only once.**

pasin (changes requested):

Objective-C/Tests/CBLTestCase.m
pasin: In the change I suggested, it will call `+ (void) resetApiVersion` and if there is a crash, let look at that and fix it.

Objective-C/CBLConsoleLogger.m
velicuvlad: Another reason why I don't really want, although I tried, is that each individual sink doesn't even have to know about the api version. From the implementation view, only CBLLogSinks is the manager. It can change its internal variable depending of what get/set calls does for each sink. 

Indeed, for a "perfect" implementation it should also check if the call above is received from CBLLog/db.log... but I couldn't manage that. And so, I bypass each call of `updateSinks` inside CBLLog, ConsoleLogger, FileLogger to allow for the call (set) and revert back to whatever value was received. And it will only happen when flag is None which is only right after init or Old, which means another old api call was made beforehand. 

If the flag received is New, it means there was 1 call of the new API, after init. And so this then will throw due to the check on the old api.

i.e `ConsoleLogger`

```
- (void) setDomains: (CBLLogDomain)domains {
    LogAPI version;
    CBL_LOCK(self) {
        version = [CBLLogSinks vAPI];
        [CBLLogSinks checkLogApiVersion: LogAPIOld];
        _domains = domains;
    }
    [self updateConsoleLogSink];
    [CBLLogSinks setVAPI: version];
}

- (CBLLogDomain) domains {
    CBL_LOCK(self) {
        [CBLLogSinks checkLogApiVersion: LogAPIOld];
        return _domains;
    }
}
```

Objective-C/Tests/CBLTestCase.m
velicuvlad: I received a crash when setting to nil on test tearDown, as expected. I will gatekeep against and check

bmeike (approved):
Yeah... address Pasin's comments and it will be fine.  My apologies for handing you such a PITA.

Objective-C/CBLConsoleLogger.m
pasin: From my understanding, this will not work because when setting log sink at the new API, it will set the API to new (Actually, it works now because the code to mark that the new API is used is MISSING. Once adding that, you can see that this logic doesn't work). Also can you clarify why the version needs to be gotten first and set it back.

Objective-C/LogSink/CBLLogSinks.mm
```diff
@@ -0,0 +1,202 @@
...
+        kCBL_LogDomainListener  = c4log_getDomain("Listener", true);
+        
+        // Create the default warning console log:
+        self.console = [[CBLConsoleLogSink alloc] initWithLevel: kCBLLogLevelWarning];
+    });
+}
+
++ (void) setConsole:(CBLConsoleLogSink*)console {
+    CBL_LOCK(self) {
+        _console = console;
```
pasin, line 69: Need to mark that the new API is used?

pasin (changes requested):

Objective-C/LogSink/CBLLogSinks.mm
velicuvlad: it is done in each individual init for Console/File/Custom LogSinks. It is easier there than on the set/get because there are some additional places where checks are done and I should reset the internal vAPI again. Such place is:

```
+ (void) checkFileLogging {
    if (!CBLLogSinks.file) {
        CBLWarn(Database, @"File logging is disabled. Log files required for product support are not being generated.");
    }
}
```

velicuvlad: It seems reasonable to not be allowed to even allocate mem for the a new api object that you can't use, as you are already using the old api at that time. It will still leave the option to set to nil, which I would argue is the only gatekeep we could/should leave for the user: "use any API to stop logging" for convenience.

pasin: I don't think that is necessary at all. Especially, the trade-off is having the code that verifies old and new API in multiple places with some logic that is hard to understand and not sure if that is actually working. Let's see this:

In CBLLog.mm

```
- (void) setCustom: (id<CBLLogger>)custom {
     LogAPI version;
     CBL_LOCK(self) {
         version = [CBLLogSinks vAPI];
         [CBLLogSinks checkLogApiVersion: LogAPIOld];
         _custom = custom;
     }
     [self updateCustomLogSink];
     [CBLLogSinks setVAPI: version];
 }
 
 - (void) updateCustomLogSink {
     [CBLLogSinks setVAPI: LogAPINew];
     CBLCustomLogSinkBridge* bridge = [[CBLCustomLogSinkBridge alloc] initWithLogger: _custom];
     CBLLogSinks.custom = [[CBLCustomLogSink alloc] initWithLevel: _custom.level logSink: bridge];
 }
```

1. The code starts with getting the current API version. 
2. Verify that this is an OLD API usage.
3. Set the API to thew NEW version so that CBLCustomLogSinkBridge can be created.
4. Set the custom log lock.
5. Set the current API version from step 1 back.

It's so hard to understand how it works just until now, and I don't think it's justified the tradeoff.

pasin: The main goal to validate the old and the new API is trying to send the message to the users that they shouldn't mix using the old and the new API. We shouldn't try to complicate things too much to cover all of the edge case of using the API.

Objective-C/CBLConsoleLogger.m
velicuvlad: The code that marks that NEW Api is used is when checking for NEW with `[CBLLogSinks checkLogApiVersion: LogAPINew];` in the `initWithLevel` of every NEW api sink. The first call of that will set it to NEW. the only way that doesn't throw is when vAPI was previously None or New. And it suffice for the new api.

1. On the OLD api, the current version needs to be gotten first because we need to know if it was Old or None (at the old api level set/get). When it is Old, we change it to new to allow mem allocation for a new sink (the one in `updateXLogSink` - see above), then we revert to Old (so we guard against future NEW api calls). If it is None, it basically means it is somewhere during init, hence we set it back to None.

2. As vAPI is dynamically assigned on `checkLogApiVersion` call to Old (set/get for old api) or New (and initWithLevel for new api log sinks) whenever that it is None. So in order to allow `updateXLogSink` calls so we need to capture it beforehand to check if that was just set from the call itself, meaning it was None before, or was it already Old. We are mainly interested into keeping it if it was Old, as that will further guard for NEW api call down the line. For New will, of course, throw right away.

Objective-C/CBLLog.mm
velicuvlad: I will remove this one, is not necessary due to keeping track of `version` in the old api implm.

Objective-C/LogSink/CBLLogSinks.mm
pasin: When looking the old implementation, the old API will not set null value to any log sink so the validation API can assume that if the log sink is new, the new API is used (I have also updated my comment above regarding this).

```
+ (void) checkLogApiVersion: (id<CBLLogApiSource>)source {
    // The old API will not set the source to nil so when the source is nil, 
    // assume that it's from the new API.
    CBLLogAPI veresion = source ? souce.version :  kCBLLogAPINew;
     if (_vAPI == kCBLLogAPINone) { 
        _vAPI = version;
    } else if (_vAPI != version) { 
        [NSException raise: NSInternalInconsistencyException format: @"Cannot use both new and old Logging API simultaneously."];
     } 
}
```

Objective-C/CBLConsoleLogger.m
velicuvlad:
```+ (void) checkLogApiVersion: (id<CBLLogApiSource>)source {
    // The old API will not set the source to nil so when the source is nil, 
    // assume that it's from the new API.
    CBLLogAPI veresion = source ? souce.version :  kCBLLogAPINew;
     if (_vAPI == kCBLLogAPINone) { 
        _vAPI = version;
    } else if (_vAPI != version) { 
        [NSException raise: NSInternalInconsistencyException format: @"Cannot use both new and old Logging API simultaneously."];
     } 
}
```

The old one does set nil, at least for the old file logger. if config is nil during `setLevel` or `setConfig` then CBLLogSinks.file should be nil.

velicuvlad:
```
[[CBLFileLogSink alloc] initWithLevel: kCBLLogLevelNone
                            directory: @""
                         usePlaintext: false
                         maxKeptFiles: 0
                          maxFileSize: 0
                      ];
```

Objective-C/LogSink/CBLLogSinks.mm
```diff
@@ -0,0 +1,214 @@
...
+
++ (CBLLogAPI) vAPI {
+    return _vAPI;
+}
+
++ (void) setVAPI: (CBLLogAPI)version {
+    CBL_LOCK(self) {
+        _vAPI = version;
+    }
+}
```
pasin, line 198: These two functions are not needed anymore?

Objective-C/Tests/LogTest.m
```diff
@@ -46,44 +36,23 @@ @implementation LogTest {
...
-        CBLDatabase.log.file.level = _backup.level;
-        CBLDatabase.log.file.config = _backup.config;
-        _backup = nil;
-    }
-    CBLDatabase.log.custom = nil;
-    CBLDatabase.log.console.level = _backupConsoleLevel;
-    CBLDatabase.log.console.domains = _backupConsoleDomain;
-    
+    CBLLogSinks.file = _backup;
+    CBLLogSinks.console = [[CBLConsoleLogSink alloc] initWithLevel: kCBLLogLevelWarning];
```
pasin, line 48: Should set to a back up. For example, we want to debug an issue by turning a console log at DEBUG level on in the base class's setup, this setting will just turn it back to WARNING.

Objective-C/Tests/LogTest.m
```diff
@@ -46,44 +36,23 @@ @implementation LogTest {
...
+    CBLLogSinks.custom = nil;
+    _backup = nil;
 }
 
 - (NSArray<NSURL*>*) getLogsInDirectory: (NSString*)directory
-                             properties: (nullable NSArray<NSURLResourceKey>*)keys
-                           onlyInfoLogs: (BOOL)onlyInfo {
+                                      properties: (nullable NSArray<NSURLResourceKey>*)keys
+                                    onlyInfoLogs: (BOOL)onlyInfo {
```
pasin, line 55: Indent is off.

Objective-C/Tests/LogTest.m
```diff
@@ -121,7 +90,8 @@ - (void) writeAllLogs: (NSString*)string {
     CBLWarnError(Database, @"%@", string);
 }
 
-- (BOOL) isKeywordPresentInAnyLog: (NSString*)keyword path: (NSString*)path {
+- (BOOL) isKeywordPresentInAnyLog: (NSString*)keyword path: (nullable NSString*)path {
+    if (!path) return NO;
```
pasin, line 94: What is the case that this function is called with path == nil? If there is a case, maybe it's more straightforward to just assert the path in that test as it doesn't seem like it's responsibility of this method to verify the path?

Objective-C/Tests/LogTest.m
```diff
@@ -154,23 +123,19 @@ - (void) testCustomLoggingLevels {
 }
 
 - (void) testFileLoggingLevels {
-    CBLLogFileConfiguration* config = [self logFileConfig];
-    config.usePlainText = YES;
-    CBLDatabase.log.file.config = config;
-    
     for (NSUInteger i = 5; i >= 1; i--) {
-        CBLDatabase.log.file.level = (CBLLogLevel)i;
+        CBLLogSinks.file = [[CBLFileLogSink alloc] initWithLevel: (CBLLogLevel)i directory: logFileDirectory usePlaintext: YES maxKeptFiles: kCBLDefaultFileLogSinkMaxKeptFiles maxFileSize: kCBLDefaultLogFileMaxSize];
```
pasin, line 127: This line is too long. Please break it down into multiple lines.

Objective-C/Tests/LogTest.m
```diff
@@ -227,18 +190,12 @@ - (void) testFileLoggingDefaultBinaryFormat {
...
 
 - (void) testFileLoggingUsePlainText {
-    CBLLogFileConfiguration* config = [self logFileConfig];
-    AssertEqual(config.usePlainText, kCBLDefaultLogFileUsePlaintext);
-    config.usePlainText = YES;
-    Assert(config.usePlainText);
-    CBLDatabase.log.file.config = config;
-    CBLDatabase.log.file.level = kCBLLogLevelInfo;
-    Assert(CBLDatabase.log.file.config.usePlainText);
+    CBLLogSinks.file = [[CBLFileLogSink alloc] initWithLevel: kCBLLogLevelInfo directory: logFileDirectory usePlaintext: YES maxKeptFiles: kCBLDefaultFileLogSinkMaxKeptFiles maxFileSize: kCBLDefaultLogFileMaxSize];
```
pasin, line 193: Same here, please break it down to multiple lines.

Objective-C/Tests/LogTest.m
```diff
@@ -301,58 +254,46 @@ - (void) testEnableAndDisableCustomLogging {
...
+    CBLLogSinks.file = nil;
     
     NSString* inputString = [[NSUUID UUID] UUIDString];
     [self writeAllLogs: inputString];
     
-    AssertFalse([self isKeywordPresentInAnyLog: inputString path: config.directory]);
+    AssertFalse([self isKeywordPresentInAnyLog: inputString path: CBLLogSinks.file.directory]);
     
-    CBLDatabase.log.file.level = kCBLLogLevelVerbose;
+    CBLLogSinks.file = [[CBLFileLogSink alloc] initWithLevel: kCBLLogLevelVerbose directory: logFileDirectory usePlaintext: YES maxKeptFiles: kCBLDefaultFileLogSinkMaxKeptFiles maxFileSize: kCBLDefaultLogFileMaxSize];
```
pasin, line 294: Break it down to multiple lines. Generally, please keep the line up to around 120 chars.

Objective-C/Tests/LogTest.m
```diff
@@ -367,13 +308,10 @@ - (void) testFileLoggingReEnableLogging {
 }
 
 - (void) testFileLoggingHeader {
-    CBLLogFileConfiguration* config = [self logFileConfig];
-    config.usePlainText = YES;
-    CBLDatabase.log.file.config = config;
-    CBLDatabase.log.file.level = kCBLLogLevelVerbose;
+    CBLLogSinks.file = [[CBLFileLogSink alloc] initWithLevel: kCBLLogLevelVerbose directory: logFileDirectory usePlaintext: YES maxKeptFiles: kCBLDefaultFileLogSinkMaxKeptFiles maxFileSize: kCBLDefaultLogFileMaxSize];
```
pasin, line 311: Break it down to multiple lines. Generally, please keep the line up to around 120 chars.

Objective-C/Tests/CBLTestCase.m
```diff
@@ -90,6 +91,10 @@ - (void) tearDown {
     [super tearDown];
 }
 
++ (void) tearDown {
+    [CBLLogSinks resetApiVersion];
```
pasin, line 95: This may not work when we want to debug an issue by enable console log in the setup method. Probably move to the regular setup method in this class?

pasin (approved):
I have some format and test related comments. Please take a look. The rest look fine. Once the swift API is done, let come back and get another look for the number types that you have raised before.

Objective-C/Tests/CBLTestCase.m
velicuvlad: That is true if you enable console logging at the CBLTestCase level setup method.

I moved it under -setUp and I have added one in -tearDown as well. It is needed in case there's additional logging configuration made in test classes. For now is only in the log tests classes, but it might be the case in the future. 

Tested with enabling verbose logging using new api, before reset and it work. Marked with a comment.

Objective-C/Tests/LogTest.m
velicuvlad: I will remove this, alongside test `testFileLoggingDisabled` because it is redundant. The only check we can do is that .directory is nil when the .file is nil... which is always true. We already set it to nil in the next test `testFileLoggingReEnableLogging` so I will move the assert for .directory there.

velicuvlad: I will keep it for the old test because there we do set level None, but we actually create a file log

PR #3352, opened 2025-01-09
Update LiteCore submodule to 4.0.0-33

velicuvlad:


velicuvlad:
same deal again... This build will suffice for testing, but I would like to ask for a new LiteCore containing https://github.com/couchbase/couchbase-lite-core/pull/2203 (which is not a breaking change) as it is shown in the first line of each log file... and should match 4.0 both platform and LiteCore.

callumbirks (approved):

PR #3357, opened 2025-01-29
CBL-5193 : Partial Index API

velicuvlad:
- CBL-5193
- CBL-5194

For Swift, `where` is a reserved keyword. So I had to escape it via backticks. There is no impact for the user.

- LiteCore 3.2.2-8
- update testVectorMatchLimitBoundary to expect exception only on more than 10k

### DIFF
--- a/Objective-C/CBLFullTextIndexConfiguration.h
+++ b/Objective-C/CBLFullTextIndexConfiguration.h
@@ -2,7 +2,7 @@
 //  CBLFullTextIndexConfiguration.h
 //  CouchbaseLite
 //
-//  Copyright (c) 2024 Couchbase, Inc All rights reserved.
+//  Copyright (c) 2025 Couchbase, Inc All rights reserved.
 //
 //  Licensed under the Apache License, Version 2.0 (the "License");
 //  you may not use this file except in compliance with the License.
@@ -28,6 +28,12 @@ NS_ASSUME_NONNULL_BEGIN
 @interface CBLFullTextIndexConfiguration : CBLIndexConfiguration
 
 /**
+ A predicate expression defining conditions for indexing documents.
+ Only documents satisfying the predicate are included, enabling partial indexes.
+ */
+@property (nonatomic, readonly, nullable) NSString* where;
+
+/**
  Set the true value to ignore accents/diacritical marks. 
  */
 @property (nonatomic, readonly) BOOL ignoreAccents;
@@ -41,11 +47,28 @@ NS_ASSUME_NONNULL_BEGIN
 @property (nonatomic, readonly, nullable) NSString* language;
 
 /**
- Constructor for creating a full-text index by using an array of expression strings
+ Initializes a full-text index by using an array of expression strings
+ @param expressions The array of expression strings.
+ @param ignoreAccents The flag to ignore accents/diacritical marks.
+ @param language Optional language code which is an ISO-639 language such as "en", "fr", etc.
+ @return The full-text index configuration object.
  */
 - (instancetype) initWithExpression: (NSArray<NSString*>*)expressions
                       ignoreAccents: (BOOL)ignoreAccents
-                           language: (NSString* __nullable)language;
+                           language: (nullable NSString*)language;
+
+/**
+ Initializes a full-text index with an array of expression strings and an optional where clause for a partial index.
+ @param where Optional where clause for partial indexing.
+ @param expressions The array of expression strings.
+ @param ignoreAccents The flag to ignore accents/diacritical marks.
+ @param language Optional language code which is an ISO-639 language such as "en", "fr", etc.
+ @return The full-text index configuration object.
+ */
+- (instancetype) initWithExpression: (NSArray<NSString*>*)expressions
+                              where: (nullable NSString*)where
+                      ignoreAccents: (BOOL)ignoreAccents
+                           language: (nullable NSString*)language;
 
 @end
 
--- a/Objective-C/CBLFullTextIndexConfiguration.m
+++ b/Objective-C/CBLFullTextIndexConfiguration.m
@@ -2,7 +2,7 @@
 //  CBLFullTextIndexConfiguration.m
 //  CouchbaseLite
 //
-//  Copyright (c) 2021 Couchbase, Inc All rights reserved.
+//  Copyright (c) 2025 Couchbase, Inc All rights reserved.
 //
 //  Licensed under the Apache License, Version 2.0 (the "License");
 //  you may not use this file except in compliance with the License.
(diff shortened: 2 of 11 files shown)


Objective-C/CBLFullTextIndexConfiguration.h
```diff
@@ -40,13 +40,27 @@ NS_ASSUME_NONNULL_BEGIN
...
                       ignoreAccents: (BOOL)ignoreAccents
                            language: (NSString* __nullable)language;
 
+/**
+ Constructor for creating a partial full-text index by using an array of expression strings
+ */
+- (instancetype) initWithExpression: (NSArray<NSString*>*)expressions
+                      ignoreAccents: (BOOL)ignoreAccents
+                           language: (NSString* __nullable)language
+                              where: (nullable NSString*)where;
```
pasin, line 62: Seem like where and language are using a different way to make nullable. Can you check which method is the preferred or right one?

Objective-C/CBLFullTextIndexConfiguration.m
```diff
@@ -37,14 +37,28 @@ - (instancetype) initWithExpression: (NSArray<NSString*>*)expressions
...
+    self = [super initWithIndexType: kC4FullTextIndex
+                        expressions: expressions];
+    if (self) {
+        _where = where;
+        // there is no default 'ignoreAccents', since its NOT an optional argument.
+        _ignoreAccents = ignoreAccents;
+        _language = language;
+    }
+    return self;
+}
```
pasin, line 53: Can you make the initWithExpression:ignoreAccents:language constructor calls this constructor?

Objective-C/CBLValueIndexConfiguration.m
```diff
@@ -22,8 +22,26 @@
...
     return [super initWithIndexType: kC4ValueIndex expressions: expressions];
 }
 
+- (instancetype) initWithExpression: (NSArray<NSString*>*)expressions
+                              where: (nullable NSString*)where {
+    self = [super initWithIndexType: kC4ValueIndex expressions: expressions];
+    if (self) {
+        _where = where;
+    }
+
```
pasin, line 37: Extra line here.

pasin: Can you make initWithExpression: calls this constructor?

pasin (changes requested):

Swift/IndexConfiguration.swift
```diff
@@ -36,19 +36,24 @@ public struct FullTextIndexConfiguration: IndexConfiguration, IndexConfigConvert
...
-    public var language: String?
+    public let language: String?
+    
+    /// A predicate expression defining conditions for indexing documents.
+    /// Only documents satisfying the predicate are included, enabling partial indexes.
+    public let `where`: String?
     
     /// Constructor for creating a full-text index by using an array of N1QL expression strings
-    public init(_ expressions: [String], ignoreAccents: Bool? = FullTextIndexConfiguration.defaultIgnoreAccents, language: String? = nil) {
+    public init(_ expressions: [String], ignoreAccents: Bool? = FullTextIndexConfiguration.defaultIgnoreAccents, language: String? = nil, where: String? = nil) {
```
pasin, line 46: I'm not sure on this. If moving where next to expressions, will it break any users?

pasin (changes requested):

Objective-C/CBLValueIndexConfiguration.h
```diff
@@ -27,11 +27,23 @@ NS_ASSUME_NONNULL_BEGIN
...
+ */
+@property (nonatomic, readonly, nullable) NSString* where;
+
 /**
  Constructor for creating a value index by using an array of expression strings.
  */
 - (instancetype) initWithExpression: (NSArray<NSString*>*)expressions;
 
+/**
+ Constructor for creating a partialvalue index by using an array of expression strings.
```
pasin, line 42: partialvalue -> missing space.

Mentioning about where will make it clearer.
```
Initializes a partial value index with an array of expression strings and a where clause.

```

Swift/IndexConfiguration.swift
```diff
@@ -57,15 +62,21 @@ public struct ValueIndexConfiguration: IndexConfiguration, IndexConfigConvertabl
...
     public let expressions: [String]
     
+    
+    /// A predicate expression defining conditions for indexing documents.
+    /// Only documents satisfying the predicate are included, enabling partial indexes.
+    public let `where`: String?
+    
     /// Constructor for creating a value index by using an array of N1QL expression strings.
-    public init(_ expressions: [String]) {
+    public init(_ expressions: [String], where: String? = nil) {
```
pasin, line 71: I think mentioning about the where parameter as well is better.

```
/// Initializes a value index using an array of N1QL expression strings, with an optional where clause.
```

pasin (changes requested):

Swift/IndexConfiguration.swift
velicuvlad: I missed that only expressions can be NOT named, but the rest require it to. So no, it wouldnt make a difference. will move it

Objective-C/CBLFullTextIndexConfiguration.h
```diff
@@ -41,11 +41,25 @@ NS_ASSUME_NONNULL_BEGIN
...
- Constructor for creating a full-text index by using an array of expression strings
+ A predicate expression defining conditions for indexing documents.
+ Only documents satisfying the predicate are included, enabling partial indexes.
+ */
+@property (nonatomic, readonly, nullable) NSString* where;
+
+/**
+ Initializes a partial full-text index by using an array of expression strings
+ */
+- (instancetype) initWithExpression: (NSArray<NSString*>*)expressions
```
pasin, line 52: Sorry, I might have suggested to put the word partial there in the comment. It is not correct here as there is no where parameter.

Objective-C/CBLFullTextIndexConfiguration.h
```diff
@@ -41,11 +41,25 @@ NS_ASSUME_NONNULL_BEGIN
...
+                           language: (nullable NSString*)language;
+
+/**
+ Initializes a partial full-text index with an array of expression strings and a where clause.
  */
 - (instancetype) initWithExpression: (NSArray<NSString*>*)expressions
                       ignoreAccents: (BOOL)ignoreAccents
-                           language: (NSString* __nullable)language;
+                           language: (nullable NSString*)language
+                              where: (nullable NSString*)where;
```
pasin, line 62: Sorry, I didn't notice this in my first review. I feel like where should be next to the expression and that is based on what is in the API spec as well. Can you move where next to the expressions? 

For the comment, feel like mentioning the partial index next to where clause feels better as it's an optional. What do you think?

```
Initializes a full-text index with an array of expression strings and an optional where clause for a partial index.
```

bmeike (commented):
I'm not familiar enough with ObjC to approve this.  Pasin's suggestions look pretty good to me.

Objective-C/CBLValueIndexConfiguration.h
```diff
@@ -28,10 +28,22 @@ NS_ASSUME_NONNULL_BEGIN
...
+ Initializes a value index by using an array of expression strings.
  */
 - (instancetype) initWithExpression: (NSArray<NSString*>*)expressions;
 
+/**
+ Initializes a partial value index with an array of expression strings and a where clause.
+ */
+- (instancetype) initWithExpression: (NSArray<NSString*>*)expressions
+                              where: (nullable NSString*)where;
+
```
pasin, line 46: Sorry, I have suggested that comment. After reading it, it doesn't feel right as the where is optional. Probably something like this?

```
Initializes a value index with an array of expression strings and an optional where clause for a partial index.
```

pasin (changes requested):

Objective-C/CBLFullTextIndexConfiguration.h
velicuvlad: that was a typo, will fix

Objective-C/CBLValueIndexConfiguration.h
velicuvlad: Changed this to explicit optional. I have also now added API description for the constructors and its params

PR #3358, opened 2025-01-30
CBL-6456 : Implement Swift LogSink API

pasin:
* Implemented Swift LogSink API (CBL-6456) and tests (CBL-6461).

* Changed domain property to domains in CBLConsoleLogSink and CBLCustomLogSink to match API spec.

* Changed maxKepthFiles type to NSInteger and maxFileSize type to long long in CBLFileLogSink.

* Fixed missing a constructor with domains parameter in CBLCustomLogSink.

* Fixed missing public symbols for Objective-C LogSink API.

* Fixed using out-of-scope memory access from using CBLStringBytes when setting directory and header to C4 API in CBLFileLogSink.

* Fixed bug when setting null LogFileConfiguration from logger API.

* Updated and completed API docs for Objective-C LogSink API.


Objective-C/Internal/CBLStringBytes.h
```diff
@@ -24,22 +24,31 @@
...
-
-    NOTE: Since the slice may point directly into the NSString, if the string is mutable do not
-    mutate it while the stringBytes object is in scope! (Releasing the string is OK, as
-    stringBytes retains it.) */
+    stack based buffer if the inline storage is enabled, or into a temporary heap block.
+    NOTE:
+    - Since the slice may point directly into the NSString, if the string is mutable do not
+      mutate it while the stringBytes object is in scope! (Releasing the string is OK, as
+      stringBytes retains it.)
+    - Dissable the inline storage If the CBLStringBytes is used outside the scope that is created as
```
velicuvlad, line 32: Disable*

Swift/LogSinks.swift
```diff
@@ -0,0 +1,87 @@
...
+    
+    /// The file log sink, disabled by default.
+    public static var file: FileLogSink? = nil {
+        didSet {
+            if let file = self.file {
+                let level = CBLLogLevel(rawValue: UInt(file.level.rawValue))!
+                CBLLogSinks.file = CBLFileLogSink(level: level,
+                                                  directory: file.directory,
+                                                  usePlaintext: file.usePlaintext,
+                                                  maxKeptFiles: Int(file.maxKeptFiles),
```
velicuvlad, line 46: I think it should be Int64(file.maxKeptFiles)

Swift/Tests/CBLTestCase.swift
```diff
@@ -80,6 +80,8 @@ class CBLTestCase: XCTestCase {
     override func setUp() {
         super.setUp()
         
+        LogSinks._resetApiVersion()
```
velicuvlad, line 83: I would have this at the end of setUp, same as Obj-C, just because any logging after this might crash tests if the tests that are run are using a different API version vs any "debug" logging made in setUp. I believe it is the most fail-safe we can do.

Swift/LogSinks.swift
pasin: Int is correct as it's NSInteger not long long.

Swift/LogSinks.swift
```diff
@@ -0,0 +1,87 @@
...
+        
+        func writeLog(with level: CBLLogLevel, domain: CBLLogDomain, message: String) {
+            let logLevel = LogLevel.init(rawValue: UInt8(level.rawValue))!
+            let logDomain = LogDomain.init(rawValue: UInt8(domain.rawValue))!
+            logSink.writeLog(level:logLevel, domain: logDomain, message: message)
+        }
+    }
+    
+    /// <Unsupported API> Internally used for testing purpose.
+    public static func _resetApiVersion() {
```
velicuvlad, line 84: We should make this `internal static func ...` and then have `@testable import CouchbaseLiteSwift` in `CBLTestCase`.

Swift/Tests/CBLTestCase.swift
velicuvlad, line 1: `@testable import CouchbaseLiteSwift`

velicuvlad (changes requested):

velicuvlad (approved):

PR #3365, opened 2025-02-28
Collection.publish() Swift Enhancement APIs

velicuvlad:
- CBL-6709 Collection Change
- CBL-6710 Document Change

Will add more tests when I have the Test Spec ready. These are just confirm that the listener is actually added and that its token gets removed successfully on completion.

### DIFF
--- a/Swift/Collection.swift
+++ b/Swift/Collection.swift
@@ -18,6 +18,7 @@
 //
 
 import Foundation
+import Combine
 import CouchbaseLiteSwift_Private
 
 /// A `Collection` represent a collection which is a container for documents.
@@ -300,6 +301,42 @@ public final class Collection : CollectionChangeObservable, Indexable, Equatable
         }
         
         return ListenerToken(token)
+    }
+    
+    // MARK: Combine Publisher
+    
+    @available(iOS 13.0, *)
+    public func publish(on queue: DispatchQueue = .main) -> AnyPublisher<CollectionChange, Never> {
+        let subject = PassthroughSubject<CollectionChange, Never>()
+        var token: ListenerToken?
+        
+        token = self.addChangeListener(queue: queue) { change in
+            subject.send(change)
+        }
+
+        return subject
+            .receive(on: queue)
+            .handleEvents(receiveCompletion: { _ in
+                token?.remove()
+            })
+            .eraseToAnyPublisher()
+    }
+    
+    @available(iOS 13.0, *)
+    public func publish(for id: String, on queue: DispatchQueue = .main) -> AnyPublisher<DocumentChange, Never> {
+        let subject = PassthroughSubject<DocumentChange, Never>()
+        var token: ListenerToken?
+
+        token = self.addDocumentChangeListener(id: id, queue: queue) { change in
+            subject.send(change)
+        }
+
+        return subject
+            .receive(on: queue)
+            .handleEvents(receiveCompletion: { _ in
+                token?.remove()
+            })
+            .eraseToAnyPublisher()
     }
     
     // MARK: Indexable
--- a/Swift/Collection.swift
+++ b/Swift/Collection.swift
@@ -315,14 +315,7 @@ public final class Collection : CollectionChangeObservable, Indexable, Equatable
 
         return subject
             .receive(on: queue)
-            .handleEvents(
-                receiveCompletion: { _ in
-                    token.remove()
-                },
-                receiveCancel: {
-                    token.remove()
-                }
-            )
+            .handleEvents(receiveCancel: { token.remove() })
             .eraseToAnyPublisher()
     }
     
@@ -336,14 +329,7 @@ public final class Collection : CollectionChangeObservable, Indexable, Equatable
 
         return subject
             .receive(on: queue)
-            .handleEvents(
-                receiveCompletion: { _ in
-                    token.remove()
-                },
-                receiveCancel: {
-                    token.remove()
-                }
-            )
+            .handleEvents(receiveCancel: { token.remove() })
             .eraseToAnyPublisher()
     }
     
--- a/Swift/Collection.swift
+++ b/Swift/Collection.swift
@@ -308,34 +308,42 @@ public final class Collection : CollectionChangeObservable, Indexable, Equatable
     @available(iOS 13.0, *)
     public func changePublisher(on queue: DispatchQueue = .main) -> AnyPublisher<CollectionChange, Never> {
         let subject = PassthroughSubject<CollectionChange, Never>()
-        var token: ListenerToken?
         
-        token = self.addChangeListener(queue: queue) { change in
+        let token = self.addChangeListener(queue: queue) { change in
             subject.send(change)
         }
 
         return subject
             .receive(on: queue)
-            .handleEvents(receiveCompletion: { _ in
-                token?.remove()
-            })
+            .handleEvents(
+                receiveCompletion: { _ in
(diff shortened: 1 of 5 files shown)


velicuvlad:
Will hold until 3.2.2 is out for good.

callumbirks (approved):

bmeike (commented):
✅

Swift/Collection.swift
```diff
@@ -302,6 +303,42 @@ public final class Collection : CollectionChangeObservable, Indexable, Equatable
...
+        let subject = PassthroughSubject<CollectionChange, Never>()
+        var token: ListenerToken?
+        
+        token = self.addChangeListener(queue: queue) { change in
+            subject.send(change)
+        }
+
+        return subject
+            .receive(on: queue)
+            .handleEvents(receiveCompletion: { _ in
```
pasin, line 319: I'm not 100% on this but from the [doc](https://developer.apple.com/documentation/combine/passthroughsubject/handleevents(receivesubscription:receiveoutput:receivecompletion:receivecancel:receiverequest:), should it handles `receivedCancel` event to remove the token instead?

Swift/Collection.swift
```diff
@@ -302,6 +303,42 @@ public final class Collection : CollectionChangeObservable, Indexable, Equatable
...
     }
     
+    // MARK: Combine Publisher
+    
+    @available(iOS 13.0, *)
+    public func changePublisher(on queue: DispatchQueue = .main) -> AnyPublisher<CollectionChange, Never> {
+        let subject = PassthroughSubject<CollectionChange, Never>()
+        var token: ListenerToken?
+        
+        token = self.addChangeListener(queue: queue) { change in
```
pasin, line 313: There is no need to have the separate lines for defining the token variable.

```
let token = self.addChangeListener(queue: queue) { .... }
```

Swift/Collection.swift
pasin: Probably should have the test case to test this.

pasin (changes requested):

Swift/Collection.swift
velicuvlad: Both should be there if we want to handle both cases. I haven't even considered this... how would a listener for collection/document would be "canceled"? I couldn't think of an example that would refer to the existing implementation as canceled. But as we do allow for the token to be removed AT ANY TIME... I guess we need to handle all "such" scenarios - which includes a "cancel" event on the publisher.

Swift/Collection.swift
velicuvlad: you are right, `addChangeListener` doesn't return Optional.

pasin: No need to have a separate line here also for the optional case.

Swift/Collection.swift
pasin: Only receiveCancel is needed here because there is no upstream publisher for the PassthroughSubject created here to send the complete event. The receiveCancel event is what the downstream subscribers send whey they call cancel() on the returned Cancellable. 

As discussed in the meeting yesterday, the PassthroughSubject can send the complete event to the downstream event when the database is closed; it's not a mandatory. It requires a larger changes to the database to be able to send close events to the PassthroughSubject which I think we shouldn't do it in the patch release but in the minor or major release instead.

pasin (changes requested):

Swift/Tests/PublisherTest.swift
```diff
@@ -0,0 +1,86 @@
...
+
+import Foundation
+import Combine
+import XCTest
+@testable import CouchbaseLiteSwift
+
+@available(iOS 13.0, *)
+class PublisherTest: CBLTestCase {
+    
+    var cancellable: AnyCancellable?
```
pasin, line 27: From the current tests, it doesn't need to keep the cancellable object here.

Swift/Tests/PublisherTest.swift
```diff
@@ -0,0 +1,86 @@
...
+//  limitations under the License.
+//
+
+import Foundation
+import Combine
+import XCTest
+@testable import CouchbaseLiteSwift
+
+@available(iOS 13.0, *)
+class PublisherTest: CBLTestCase {
```
pasin, line 25: I think it might be easier to put separate tests into the CollectionTest, ReplicatorTest, and QueryTest class especially the publishers for the Replicator, otherwise this test class will need to bring several things from the ReplicatorTest. Also the tests can stay closer to the current add/remove change listener tests as they are in the same category.

pasin (changes requested):

Swift/Tests/PublisherTest.swift
```diff
@@ -40,13 +40,20 @@ class PublisherTest: CBLTestCase {
 
         wait(for: [expectation], timeout: 2.0)
     }
-            
-    func testCollectionChangePublisherRemovesListener() throws {
+    
+    func testCollectionChangePublisherRemoveListenerCompletion() throws {
```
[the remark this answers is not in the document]
velicuvlad: i will check on wether setting to `nil` does cancel or complete the event. But I don't think we need to test wether the change events are still received after calling cancel. If cancel fails to remove the listener then the test would fail as the token is not freed when closing the db at the end of test, right?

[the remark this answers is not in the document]
pasin: I don't think the test will fail if the token is not actually removed. You can comment out the code that remove the token and run the test to see the behavior.

Swift/Tests/PublisherTest.swift
```diff
@@ -0,0 +1,86 @@
...
+class PublisherTest: CBLTestCase {
+    
+    var cancellable: AnyCancellable?
+    
+    func testCollectionPublisherAddsListener() throws {
+        let expectation = self.expectation(description: "Got the change")
+        
+        cancellable = defaultCollection!.changePublisher()
+            .sink { change in
+                XCTAssertFalse(change.documentIDs.isEmpty, "THE change")
```
pasin, line 34: > "THE change".
Can you change the message to be more meaningful such as "documentIDs must not be empty"  if you want to include the message as well?

Also, I think it's a good idea to assert whether documentIDs contain "doc1" or not.

pasin (changes requested):

Swift/Tests/PublisherTest.swift
```diff
@@ -0,0 +1,76 @@
...
+                XCTAssert(change.documentIDs.first == "doc1")
+                expect.fulfill()
+            }
+            .store(in: &cancellables)
+        
+        // it also saves the document into the db
+        let doc = try generateDocument(withID: "doc1")
+        
+        try defaultCollection!.save(document: doc)
+        XCTAssertNotNil(cancellables)
```
pasin, line 54: cancellables is always not nil. Should check for the size == 1 instead.

Swift/Tests/PublisherTest.swift
```diff
@@ -0,0 +1,76 @@
...
+                XCTAssert(change.documentID == "doc1")
+                expect.fulfill()
+            }
+            .store(in: &cancellables)
+
+        // it also saves the document into the db
+        let doc = try generateDocument(withID: "doc1")
+        
+        try defaultCollection!.save(document: doc)
+        XCTAssertNotNil(cancellables)
```
pasin, line 73: Same as above

pasin (changes requested):

pasin (approved):

PR #3368, opened 2025-03-11
CBL-6711: Query.publish() Swift Enhancement API

velicuvlad:


Swift/Collection.swift
```diff
@@ -305,6 +305,12 @@ public final class Collection : CollectionChangeObservable, Indexable, Equatable
     
     // MARK: Combine Publisher
     
+    /// Returns a Combine publisher that emits `CollectionChange` events when documents
+    /// in the collection are changed.
+    ///
+    /// - Parameter queue: The `DispatchQueue` for event delivery. Defaults to the main queue.
+    /// - Returns: A `PassthroughSubject<CollectionChange, Never>` that emits collection changes.
+    /// - Note: Only available on  iOS 13.0 and later.
```
pasin, line 313: Is there an extra space between `on` and `iOS`?

Swift/Tests/PublisherTest.swift
```diff
@@ -22,7 +22,7 @@ import XCTest
 @testable import CouchbaseLiteSwift
 
 @available(iOS 13.0, *)
-class PublisherTest: CBLTestCase {
+class PublisherTest: QueryTest {
```
pasin, line 25: Does this also run all QueryTest tests as well? If that happens, alternatively, you could move `createDocNumbered` to the base test class.

Swift/Query.swift
```diff
@@ -124,6 +125,28 @@ public class Query {
...
+        let subject = PassthroughSubject<QueryChange, Never>()
+        
+        let token = self.addChangeListener(withQueue: queue) { change in
+            subject.send(change)
+        }
+
+        return subject
+            .receive(on: queue)
+            .handleEvents(receiveCancel: { token.remove() })
+            .eraseToAnyPublisher()
```
bmeike, line 147: just out of curiosity, what does "eraseToAnyPublisher" do?

bmeike (commented):
LGTM...

Swift/Tests/PublisherTest.swift
velicuvlad: no it doesnt, but good catch. That is a leftover from something I've tried previously. It is not needed - `createDocNumbered` is in CBLTestCase. I will change it back

Swift/Query.swift
velicuvlad: it hides the publisher type converting it to AnyPublisher for subscription. It is future proof in case we change any implementation.

PR #3369, opened 2025-03-14
CBL-6818: Allow save/purge document from the same collection but different collection instance

velicuvlad:


Objective-C/Tests/DocumentTest.m
```diff
@@ -1577,6 +1577,19 @@ - (void) testPurgeDocument {
...
+    NSError* error;
+    CBLCollection* defaultCollection1 = [self.db defaultCollection: &error];
+    CBLCollection* defaultCollection2 = [self.db defaultCollection: &error];
+    
+    CBLMutableDocument* doc1 = [self createDocument: @"doc1"];
+    [self saveDocument: doc1];
+    
+    // Get using 1, delete using 2
+    doc1 = [[defaultCollection1 documentWithID: doc1.id error:&error] toMutable];
+    [defaultCollection2 deleteDocument:doc1 error:&error];
```
pasin, line 1590: Look like iOS does cache the default collection object (which is fine) so that is the reason why this test is passed. Use non-default collection instead.

pasin (changes requested):

pasin (approved):
Can you also add the same test but with save?

Objective-C/Tests/DocumentTest.m
```diff
@@ -2262,6 +2262,62 @@ - (void) testDocumentResaveInAnotherCollection {
     }];
 }
 
+- (void) withOperation: (BOOL (^)(CBLCollection*, CBLMutableDocument*, NSError*)) operation {
```
pasin, line 2265: Method name is too generic and needs description. Maybe, `useDifferentCollectionInstance`.

Swift/Tests/DocumentTest.swift
```diff
@@ -1306,6 +1306,18 @@ class DocumentTest: CBLTestCase {
         try defaultCollection!.purge(id: doc2.id)
     }
     
+    func testDeleteDocumentWithDifferentCollectionInstace() throws {
```
pasin, line 1309: Is this duplicated with test below?

pasin (changes requested):

pasin (approved):

PR #3370, opened 2025-03-25
CBL-6716: Implement Swift Codable API

callumbirks:


Swift/Collection.swift
```diff
@@ -93,6 +93,30 @@ public final class Collection : CollectionChangeObservable, Indexable, Equatable
...
+    }
+    
+    public func document<T: DocumentDecodable>(id: String, as type: T.Type) throws -> T? {
+        guard let doc = try document(id: id)?.toMutable() else {
+            return nil
+        }
+        let decoder = DocumentDecoder(document: doc)
+        let object = try T(from: decoder)
+        if getDocumentRef(object: object) == nil {
+            throw CBLError.create(CBLError.invalidParameter, description: "Type \(String(describing: type)) is missing a @DocumentId field")
```
pasin, line 115: "@DocumentID"

Swift/Collection.swift
```diff
@@ -227,6 +364,19 @@ public final class Collection : CollectionChangeObservable, Indexable, Equatable
         try impl.purgeDocument(withID: id)
     }
     
+    /// Purge a document represented by the specified encodable model object from
+    /// the collection.
+    /// If the object is not linked to a document in the collection, the NotFound error will be thrown.
+    public func purgeDocument<T: DocumentEncodable>(for object: T) throws {
+        guard let docRef = getDocumentRef(object: object) else {
+            throw NSError(domain: CBLErrorDomain, code: CBLErrorInvalidParameter, userInfo: [NSLocalizedDescriptionKey : "Cannot purge document for object: No @DocumentId found on the object"])
```
pasin, line 372: @DocumentID

Swift/DocumentDecoder.swift
```diff
@@ -0,0 +1,121 @@
...
+        guard let nested = nestedValue as? ArrayObject else {
+            throw CBLError.create(CBLError.decodingError, description: "Type mismatch for key '\(key.stringValue)': expected array, found \(String(describing: nestedValue))")
+        }
+        let nestedDecoder = FleeceDecoder(fleeceValue: .array(nested))
+        return FleeceArrayDecodingContainer(decoder: nestedDecoder, array: nested)
+    }
+    
+
+    func superDecoder() throws -> any Decoder {
+        guard let nestedValue = decoder.document.value(forKey: "super") else {
```
pasin, line 103: Is key "super" the key we defined?

Swift/DocumentModel.swift
```diff
@@ -0,0 +1,69 @@
+//
+//  DocumentModel.swift
```
pasin, line 2: Probably this file is created when we still have the DocumentModel macro? Should the file be DocumentID.swift as the only class here is DocumentID?

Swift/DocumentModel.swift
```diff
@@ -0,0 +1,69 @@
...
+
+    public convenience init(from decoder: Decoder) throws {
+        if let dec = decoder as? DocumentDecoder {
+            // If decoder has a document, we should assign it to self
+            self.init(wrappedValue: dec.document.id)
+            self.revID = dec.document.revisionID
+        } else {
+            // Otherwise, try to decode the ID string.
+            let id = try String?.init(from: decoder)
+            self.init(wrappedValue: id)
```
pasin, line 52: What does this mean if the type is not string?

Swift/Tests/ArrayTest.swift
```diff
@@ -146,7 +146,7 @@ class ArrayTest: CBLTestCase {
         let data = arrayOfAllTypes()
         let array = MutableArrayObject()
         for _ in 0...11 {
-            array.addValue(NSNull())
+            array.addValue(nil)
```
pasin, line 149: Why changing this?

Objective-C/CBLSharedKeys.h
```diff
@@ -0,0 +1,31 @@
...
+//  Copyright © 2025 Couchbase. All rights reserved.
+//
+
+#import <Foundation/Foundation.h>
+
+@class CBLSharedKeys;
+
+NS_ASSUME_NONNUL_BEGIN
+
+@protocol CBLSharedKeys <NSObject, NSFastEnumeration>
```
pasin, line 15: What is the purpose of this protocol and class?

callumbirks:
@pasin I believe I have addressed all comments.

Objective-C/CBLDatabase.mm
```diff
@@ -357,7 +357,7 @@ - (BOOL) inBatch: (NSError**)outError usingBlock: (void (NS_NOESCAPE ^)())block
     return [self inBatch: outError usingBlockWithError: ^(NSError **) { block(); }];
 }
 
-- (BOOL) withTransaction: (NSError**)outError usingBlockWithError: (BOOL (NS_NOESCAPE ^)(NSError**))block NS_REFINED_FOR_SWIFT {
+- (BOOL) maybeBatch: (NSError**)outError usingBlockWithError: (BOOL (NS_NOESCAPE ^)(NSError**))block {
```
pasin, line 360: Not sure why `maybe`. The original `withTransaction` is fine too as it's an internal function or could be just `inBatchWithError:`?

Swift/Database.swift
```diff
@@ -591,4 +599,24 @@ extension CBLDatabase {
             }
         })
     }
+    
+    /// The same as `inBatch(() -> Void)`, but the closure can return a Bool.
+    /// If the closure returns `false`, the transaction will be aborted, and this function will return false.
+    internal func maybeBatch(block: () throws -> Bool) throws -> Bool {
+        var err: NSError? = nil
+        guard __maybeBatch(&err, usingBlockWithError: { (errPtr) -> Bool in
```
pasin, line 607: Where is this __maybeBatch defined?

Swift/Database.swift
```diff
@@ -591,4 +599,24 @@ extension CBLDatabase {
             }
         })
     }
+    
+    /// The same as `inBatch(() -> Void)`, but the closure can return a Bool.
+    /// If the closure returns `false`, the transaction will be aborted, and this function will return false.
+    internal func maybeBatch(block: () throws -> Bool) throws -> Bool {
```
pasin, line 605: Can this and `internal func maybeBatch(using block: () throws -> Bool ) throws -> Bool {` be combined?

pasin: From slack chat, I think I got it.

Swift/FleeceDecoder.swift
```diff
@@ -0,0 +1,531 @@
...
+    case int(Int)
+    case uint(UInt)
+    case float(Float)
+    case double(Double)
+    case string(String)
+    case blob(Blob)
+    case array(ArrayObject)
+    case dictionary(DictionaryObject)
+    
+    private init?(_ value: Any) {
```
pasin, line 457: I think I'm a bit disconnected here. Why FleeceValue only has Int and UInt type? If it means to target the largest value, should it be Int64 and UInt64?

Swift/Database.swift
callumbirks: This is because of the NS_REFINED_FOR_SWIFT on the `maybeBatch` function in CBLDatabase+Swift.h

Swift/FleeceDecoder.swift
callumbirks: Hmmm that's a good point

Objective-C/CBLDatabase.mm
callumbirks: I liked maybe because it won't batch (transaction is aborted) if the closure returns false at any point

pasin (approved):

PR #3372, opened 2025-04-15
Rename Collection APIs for Codable

velicuvlad:
- Updated tests
- LiteCore 3.2.3-8


pasin (approved):

pasin (approved):
`testCollectionChangePublisher` test failed. Can you check?

velicuvlad:
yes, it is only in Jenkins because of slowness. Seems like a 20 sec wait for expectation is sometimes too small... it passed on the 3rd run

PR #3393, opened 2025-05-30
UserAgent and some test fixing

velicuvlad:
- example of userAgent string `CouchbaseLite/3.2.x (Swift; macOS 15.3.1) Build/0 Commit/f7601574+CHANGES LiteCore/0.0.0 (6cc2f7f1e2a09767+64d62a79715ceeff)`
- moved publisher tests to expect only 1 notification/1 save as we can't guarantee the same amount. Expecting 2 changes on the same document was very flaky

### DIFF
--- a/Swift/Tests/PublisherTest.swift
+++ b/Swift/Tests/PublisherTest.swift
@@ -39,7 +39,7 @@ class PublisherTest: CBLTestCase {
     
     func testCollectionChangePublisher() throws {
         let expect = self.expectation(description: "Collection changed")
-        expect.expectedFulfillmentCount = 2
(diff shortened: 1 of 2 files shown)


Swift/Tests/PublisherTest.swift
```diff
@@ -39,7 +39,7 @@ class PublisherTest: CBLTestCase {
     
     func testCollectionChangePublisher() throws {
         let expect = self.expectation(description: "Collection changed")
-        expect.expectedFulfillmentCount = 2
+        expect.expectedFulfillmentCount = 1
```
pasin, line 42: It's one by default.

pasin (approved):

PR #3396, opened 2025-06-03
CBL-7061: Swift Codable decoding fails for some ISO8601 formats

callumbirks:


velicuvlad (approved):

bmeike (approved):
I think there should be a test that validates the `else if` clause.

Otherwise, this is all good.

PR #3413, opened 2025-07-07
CBL-7035: Improve Index API docs

velicuvlad:


Objective-C/CBLIndexable.h
```diff
@@ -28,21 +28,53 @@ NS_ASSUME_NONNULL_BEGIN
 /** The Indexable interface defines a set of functions for managing the query indexes. */
 @protocol CBLIndexable <NSObject>
 
-/** Return all index names, or nil if an error occurred. */
+/**
+ Get the names of all indexes of the collection.
+ 
+ @param error On return, the error if any.
+ @return Names of all indexes of the collection or nil if no indexes exist.
```
pasin, line 35: If no indexes exists, I think it will return an empty array (please double check), only when there is an error, it will return nil.

Objective-C/CBLIndexable.h
```diff
@@ -28,21 +28,53 @@ NS_ASSUME_NONNULL_BEGIN
...
+ */
 - (BOOL) deleteIndexWithName: (NSString*)name error: (NSError**)error;
 
-/** Get an index object by name. */
+/**
+ Get an index object by name.
+ 
+ @param name The index name.
+ @param error On return, the error if any.
+ @return The index object or nil.
```
pasin, line 76: The index object, or nil if the index does not exist or an error occurred.

Swift/Collection.swift
```diff
@@ -504,27 +504,48 @@ public final class Collection: CollectionChangeObservable, Indexable, Equatable,
...
         try impl.createIndex(withName: name, config: config.toImpl())
     }
     
-    /// Create an index with the index name and index instance.
+    /// Create an index with the index name and index instance
+    /// - Parameters:
+    ///   - index: Index instance.
+    ///   - name: The index name.
+    /// - Returns: True if the index is created successfully.
+    /// - Throws: If name or config are invalid, if the collection is invalid or the database is closed.
```
pasin, line 529: There is no config parameter here.

Swift/Collection.swift
```diff
@@ -504,27 +504,48 @@ public final class Collection: CollectionChangeObservable, Indexable, Equatable,
...
+    /// - Throws: If name is invalid, if the collection is invalid or the database is closed.
     public func deleteIndex(forName name: String) throws {
         try impl.deleteIndex(withName: name)
     }
     
     /// Get an index by name. Return nil if the index doesn't exists.
+    /// - Parameters:
+    ///   - name: The index name.
+    /// - Returns: The index object if found, otherwise nil.
+    /// - Throws: If name is invalid, if the collection is invalid or the database is closed.
```
pasin, line 548: I don't think we validate the name. Also from swift, name can't be nil.

Swift/Collection.swift
```diff
@@ -504,27 +504,48 @@ public final class Collection: CollectionChangeObservable, Indexable, Equatable,
...
+    /// - Note: To be used with Index Builder
     public func createIndex(_ index: Index, name: String) throws {
         try impl.createIndex(index.toImpl(), name: name)
     }
     
     /// Delete an index by name.
+    /// - Parameters:
+    ///   - name: The index name.
+    /// - Returns: True if the index is delete successfully.
+    /// - Throws: If name is invalid, if the collection is invalid or the database is closed.
```
pasin, line 539: I don't think we validate the name. Also from swift, name can't be nil. 

It could be good to check if it's throwing an error when the index doesn't exist or no-ops.

Maybe something generic is more safe such as

Throws: If the collection is invalid, the database is closed, or an error occurs.

PR #3434, opened 2025-08-15
CBL-7351: Return error for AnchorTrusted in CBLCheckTrust

pasin:
* MultipeerReplicatorTest.testAuthenticateWithRootCertsFailed failed on iOS devices but passed on macOS.

* The issue was that CBLTrustCheck didn’t return an error for AnchorTrusted problems — it didn’t return NO. On macOS the trust validation failed outright, but on iOS it was a recoverable error and required additional checks.

* Refactor CBLTrustCheck to skip host check only when the host is not specified.

* Add comments in CBLWebSocket and simplify the trust check calls.

* Update TrustCheckTest and clean up MultipeerReplicatorTest.

* Manual replication test performed to verify CBLTrustCheck and CBLWebSocket still work with standard wss:// servers.

### DIFF
--- a/Objective-C/Tests/MultipeerReplicatorTest.m
+++ b/Objective-C/Tests/MultipeerReplicatorTest.m
@@ -85,13 +85,13 @@ typedef void (^MultipeerCollectionConfigureBlock)(CBLMultipeerCollectionConfigur
 
 // TODO: Too flaky, enable after all LiteCore issues have been handled.
 - (BOOL) isExecutionAllowed {
-#if TARGET_OS_SIMULATOR
+#if TARGET_OS_SIMULATOR || TARGET_OS_MACCATALYST || TARGET_OS_OSX
     // Cannot be tested on the simulator as local network access permission is required.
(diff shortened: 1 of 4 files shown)


Objective-C/Tests/MultipeerReplicatorTest.m
```diff
@@ -85,13 +85,13 @@ - (void) tearDown {
 
 // TODO: Too flaky, enable after all LiteCore issues have been handled.
 - (BOOL) isExecutionAllowed {
-#if TARGET_OS_SIMULATOR
+#if TARGET_OS_SIMULATOR || TARGET_OS_MACCATALYST || TARGET_OS_OSX
```
velicuvlad, line 88: So this will only work for an iOS, iPadOS or tvOS devices. As we don't support iPadOS and tvOS, maybe we can have this check only as?
```
#if TARGET_OS_IPHONE 
return self.keyChainAccessAllowed
#else
return NO
```

PR #3444, opened 2025-09-04
Deprecate the old logging API in Swift

velicuvlad:


pasin:
This requires LiteCore update to get the fleece build issue fixed for XCode 16.

PR #3449, opened 2025-09-10
CBL-7313: Remove Deprecated Logging APIs

velicuvlad:


Objective-C/Internal/CBLLogSinks+Swift.h
```diff
@@ -17,9 +17,14 @@
...
 
 #import "CBLLogSinks.h"
+#import "CBLLogTypes.h"
+
+NS_ASSUME_NONNULL_BEGIN
 
 @interface CBLLogSinks ()
 
-+ (void) resetApiVersion;
++ (void) writeSwiftLog: (CBLLogDomain)domain level: (CBLLogLevel)level message: (NSString*)message;
```
pasin, line 26: This should be at the CBLLog not CBLLogSinks. CBLLogSinks is only for setting up the destinations where the log messages go, not for logging.

Objective-C/LogSink/CBLLogSinks.mm
```diff
@@ -114,6 +111,34 @@ + (CBLFileLogSink*) file {
...
+}
+
++ (instancetype) sharedInstance {
+    static CBLLogSinks* sharedInstance = nil;
+    static dispatch_once_t onceToken;
+    dispatch_once(&onceToken, ^{
+        sharedInstance = [[self alloc] initWithDefault];
+    });
+    return sharedInstance;
+}
```
pasin, line 132: This is not needed as the created CBLLogSinks instance is no used.

Objective-C/LogSink/CBLLogSinks.mm
```diff
@@ -168,6 +185,51 @@ + (void) writeCBLLog: (C4LogDomain)c4domain level: (C4LogLevel)c4level message:
     [CBLLogSinks.custom writeLogWithLevel: level domain: domain message :message];
 }
 
+void writeCBLLogMessage(C4LogDomain domain, C4LogLevel level, NSString *msg, ...) {
```
pasin, line 188: I think we should separate the responsibility of Log/Logger and LogSinks.
This should be at CBLLog.m/mm where the code uses for writing their log messages. Then CBLLog will use CBLLogSink(s) to forward to log to the enabled destination.

Objective-C/CBLDatabase.mm
```diff
@@ -108,7 +107,7 @@ + (void) initialize {
     if (self == [CBLDatabase class]) {
         NSLog(@"%@", [CBLVersion userAgent]);
         // Initialize logging
-        CBLAssertNotNil(CBLLog.sharedInstance);
+        CBLAssertNotNil(CBLLogSinks.sharedInstance);
```
pasin, line 110: This can be just `CBLAssertNotNil(CBLLogSinks.self);`.

pasin (changes requested):

PR #3513, opened 2026-04-21
CBL-8188 : Fix deadlock in CBLDatabase unregisterActiveService

pasin:
When close() waits for active services to stop, it holds _closeCondition and rechecks _activeServices under _mutex. unregisterActiveService() used to do the opposite: hold _mutex, then lock _closeCondition to broadcast. If an async service shutdown finished during close(), the two threads could deadlock on opposite locks.

This stayed latent because it requires a narrow timing window while closing with an active service that unregisters from another thread, so most shutdown paths never overlapped that way. This was discovered while running multipeer replicator tests.

Fix by removing the service under _mutex, recording whether the active set became empty, and broadcasting on _closeCondition only after releasing _mutex.

### DIFF
--- a/Objective-C/CBLDatabase.mm
+++ b/Objective-C/CBLDatabase.mm
@@ -900,13 +900,17 @@ static C4DatabaseConfig2 c4DatabaseConfig2 (CBLDatabaseConfiguration *config) {
 }
 
 - (void) unregisterActiveService: (id<CBLDatabaseService>)service {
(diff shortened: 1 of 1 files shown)


pasin:
PR validation failure is fixed in https://github.com/couchbase/couchbase-lite-ios/pull/3515.

jianminzhao (approved):

PR #3542, opened 2026-06-05
Switch to use RevTree and Restore Legacy APIs

pasin:
* The release/3.4 is branched from 4.1.0, and this ticket is preparing for the 3.4.0 release.

* Switched back to use RevTree instead of VV.
 
* Restored the legacy deprecated APIs removed in 4.0.0 to avoid breaking changes from 3.3.

* Not Restored DatabaseConfiguration's mmapEnabled property (CBL-8396).

* Fixed default conflict resolution logic when two revisions have the same generation (CBL-8395).

* Companion PR : https://github.com/couchbaselabs/couchbase-lite-ios-ee/pull/341


pasin:
This is the strategy chosen for rolling out the 3.4 release and was agreed upon in a sync up meeting.

borrrden (approved):

